PR #190, opened 2021-09-09
Reverse register rename rewinding order on flush

viper12590:
When SimEng flushes an instruction that sets multiple destinations as
same register, the rewinding of register renaming fails.
This is because the order of applying rewinding by historyTable_ is
in the wrong way.
Ultimately, it keeps the wrong physical register (which was freed) in mappingTable_.
Therefore, the order of calling rewind() was reversed to have the correct
order of updating mappingTable_ with historyTable_.

### DIFF
--- a/src/lib/pipeline/ReorderBuffer.cc
+++ b/src/lib/pipeline/ReorderBuffer.cc
@@ -79,7 +79,9 @@ void ReorderBuffer::flush(uint64_t afterSeqId) {
       break;
     }
 
-    for (const auto& reg : uop->getDestinationRegisters()) {
+    auto destinations = uop->getDestinationRegisters();
+    for (int i = destinations.size() - 1; i >= 0; i--) {
+      const auto& reg = destinations[i];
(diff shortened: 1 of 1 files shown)


src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -79,7 +79,9 @@ void ReorderBuffer::flush(uint64_t afterSeqId) {
       break;
     }
 
-    for (const auto& reg : uop->getDestinationRegisters()) {
+    auto destinations = uop->getDestinationRegisters();
```
jj16791, line 82: May be worth adding a small comment above the loop describing why it's done in reverse?

FinnWilkinson (approved):
Looks fine to me.

mutalibmohammed (approved):
Looks good to me

jj16791 (approved):
Ensure that the squash and merge message is reflective of the changes made and concise.

PR #191, opened 2021-09-16
Optimizations

mutalibmohammed:
The main theme of this pull request is optimizing memory allocations/deallocations. It was found while profiling that on average, 10-20 memory allocations/deallocations would occur per cycle. By reducing the no. of allocations, I hope to reduce cycle time.

New language features and smaller changes were preferred over big structural changes to improve performance.

* 6bf6aad3 ... 5410557e primarily change function parameters to accept const references. This allows them to accept both lvalue and rvalue references; avoiding a copy.

*  member initializer list were also added/improved wherever possible. This change avoids having to a call the constructor of a member twice.

>Several of these changes were suggested by the static analyzer `CppCheck`.

* b0c1773b uses a new C++17 method `try_emplace` to construct an element in place avoiding a copy. It also has better syntax to pass constructor arguments compared to pre C++17 methods that require `std::piecewise_construct`.

* 16f27587 changes `getSupportedPorts` to return a reference to a vector. The `aarch64::Instruction` class has a member vector that stores the ports, and it is alive for as long as the instruction instance is, hence we return a reference to it to avoid copying.

* a0aebb11 adds a new CMake flag `SIMENG_OPTIMIZE`. This flag turns on Link-Time Optimizations allowing the compiler to optimize across translation units. Additionally, on x86 systems, it sets the `-march` and `-mtune` compiler flags.

* 856dc967 overloads `setMemoryAddresses` to accept rvalue references. This overload can reuse the resources of its parameter. The vectors created in `Instruction::generateAddresses` are deleted at the end of function. This presents an opportunity to reuse resources.

* bea1f887 creates a new member `flushed_` to avoid allocating a map every flush. The sets are still allocated/deallocated every flush, perhaps using a stack allocator here could be beneficial.

* f4a1dbb4 and ac830725 adds a second parameter `capacity` to the `RegisterValue` ctor that takes an array as a reference. In aarch64, FP, SVE and NEON instructions use the same register bank of 256 byte registers. The updated ctor makes it easier to create `RegisterValue` of the correct size. 

    `Instruction::execute` was updated to use the new ctor. This avoids data copying and an extra ctor call at the end of the function when zero extending.

>The `RegisterValue` class along with `Instruction` class are a major source of memory allocations. The next few commits focus on reducing it.

* b40fde09 experiments with using C++17's new `std::unsynchronized_pool_resource`. It is a general-purpose, non thread-safe memory pool implementation. However, it was abandoned as it was not implemented in LLVM/Clang. 

* a0684490 experiments with using raw pointers instead of shared_ptr. This required implementing copy and move constructors. This change resulted in a performance regression and was reverted.

* ee4ec4dd and cbd0d39e implement a simple memory pool implementation that is based on a free list. Each node of the free list is stored in the free chunks of memory. The memory pool grows by a factor of 2 when exhausted.

* ce708b07 changes the `predict` and `update` functions in `BranchPredictor` to accept a reference to shared_ptr. This is to avoid the cost of atomic increment/decrement. The lifetime of the ptr is expected to out last `predict` and `update`, hence it is safe.

### DIFF
--- a/src/include/simeng/RegisterValue.hh
+++ b/src/include/simeng/RegisterValue.hh
@@ -5,8 +5,12 @@
 #include <cassert>
 #include <cstring>
 #include <memory>
+#include <memory_resource>
 
 namespace simeng {
+
+/** A memory pool used to allocate memory for the RegisterValue class. */
+extern std::pmr::unsynchronized_pool_resource mem_pool;
 
 /** A class that holds an arbitrary region of immutable data, providing casting
  * and data accessor functions. For values smaller than or equal to
@@ -32,12 +36,15 @@ class RegisterValue {
                                    0);
       }
     } else {
-      void* data = calloc(1, bytes);
+      void* data = mem_pool.allocate(bytes);
+      std::memset(data, 0, bytes);
 
       T* view = reinterpret_cast<T*>(data);
       view[0] = value;
 
-      this->ptr = std::shared_ptr<char>(static_cast<char*>(data), free);
+      this->ptr = std::shared_ptr<char>(
+          static_cast<char*>(data),
+          [bytes](void* ptr) { mem_pool.deallocate(ptr, bytes); });
     }
   }
 
@@ -51,11 +58,14 @@ class RegisterValue {
     if (isLocal()) {
       dest = this->value;
     } else {
-      dest = static_cast<char*>(calloc(1, capacity));
-      this->ptr = std::shared_ptr<char>(dest, free);
+      dest = static_cast<char*>(mem_pool.allocate(capacity));
+      this->ptr = std::shared_ptr<char>(
+          dest, [capacity](void* ptr) { mem_pool.deallocate(ptr, capacity); });
     }
     assert(dest && "Attempted to dereference a NULL pointer");
     std::memcpy(dest, ptr, bytes);
+
+    if (capacity > bytes) std::fill(dest + bytes, dest + capacity, 0u);
   }
 
   /** Create a new RegisterValue of size `bytes`, copying data from `ptr`. */
--- a/src/include/simeng/RegisterValue.hh
+++ b/src/include/simeng/RegisterValue.hh
@@ -5,12 +5,8 @@
 #include <cassert>
 #include <cstring>
 #include <memory>
-#include <memory_resource>
 
 namespace simeng {
-
-/** A memory pool used to allocate memory for the RegisterValue class. */
-extern std::pmr::unsynchronized_pool_resource mem_pool;
 
 /** A class that holds an arbitrary region of immutable data, providing casting
  * and data accessor functions. For values smaller than or equal to
@@ -18,7 +14,32 @@ extern std::pmr::unsynchronized_pool_resource mem_pool;
  * allocated and the data is stored there. */
 class RegisterValue {
  public:
-  RegisterValue();
+  RegisterValue() noexcept {};
+
+  RegisterValue(const RegisterValue& other) noexcept : bytes(other.bytes) {
+    if (isLocal()) {
+      std::memcpy(value, other.value, bytes);
+    } else {
+      ptr = new (std::nothrow) char[bytes];
+      std::memcpy(ptr, other.ptr, bytes);
+    }
+  }
+  RegisterValue& operator=(RegisterValue other) noexcept {
+    swap(*this, other);
+    return *this;
+  }
+
+  RegisterValue(RegisterValue&& other) noexcept { swap(*this, other); }
+
+  ~RegisterValue() { delete[] ptr; }
+
+  friend void swap(RegisterValue& first, RegisterValue& second) {
+    using std::swap;
+
+    swap(first.bytes, second.bytes);
+    swap(first.ptr, second.ptr);
+    swap(first.value, second.value);
+  }
 
   /** Create a new RegisterValue from a value of arbitrary type (except
    * pointers), zero-extending the allocated memory space to the specified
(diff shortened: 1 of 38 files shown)


jj16791:
#rerun tests

jj16791:
#rerun tests

src/include/simeng/RegisterValue.hh
```diff
@@ -31,35 +37,49 @@ class RegisterValue {
                                    0);
       }
     } else {
-      void* data = calloc(1, bytes);
+      // void* data = calloc(1, bytes);
```
FinnWilkinson, line 40: Remove comment

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -2,6 +2,11 @@
 #include <limits>
 #include <tuple>
 
+// Temporary; until execute has been verified to work correctly.
+#ifndef NDEBUG
+#include <iostream>
+#endif
+
```
FinnWilkinson, line 9: Still needed?

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -907,19 +908,19 @@ void Instruction::execute() {
...
-      results[0] = out;
+      results[0] = {out, 256};
       break;
     }
     case Opcode::AArch64_CPYi32: {  // dup vd, vn.s[index]
       const uint32_t* vec = operands[0].getAsVector<uint32_t>();
-      uint32_t out[4] = {vec[metadata.operands[1].vector_index], 0, 0, 0};
-      results[0] = out;
+      uint32_t out[1] = {vec[metadata.operands[1].vector_index]};
+      results[0] = {out, 256};
```
FinnWilkinson, line 917: could reducing the size of the output array not lead to random data being left in the vector instead of being set to `0`?

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -1720,41 +1706,39 @@ void Instruction::execute() {
     }
     case Opcode::AArch64_FCVTDSr: {  // fcvt dd, sn
       // TODO: Handle NaNs, denorms, and saturation?
-      double out[2] = {static_cast<double>(operands[0].get<float>()), 0.0};
-      results[0] = out;
+      results[0] =
+          RegisterValue(static_cast<double>(operands[0].get<float>()), 256);
```
FinnWilkinson, line 1710: Result gone from 2 element array of double to a single `RegisterValue` of 256 bytes? Or am I interpreting this incorrectly?

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -2954,47 +2899,47 @@ void Instruction::execute() {
       break;
     }
     case Opcode::AArch64_LDNPSi: {  // ldnp st1, st2, [xn, #imm]
-      results[0] = memoryData[0].zeroExtend(4, 16);
-      results[1] = memoryData[1].zeroExtend(4, 16);
+      results[0] = memoryData[0].zeroExtend(4, 256);
+      results[1] = memoryData[1].zeroExtend(4, 256);
```
FinnWilkinson, line 2903: Again, unless I am interpreting this wrong - zero extending up to 256 bytes?

src/include/simeng/RegisterValue.hh
mutalibmohammed: Good catch.

src/lib/arch/aarch64/Instruction_execute.cc
mutalibmohammed: I think so. I have only tested it on my end, so it would be better if the team could test as well.

src/lib/arch/aarch64/Instruction_execute.cc
mutalibmohammed: You are interpreting this correctly. 

Arrays were used previously as a way to construct 256 byte `RegisterValue`'s. First, create an array of size 16 bytes (padded with 0s) within the case statement. At the end of the `execute` function,  each `RegisterValue` would be checked for certain conditions. If the conditions are met, it is zero extended from 16 bytes to 256 bytes as shown below:
``` cpp
if ((destinationRegisters[i].type == RegisterType::VECTOR) && !isSVEData_)
      results[i] = results[i].zeroExtend(16, 256); 
```
My change reduces this process to a single constructor call. In this case, we do not need the array at all.

src/lib/arch/aarch64/Instruction_execute.cc
mutalibmohammed: I'm not sure what you mean by random data being left in the vector, could you please clarify?

The 0s are present to pad the array up to 16 bytes. And as explained below, the `RegisterValue` will be zero-extended to 256 bytes at the end of the function. I'm not sure why it is 16 bytes, perhaps because the registers for FP/NEON instruction       are 16 bytes.

The arrays no longer need to padded with the updated constructor as they are automatically zero extended to 256 bytes.

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -199,8 +204,7 @@ void Instruction::execute() {
     case Opcode::AArch64_ADDv1i64: {  // add dd, dn, dm
       const uint64_t n = operands[0].get<uint64_t>();
       const uint64_t m = operands[1].get<uint64_t>();
-      uint64_t out[2] = {static_cast<uint64_t>(n + m), 0};
-      results[0] = out;
+      results[0] = {n + m, 256};
```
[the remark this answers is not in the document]
mutalibmohammed: `256` in the second statement is the `bytes` parameter in the `RegisterValue` constructor. I am using the [list initialization syntax.](https://en.cppreference.com/w/cpp/language/list_initialization) I chose this form in the interest of verbosity, but I realize that this could be confusing.

`result[0] = out` is equivalent to ` result[0] = RegisterValue(out)`. The constructor is called implicitly in the first case. The only change is that the size is passed as well.

See explanation below as to why an array is used and why the second element is 0.

src/lib/arch/aarch64/Instruction_execute.cc
mutalibmohammed: That's correct. We no longer need to zero extend at the end of the execute function as they were being done previously.

src/lib/arch/aarch64/Instruction_execute.cc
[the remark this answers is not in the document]
FinnWilkinson: OK this makes sense, thanks for clarifying

src/lib/arch/aarch64/Instruction_execute.cc
FinnWilkinson: I didn't realise that it would be zero-extended at the end of the function.... thanks again for clarification

jj16791:
#rerun tests

jj16791:
#rerun tests

jj16791:
#rerun tests

jj16791:
#rerun tests

viper12590 (approved):
Looks good to me.

jj16791 (approved):

FinnWilkinson (approved):
All looks good

PR #192, opened 2021-09-16
Bug fixes

mutalibmohammed:
When SimEng flushes an instruction that sets multiple destinations as same register, the rewinding of register renaming fails. This is because the order of applying rewinding by `historyTable_` is in the wrong way.

Ultimately, it keeps the wrong physical register (which was freed) in `mappingTable_`. Therefore, the order of calling `rewind()` was reversed to have the correct order of updating `mappingTable_`  with `historyTable_`.

In `FetchUnit::tick`, If the `pc_` was not aligned to the blockSize boundary and the `fetchBuffer_` was empty, the `fetchData` would not be copied but used directly as an optimization. However, if the `fetchData` was not enough to start decoding right away, the function would exit and the `fetchData` would be lost.  To fix the bug, the optimization was removed and `fetchData` is always copied onto the `fetchBuffer_`. There was no observed performance difference.

Additionally, the pointer to the buffer passed to `predecode` is not guaranteed to be aligned. This caused misalignment bugs as `aarch64::predecode` was expecting it to be 4 byte aligned. A workaround proposed by @FinnWilkinson fixed the bug by copying the buffer into a local variable.

The `ROR` implementation was found to be buggy, hence a modified version using modular arithmetic was implemented.

### DIFF
--- a/src/lib/arch/aarch64/InstructionMetadata.cc
+++ b/src/lib/arch/aarch64/InstructionMetadata.cc
@@ -362,10 +362,12 @@ InstructionMetadata::InstructionMetadata(const cs_insn& insn)
       break;
     case Opcode::AArch64_RET:
       // RET doesn't list use of x30 (LR) if no register is supplied
-      operandCount = 1;
-      operands[0].type = ARM64_OP_REG;
-      operands[0].reg = ARM64_REG_LR;
-      operands[0].access = CS_AC_READ;
+      if (operandCount == 0) {
+        operandCount = 1;
+        operands[0].type = ARM64_OP_REG;
+        operands[0].reg = ARM64_REG_LR;
+        operands[0].access = CS_AC_READ;
+      }
       groupCount = 1;
       groups[0] = CS_GRP_JUMP;
       break;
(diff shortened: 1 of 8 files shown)


jj16791:
#rerun tests

FinnWilkinson (approved):
All looks good to me

jj16791:
#rerun tests

viper12590 (approved):
Looks good to me

viper12590 (commented):
I found a bug that can be critical, so I commit the fix straight to this branch.
Please have a look

src/lib/arch/aarch64/InstructionMetadata.cc
```diff
@@ -362,10 +362,12 @@ InstructionMetadata::InstructionMetadata(const cs_insn& insn)
...
-      operandCount = 1;
-      operands[0].type = ARM64_OP_REG;
-      operands[0].reg = ARM64_REG_LR;
-      operands[0].access = CS_AC_READ;
+      if (operandCount == 0) {
+        operandCount = 1;
+        operands[0].type = ARM64_OP_REG;
+        operands[0].reg = ARM64_REG_LR;
+        operands[0].access = CS_AC_READ;
+      }
```
FinnWilkinson, line 370: Should probably add to the tests (or create some if none are present) to check if doing `ret xm` works correctly.

FinnWilkinson (approved):
All looks good

jj16791 (approved):

PR #193, opened 2021-09-27
Fix frsqrte behaviour (and minor changes)

viper12590:
This branch was initially created to fix the error output difference between
SimEng and hardware. (SimEng has greater error than hardware)
There was result differences between hardware and SimEng when the
instruction `frsqrte` was used.
Therefore, it was suspected to be the cause and some changes are made
as below.
Even though the error output was reduced by very small amount after the fix,
the behaviour of `frsqrte` is now exactly same as hardware.

The use of `1/sqrt(x)` and `1/sqrtf(x)` in `frsqrte` are replaced with the methods
based on the pseudo code from _Arm Architecture Reference Manual
Armv8, for Armv8-A architecture profile_.

While checking wrongly implemented instructions, better tests are added to
instructions `st2 (multi struct)`, `faddp (single precision vector)`,
`mov (general to vector)`, and `frsqrte`.

`st1 (two 16b vectors variant)` and `st2 (two 4s vectors post index variant)`
are found that they can have minor improvements.
Address generation of `st1` is simplified, so it can simply store data vector
by vector.
Type of operands in `st2` are changed to `uint32_t` to ensure non-floating point
data can also be handled.

### DIFF
--- a/src/lib/arch/aarch64/Instruction_execute.cc
+++ b/src/lib/arch/aarch64/Instruction_execute.cc
@@ -183,23 +183,7 @@ bool conditionHolds(uint8_t cond, uint8_t nzcv) {
   return (inverse ? !result : result);
 }
 
-void fpUnpack(uint32_t* sign, uint32_t* exponent, uint32_t* mantisa, float fp) {
-  uint32_t fp_casted = reinterpret_cast<uint32_t&>(fp);
-  *sign = (fp_casted >> 31) & 1;
-  *mantisa = fp_casted & (1 << 23 - 1);
-  *exponent = fp_casted >> 23 & 255;
-
-  // handling subnormal
-  if (*exponent == 0) {
-    while (*mantisa >> 23 == 0) {
-      *mantisa = (*mantisa << 1) & (1 << 23 - 1);
-      *exponent -= 1;
-    }
-    *mantisa = (*mantisa << 1) & (1 << 23 - 1);
-  }
-}
-
-uint32_t recipSqrtEstimate(uint32_t scaled) {
+uint64_t recipSqrtEstimate(uint64_t scaled) {
   assert((128 <= scaled && scaled < 512) &&
          "Invalid argument, not in the range of 128 to 511");
   if (scaled < 256) {  // 0.25 to 0.5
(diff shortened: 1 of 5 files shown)


src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -182,6 +182,126 @@ bool conditionHolds(uint8_t cond, uint8_t nzcv) {
...
+    } else {
+      uint64_t packed = (result_exp << 52) | ((estimate & 0xff) << 44);
+      assert(sizeof(packed) == sizeof(result) &&
+             "size of packed bits are different with floating point type");
+      memcpy(&result, &packed, sizeof(T));
+    }
+  }
+  return result;
+}
+
```
FinnWilkinson, line 304: I was wondering if an adapted version of the Quake-3 Fast inverse-square root function could be used here? By no means is it any more readable (probably far less so) but it is only about 10 lines of code (and very quick). Also unsure of how its accuracy compares to your implementation...

viper12590: The aim of this was to make SimEng to calculate exactly same estimate of reciprocal square root as hardware. So, I think it would be hard to achieve if we use different methods.

FinnWilkinson (approved):
All looks fine to me

jj16791 (commented):
Whilst it's good that we're improving the functional accuracy of running the minifmm benchmark, I worry that the amount of code used and increased complexity/decreased readability outweighs the accuracy gained. Whilst we want to improve accuracy, we equally want to maintain a clean/ease-to-understand code-base. It seems that there's a lot of work done for an un-proportional positive effect. For the sake of fairness, @FinnWilkinson do you have anything to add to this comment/point-of-view

FinnWilkinson:
> Whilst it's good that we're improving the functional accuracy of running the minifmm benchmark, I worry that the amount of code used and increased complexity/decreased readability outweighs the accuracy gained. Whilst we want to improve accuracy, we equally want to maintain a clean/ease-to-understand code-base. It seems that there's a lot of work done for an un-proportional positive effect. For the sake of fairness, @FinnWilkinson do you have anything to add to this comment/point-of-view

I agree that it decreases readability a fair bit, and adds alot of code - but if the accuracy of `1/sqrt(x)` etc was out for a wide range of hardware then it could potentially be justified. Also, if using `1/sqrt(x)` affected benchmark results by a resonable degree then it could be justified to use this new, more accurate implementation. 

It would also be useful to know what the original accuaracy discrepency when using `1/sqrt(x)` was; as if this is fairly small then again it may be worth accepting this discrepency in order to maintain a more readable code base.

TLDR : maybe it is worth evaluating the original accuracy problem as well as how well the new solution works on a range of benchmarks & processor models before deciding on a solution.

jj16791:
I think it's best if we leave this out for v0.9 and investigate. @viper12590 could you move this branch from your fork to a branch in SimEng please

viper12590:
I made a new branch `frsqrte-improvement` and changed the base of this PR to it.

viper12590:
I made a mistake while making a `frsqrte-improvement` branch, so I made new `frsqrte-improvement2` branch and changed PR base to it

PR #194, opened 2021-09-30
Minifmm aarch64 coverage

viper12590:
This branch adds more instructions to support OpenMP enabled minifmm.
Currently, GCC and Armclang compiled `fmm.omp-for` can be fully executed.

When SimEng runs Armclang compiled version of `fmm.omp-for`, a bug was found.
It made SimEng to stuck in infinite loop.
This bug is fixed at commit [`19b4a5f`](https://github.com/UoB-HPC/SimEng/pull/192/commits/19b4a5fa9b866981c5a210a96b32cbd96b640942)
To run Armclang version without issues, it will be ideal to merge PR [`Bug fixes #192`](https://github.com/UoB-HPC/SimEng/pull/192) before merging this PR.

### DIFF
--- a/src/lib/arch/aarch64/Instruction_execute.cc
+++ b/src/lib/arch/aarch64/Instruction_execute.cc
@@ -211,7 +211,7 @@ void Instruction::execute() {
       for (int i = 0; i < 16; i++) {
         out[i] = static_cast<uint8_t>(n[i] + m[i]);
       }
-      results[0] = out;
+      results[0] = {out, 256};
       break;
     }
     case Opcode::AArch64_ADDv1i64: {  // add dd, dn, dm
@@ -243,7 +243,7 @@ void Instruction::execute() {
       for (int i = 0; i < 4; i++) {
         out[i] = static_cast<uint16_t>(n[i] + m[i]);
       }
-      results[0] = out;
+      results[0] = {out, 256};
       break;
     }
     case Opcode::AArch64_ADDv4i32: {  // add vd.4s, vn.4s, vm.4s
@@ -315,7 +315,7 @@ void Instruction::execute() {
       for (int i = 0; i < 8; i++) {
         out[i] = static_cast<uint16_t>(n[i] + m[i]);
       }
-      results[0] = out;
+      results[0] = {out, 256};
       break;
     }
     case Opcode::AArch64_ADDv8i8: {  // add vd.8b, vn.8b, vm.8b
@@ -325,7 +325,7 @@ void Instruction::execute() {
       for (int i = 0; i < 8; i++) {
         out[i] = static_cast<uint8_t>(n[i] + m[i]);
       }
-      results[0] = out;
+      results[0] = {out, 256};
       break;
     }
     case Opcode::AArch64_ADCXr: {  // adc xd, xn, xm
@@ -6212,7 +6212,7 @@ void Instruction::execute() {
       for (int i = 0; i < 8; i++) {
         out[i] = n[i + 8] << shift;
       }
-      results[0] = out;
+      results[0] = {out, 256};
       break;
     }
     case Opcode::AArch64_USHLLv4i16_shift: {  // ushll vd.4s, vn.4h, #imm
@@ -6232,7 +6232,7 @@ void Instruction::execute() {
       for (int i = 0; i < 4; i++) {
         out[i] = n[i + 4] << shift;
       }
-      results[0] = out;
+      results[0] = {out, 256};
       break;
     }
     case Opcode::AArch64_USHLLv8i8_shift: {  // ushll vd.8h, vn.8b, #imm
@@ -6241,6 +6241,18 @@ void Instruction::execute() {
       uint16_t out[8] = {0};
       for (int i = 0; i < 8; i++) {
         out[i] = n[i] << shift;
+      }
+      results[0] = {out, 256};
+      break;
+    }
+    case Opcode::AArch64_UUNPKHI_ZZ_D: {  // uunpkhi zd.d, zn.s
+      const uint32_t* n = operands[0].getAsVector<uint32_t>();
+      const uint64_t VL_bits = 512;
+      const uint16_t partition_num = VL_bits / 64;
+      uint64_t out[32] = {0};
+
+      for (int i = 0; i < partition_num; i++) {
+        out[i] = static_cast<uint64_t>(n[partition_num + i]);
       }
       results[0] = out;
       break;
--- a/src/lib/arch/aarch64/Instruction_execute.cc
+++ b/src/lib/arch/aarch64/Instruction_execute.cc
@@ -3808,6 +3808,10 @@ void Instruction::execute() {
       results[0] = memoryData[0].zeroExtend(4, 8);
       break;
     }
+    case Opcode::AArch64_LDARX: {  // ldar xt, [xn]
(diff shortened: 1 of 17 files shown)


src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -196,13 +196,49 @@ void Instruction::execute() {
 
   executed_ = true;
   switch (metadata.opcode) {
+    case Opcode::AArch64_ADDv16i8: {  // add vd.16b, vn.16b, vm.16b
+      const uint8_t* n = operands[0].getAsVector<uint8_t>();
+      const uint8_t* m = operands[1].getAsVector<uint8_t>();
+      uint8_t out[16] = {0};
+      for (int i = 0; i < 16; i++) {
+        out[i] = static_cast<uint8_t>(n[i] + m[i]);
```
FinnWilkinson, line 204: probably don't need the static cast

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -2941,10 +2997,18 @@ void Instruction::execute() {
       results[2] = operands[2].get<uint64_t>() + offset;
       break;
     }
+    case Opcode::AArch64_LDARB: {  // ldarb wt, [xn]
+      results[0] = memoryData[0].zeroExtend(1, 8);
```
FinnWilkinson, line 3001: If this is being written to a `wt` register, shouldn't it be zero extended to `4` bytes not `8`?

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -4830,6 +4896,16 @@ void Instruction::execute() {
       results[0] = mulhi(x, y);
       break;
     }
+    case Opcode::AArch64_USHLLv16i8_shift: {  // ushll2 vd.8h, vn.16b, #imm
+      const uint8_t* n = operands[0].getAsVector<uint8_t>();
+      int64_t shift = metadata.operands[2].imm;
```
FinnWilkinson, line 4901: make `int64_t shift` const

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -4830,6 +4896,16 @@ void Instruction::execute() {
...
+      uint16_t out[8] = {0};
+      for (int i = 0; i < 8; i++) {
+        out[i] = (n[i + 8] << shift) & 0xff;
```
FinnWilkinson, line 4904: Is there any need for the `& 0xff`? As `shift` max value is source element width in bits (i.e. here it will be 8), which is always half of the destination element's width (here this is 16). Could possibly do a `static_cast<uint16_t>` to make it clear what the output format is?

Additionally, doing this operation with `0xff` would confine it to 8 bit values? i.e. `0b1111111111111111 & 0xff = 0b0000000011111111`

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -4840,6 +4916,26 @@ void Instruction::execute() {
...
+    case Opcode::AArch64_USHLLv8i8_shift: {  // ushll vd.8h, vn.8b, #imm
+      const uint8_t* n = operands[0].getAsVector<uint8_t>();
+      int64_t shift = metadata.operands[2].imm;
+      uint16_t out[8] = {0};
+      for (int i = 0; i < 8; i++) {
+        out[i] = (n[i] << shift) & 0xff;
+      }
+      results[0] = out;
+      break;
+    }
```
FinnWilkinson, line 4938: Same comment from above applies here (adjusted for vector width differences)

test/regression/aarch64/instructions/neon.cc
```diff
@@ -2059,6 +2195,76 @@ TEST_P(InstNeon, ushll) {
...
+
+    ldr q0, [x0]
+
+    ushll2 v1.4s, v0.8h, #0
+    ushll2 v2.4s, v0.8h, #8
+    ushll2 v3.4s, v0.8h, #15
+  )");
+  CHECK_NEON(1, uint32_t, {42, 1u << 13, 702, 0});
+  CHECK_NEON(2, uint32_t, {10752, 0, 48640, 0});
+  CHECK_NEON(3, uint32_t, {0, 0, 0, 0});
```
FinnWilkinson, line 2214: Perhaps add tests with `UINT16_MAX` being shifted also?

test/regression/aarch64/instructions/neon.cc
```diff
@@ -2059,6 +2195,76 @@ TEST_P(InstNeon, ushll) {
...
+
+    ldr q0, [x0]
+
+    ushll v2.8h, v0.8b, #0
+    ushll v3.8h, v0.8b, #4
+    ushll v4.8h, v0.8b, #7
+  )");
+  CHECK_NEON(2, uint16_t, {0x7F, (UINT8_MAX), 0xAA, 0x80, 0, 0xbc, 0xde, 0xf0});
+  CHECK_NEON(3, uint16_t, {0xF0, 0xF0, 0xA0, 0, 0, 0xC0, 0xE0, 0});
+  CHECK_NEON(4, uint16_t, {0x80, 0x80, 0, 0, 0, 0, 0, 0});
```
FinnWilkinson, line 2250: I don't think these are correct / make sense? Possibly due to implementation, but (for example) `UINT8_MAX << 4` will be a valid `uint16_t` value (think its 4080 or `0xFF0`)

test/regression/aarch64/instructions/neon.cc
```diff
@@ -2059,6 +2195,76 @@ TEST_P(InstNeon, ushll) {
...
+
+    ldr q0, [x0]
+
+    ushll2 v2.8h, v0.16b, #0
+    ushll2 v3.8h, v0.16b, #4
+    ushll2 v4.8h, v0.16b, #7
+  )");
+  CHECK_NEON(2, uint16_t, {0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0});
+  CHECK_NEON(3, uint16_t, {0x10, 0x20, 0x30, 0x40, 0x50, 0x60, 0x70, 0});
+  CHECK_NEON(4, uint16_t, {0x80, 0, 0x80, 0, 0x80, 0, 0x80, 0});
```
FinnWilkinson, line 2267: same as other comments, I think the implementation is slightly off, as `uint16_t` values can be up to `0xFFFF`

src/lib/arch/aarch64/Instruction_execute.cc
jj16791: both w and x registers point to a 64-bit reg so needs to be extended to 8 in both cases

src/lib/arch/aarch64/Instruction_execute.cc
jj16791: agreed

src/lib/arch/aarch64/Instruction_execute.cc
viper12590: I tested on tx2 and you're right, the correct behaviour of these instructions does not restrict the width of shifted values to the width of source.
I did not know earlier implementation of `ushll vd.4s, vn.4h, #imm` was wrong, so I used it as a reference.
I fixed both my implementation and early implementation.

FinnWilkinson (approved):
All looks good

PR #197, opened 2021-10-05
Minifmm support

viper12590:
This branch adds more instructions to support OpenMP enabled minifmm.
Currently, GCC and Armclang compiled fmm.omp-for can be fully executed.

When SimEng runs Armclang compiled version of fmm.omp-for, a bug was found.
It made SimEng to stuck in infinite loop.
To run Armclang version without issues, it will be ideal to merge PR Bug fixes #192 before merging this PR.


FinnWilkinson (approved):
All looks good

PR #199, opened 2021-10-18
Further AArch64 coverage targeting the CloverLeaf benchmark

FinnWilkinson:
Further coverage of the AArch64 ISA has been made to add support for the following CloverLeaf versions :
- CloverLeaf_Serial compiled with GNU and ArmClang
- CloverLeaf_OpenMP compiled with GNU and ArmClang

 Both benchamrks have been supported and tested using the TX2, armv8.4-a, and armv8.4-a+sve architecture targets on SimEng.

The majority of these commits have been for the ArmClang compiled versions of CloverLeaf_* targeting armv8.4-a+sve, with some helper functions also being created in areas where the same code was being repeatedly used.

The final commit `a1e3a5b` addresses an uncommon issue where specific length binary names (inc. their path) would push necessary null pointers off the top of the stack, resulting in a data abort error.


jj16791 (approved):

jj16791:
#rerun tests

PR #204, opened 2021-12-15
Variable SVE vector length

jj16791:
This PR enables the configuration of the vector length used by the simulated SVE instructions. The test suite has been updated to reflect the use of vector length values from 128 up to 2048.


jj16791:
#rerun test

jj16791:
#rerun tests

PR #210, opened 2022-01-07
Minor bux fixes

jj16791:
This PR contains some minor bug fixes including: 
- An issue whereby fresh clones/installs of SimEng which had no `specialFiles` directory. It would fail in the `RemoveExistingSFDir` function of `SpecialFileDirGen.cc` due to `No such file or directory` error.
- Removal of use of the `getdents64` system call in favour of custom implementation to avoid re-occurring unpredictable issues whereby builds fail due to the system calls inclusion.


src/lib/SpecialFileDirGen.cc
```diff
@@ -20,10 +20,10 @@ SpecialFileDirGen::SpecialFileDirGen(YAML::Node config) {
...
 void SpecialFileDirGen::RemoveExistingSFDir() {
-  const std::string rm_input =
-      "rm -r " + specialFilesParentDir_ + "/specialFiles/";
   const std::string mk_input =
-      "mkdir " + specialFilesParentDir_ + "/specialFiles/";
+      "mkdir -p " + specialFilesParentDir_ + "/specialFiles/";
+  const std::string rm_input =
+      "rm -r " + specialFilesParentDir_ + "/specialFiles/*";
   system(rm_input.c_str());
   system(mk_input.c_str());
```
FinnWilkinson, line 28: Need to swap the two system calls around

FinnWilkinson (approved):
All looks good to me

PR #212, opened 2022-01-31
Instruction_execute.cc Reconfig

FinnWilkinson:
This main objective of this Pull Request was to reduce the amount of repeated code in the `Instruction_execute.cc` switch case. Currently, each instruction will have its accosiated code contained within its switch case; but as there are many variations of the same (or similar) instructions this leads to alot of repeated code.

To combat this, 13 new header files have been created which contain 'Helper Functions' which are type agnostic, and in some cases accomodate multiple different instructions by taking in a Lambda function as part of the function's parameters (e.g. a single LogicOperation helper function could perform AND, XOR, EOR, etc). 

Most of the currently supported instructions now use these helper functions, with the exception of LOAD and STORE type instructions (excluded to not interfere with ongoing Micro-Op support), and single line instructions (e.g.  type conversion, mov, excluded due to having a small footprint in the code already).

The work included within this PR is a first step towards complete use of helper functions. As well as converting LOAD and STORE instructions to use helper functions, there are plans to further generalise the currently implemented helper functions to both reduce their code footprint and to make them easier to utilise.

### DIFF
--- a/src/include/simeng/arch/aarch64/helpers/arithmetic.hh
+++ b/src/include/simeng/arch/aarch64/helpers/arithmetic.hh
@@ -7,8 +7,9 @@ namespace arch {
 namespace aarch64 {
 class arithmeticHelp {
  public:
-  /** Helper function for instructions with the format `add rd, rn, rm`. Returns
-   * Single value. */
+  /** Helper function for instructions with the format `add rd, rn, rm`.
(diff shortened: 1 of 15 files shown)


jj16791 (approved):

FinnWilkinson:
Docs need updating to explain how to utilise the new helper functions.

PR #219, opened 2022-03-25
FFTW Support

FinnWilkinson:
This PR mainly contains more instruction implementations from the Armv8+sve ISA to support the FFTW binary. It also includes other minor updates, the most notable of which are :

 - Changed max destination registers from 3 to 4 (to accomodate for LD4D)
 - Added Virtual Counter Timer system register (cntvct_el0 for Armv8) update functionality at a given frequency
 - Added Virtual Counter Timer update frequency to the YAML config file


jj16791 (commented):
Need to fix CI failure but code looks good

PR #220, opened 2022-03-28
Micro op and fusion

jj16791:
This PR is primarily concerned with the introduction of micro-operations/instruction splitting. Key developments are:

* Splitting for various base and Neon store/load macro-ops
* New aarch64 micro-decoder which splits a macro-op into many micro-ops
* Changes to the generic pipeline to support micro-operations
* Micro-operation identification variables


FinnWilkinson (approved):
All looks good

PR #226, opened 2022-04-29
New generic branch prediction

jj16791:
These developments implement a new GenericPredictor class which contains:

- Static predictor based on pre-allocated branch type.
- A Branch Target Buffer (BTB) with a local and global indexing scheme and a 2-bit saturating counter.
- A Return Address Stack (RAS) is also in use.

Various changes have occurred to the parent branch class including:

- The passing of a branch type to help prediction (i.e. always taking unconditional branches)
- The passing of a known target extracting during pre-decoding to help prediction (i.e branch instructions to labels)
- Callable flush logic for the associated branch prediction schemes with an address as the parameter

### DIFF
--- a/src/include/simeng/pipeline/FetchUnit.hh
+++ b/src/include/simeng/pipeline/FetchUnit.hh
@@ -65,7 +65,7 @@ class FetchUnit {
   uint64_t getBranchStalls() const;
 
   /** Clear the loop buffer. */
-  void flush();
+  void flushLoopBuffer();
 
  private:
   /** An output buffer connecting this unit to the decode unit. */
--- a/src/lib/models/outoforder/Core.cc
+++ b/src/lib/models/outoforder/Core.cc
@@ -60,13 +60,13 @@ Core::Core(MemoryInterface& instructionMemory, MemoryInterface& dataMemory,
             dispatchIssueUnit_.forwardOperands(regs, values);
           },
           config["L1-Cache"]["Exclusive"].as<bool>(),
-          config["L1-Cache"]["Load-Bandwidth"].as<uint8_t>(),
-          config["L1-Cache"]["Store-Bandwidth"].as<uint8_t>(),
-          config["L1-Cache"]["Permitted-Requests-Per-Cycle"].as<uint8_t>(),
-          config["L1-Cache"]["Permitted-Loads-Per-Cycle"].as<uint8_t>(),
-          config["L1-Cache"]["Permitted-Stores-Per-Cycle"].as<uint8_t>()),
+          config["L1-Cache"]["Load-Bandwidth"].as<uint16_t>(),
+          config["L1-Cache"]["Store-Bandwidth"].as<uint16_t>(),
+          config["L1-Cache"]["Permitted-Requests-Per-Cycle"].as<uint16_t>(),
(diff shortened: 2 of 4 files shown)


src/include/simeng/BTB_BWTPredictor.hh
```diff
@@ -1,71 +1,73 @@
-#pragma once
```
FinnWilkinson, line 1: Everything in this file seems to be commented out? If not used then should the file just be deleted?

src/include/simeng/GenericPredictor.hh
```diff
@@ -0,0 +1,76 @@
...
+   * counter and a branch target. */
+  std::vector<std::pair<uint8_t, uint64_t>> btb_;
+
+  /** The previous BTB index calculated for an address. */
+  std::map<uint64_t, uint64_t> btbHistory_;
+
+  /** The number of bits used to form the saturating counter in a BTB entry. */
+  uint64_t satCntBits_;
+
+  /** A n-bit history of previous branhc directions where n is equal to
```
FinnWilkinson, line 57: misspelling of branch in comment

src/lib/BTB_BWTPredictor.cc
```diff
@@ -1,96 +1,99 @@
-#include "simeng/BTB_BWTPredictor.hh"
```
FinnWilkinson, line 1: Again, whole file seems to be commented out so may be worth just deleting file?

src/include/simeng/pipeline/FetchUnit.hh
```diff
@@ -37,6 +64,9 @@ class FetchUnit {
    * branch. */
   uint64_t getBranchStalls() const;
 
+  /** Clear the loop buffer. */
+  void flush();
+
```
FinnWilkinson, line 69: Given this is in `FetchUnit.hh`, perhaps this should be renames to `flushLoopBuffer()` to avoid any possible confusion

src/lib/models/outoforder/Core.cc
```diff
@@ -60,17 +60,23 @@ Core::Core(MemoryInterface& instructionMemory, MemoryInterface& dataMemory,
...
-                     registerAliasTable_, loadStoreQueue_,
-                     [this](auto instruction) { raiseException(instruction); }),
+          config["L1-Cache"]["Load-Bandwidth"].as<uint8_t>(),
+          config["L1-Cache"]["Store-Bandwidth"].as<uint8_t>(),
+          config["L1-Cache"]["Permitted-Requests-Per-Cycle"].as<uint8_t>(),
+          config["L1-Cache"]["Permitted-Loads-Per-Cycle"].as<uint8_t>(),
+          config["L1-Cache"]["Permitted-Stores-Per-Cycle"].as<uint8_t>()),
       fetchUnit_(fetchToDecodeBuffer_, instructionMemory, processMemorySize,
-                 entryPoint, config["Core"]["Fetch-Block-Size"].as<uint16_t>(),
+                 entryPoint, config["Fetch"]["Fetch-Block-Size"].as<uint8_t>(),
```
FinnWilkinson, line 69: Weren't all the config file options updated to take in `uint16_t` rather than `uint8_t`?

jj16791: Good spot, looks like a mistake in rebasing/merging older work

PR #227, opened 2022-05-03
Correct Result Forwarding - SE-129

FinnWilkinson:
This Pull Request addresses the naive implementation of instruction result forwarding that was previously implemented. Rather than allowing for any result to be forwarded to any dependant instruction waiting on its result, only certain instructions can recieve the forwarded result based upon the instruction types of the Producing and Consuming instructions. The mapping of allowed forwardings and the latency of the forwarding is based upon the A64FX implementation documentation.


src/lib/pipeline/DispatchIssueUnit.cc
```diff
@@ -53,10 +54,66 @@ DispatchIssueUnit::DispatchIssueUnit(
...
+      reservationStations_[rsInfo.first].ports[rsInfo.second].ready.push_back(
+          std::move(entry.uop));
+    }
+
+    // Remove completed instructions from dependantInstructions_ or increase
+    // index
+    if (supplied)
+      dependantInstructions_.erase(start + index);
+    else
+      index++;
```
jj16791, line 113: Why aren't we just incrementing the iterator here? Why are we using a seperate index value?

src/lib/pipeline/DispatchIssueUnit.cc
```diff
@@ -181,12 +240,27 @@ void DispatchIssueUnit::forwardOperands(const span<Register>& registers,
...
-        auto rsInfo = portMapping_[entry.port];
-        reservationStations_[rsInfo.first].ports[rsInfo.second].ready.push_back(
-            std::move(entry.uop));
+      int8_t forwardLatency =
+          (enableBypassLatency_)
+              ? insn->canForward(insn->getProducerGroup(),
+                                 entry.uop->getConsumerGroup())
+              : 0;
+      if (forwardLatency == 0) {
+        // If forwarding latency is 0 then can be issued immidiately
```
jj16791, line 249: Typo

src/lib/pipeline/DispatchIssueUnit.cc
```diff
@@ -53,10 +54,66 @@ DispatchIssueUnit::DispatchIssueUnit(
...
+        // Add the now-ready instruction to the relevant ready queue
+        auto rsInfo = portMapping_[entry.port];
+        reservationStations_[rsInfo.first].ports[rsInfo.second].ready.push_back(
+            std::move(entry.uop));
+      }
+
+      // Remove entry from vector
+      waitingInstructions_.erase(begin + index);
+    } else
+      index++;
```
jj16791, line 82: Similar comment about not just incrementing the iterator

src/include/simeng/pipeline/DispatchIssueUnit.hh
```diff
@@ -125,6 +124,19 @@ class DispatchIssueUnit {
...
+   * instruction, but is not permitted to have the result forwarded to them
+   * directly. */
+  std::vector<dependencyEntry> dependantInstructions_;
+
+  /** Contains instructions that have had results forwarded to them, but need to
+   * wait for x-cycles to mimic the in hardware latency of said result
+   * forwarding.
+   * Outer Pair = {ticks_ value when latency has elapsed, Inner Pair}.
+   * Inner Pair = {instruction entry, value it has been forwarded}. */
+  std::vector<std::pair<uint64_t, std::pair<dependencyEntry, RegisterValue>>>
```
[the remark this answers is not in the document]
FinnWilkinson: I agree. I have updated the `waitingInstructions_` data structure to be an unordered map : `std::unordered_map<uint64_t, std::vector<std::pair<dependencyEntry, RegisterValue>>>`

This allows us to "fire off" all instructions in the vector for the current tick count. A map was used instead of a queue due the bypass latency can vary depending on which instruction is recieving the result. This means that cycle to release on will not always be equal to or greater than current entries in the map.

jj16791:
Adding a comment to reinforce that the content contained within this PR is good for the sake of research possibilities. We need to assess if it meets for "flexibility" requirement for this.

FinnWilkinson:
Closing this PR due to being stale - this work should be re-visited in the future with an agnostic way of defining a forwarding map

PR #228, opened 2022-05-12
Missing sys reg handling

jj16791:
This PR contains a fix for handling missing system registers in the aarch64 `systemRegisterMap_` map. A missing entry will return a -1 and a decoded instruction accessing an unmapped system register will raise a new UnmappedSysReg fatal exception.


FinnWilkinson (approved):
All looks good

jj16791:
#rerun tests

PR #231, opened 2022-06-21
Initialise SST integration with SimEng and conditional compilation using cmake.

rahahahat:
This PR implements the `SST` wrapper and `simeng::MemoryInterface` needed for successful integration with SST. It also adds support for conditional compilation through `CMake`. Documentation for running and building SimEng with the SST integration enabled is temporarily available [here](https://www.notion.so/SimEng-SST-Docs-681b295312a64ee8a93d8802d314e7ac). It will be added to the source at a later stage.


rahahahat:
@jj16791 @FinnWilkinson I've just noticed that I've forgotten to document params in the `SimengCoreWrapper.h` file as well as forgotten to acquire `config_path` in the constructor. I will fix that soon

rahahahat:
> @jj16791 @FinnWilkinson I've just noticed that I've forgotten to document params in the `SimengCoreWrapper.h` file as well as forgotten to acquire `config_path` in the constructor. I will fix that soon

Done!

configs/a64fx.yaml
```diff
@@ -12,7 +12,7 @@ Core:
   Micro-Operations: True
   Vector-Length: 512
 Process-Image:
-  Heap-Size: 1073741824
+  Heap-Size: 503741824
```
FinnWilkinson, line 15: Is this change to heap-size required for SST to work?

sst/Makefile
```diff
@@ -0,0 +1,44 @@
+CXX := $(shell sst-config --CXX)
+CXXFLAGS := -std=c++17 -D__STDC_FORMAT_MACROS -fPIC -DHAVE_CONFIG_H -O3 -I/home/rahat/asimov/SST/sst-core-install/include
+LDFLAGS := $(shell sst-config --ELEMENT_LDFLAGS)
+
+SIMENG_ROOT = /home/rahat/asimov/SimEng
+SIMENG_INSTALL := /home/rahat/asimov/simeng-sst-bin
```
FinnWilkinson, line 6: Just to double check, I presume these change with main SimEng CMake? i.e. not hard-coded as (or defaults to) these paths

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,218 @@
...
+    output.verbose(CALL_INFO, 1, 0, "Simulation complete. Finalising stats....\n");
+
+    auto endTime = std::chrono::high_resolution_clock::now();
+    auto duration =
+      std::chrono::duration_cast<std::chrono::milliseconds>(endTime - start_time)
+          .count();
+    auto hz = iterations / (static_cast<double>(duration) / 1000.0);
+    auto khz = hz / 1000.0;
+    auto retired = core->getInstructionsRetiredCount();
+    auto mips = retired / static_cast<double>(duration) / 1000.0;
```
FinnWilkinson, line 69: Would be preferable to decalre the type explicitly instead of using auto for these variables

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,218 @@
...
+    iterations++;
+
+    return false;
+    } else {
+        primaryComponentOKToEndSim();
+        return true;
+    }
+}
+
+void SimengCoreWrapper::fabricateSimengCore() {
```
FinnWilkinson, line 116: As this function looks very similar to what happens in `main.cc`, will it need manually updating if SimEng's `main.cc` is changed?

sst/backward.cc
```diff
@@ -0,0 +1,42 @@
+// Pick your poison.
+//
+// On GNU/Linux, you have few choices to get the most out of your stack trace.
```
FinnWilkinson, line 3: Is there a way to automate this library installation in SST-CMake? And what happens for MacOS and non-GNU compilers?

rahahahat: I will remove this file in our final integration. I needed it for debugging.

sst/SimengCoreWrapper.cc
rahahahat: Yes, I believed we've had this conversation before. I think the plan was to have a fabricate method in Core. But I wasn't tasked with this.

sst/Makefile
rahahahat: This makefile should not be here. I will remove it.

sst/backward.cc
rahahahat: But it is very useful for debugging atleast for SST stuff. I can try and look for automating it if you want.

FinnWilkinson: I will leave this decision up to you. Again it could be something to come back to in the future.

sst/SimengCoreWrapper.cc
FinnWilkinson: OK, this may be something to come back to at a later stage (i.e. not in this PR)

sst/SimengMemInterface.cc
```diff
@@ -0,0 +1,279 @@
...
+        aggrReq->id
+    });
+
+    // Cleanup
+    aggrReq->response_map.clear();
+    delete aggrReq;
+}
+
+void SimengMemInterface::SimengMemHandlers::handle(StandardMem::WriteResp* rsp) {
+    delete rsp;
```
jj16791, line 232: Are we always assuming the write request was successful? Can an SST write request fail? If so, we should be able to handle it and act accordingly

sst/SimengMemInterface.cc
```diff
@@ -0,0 +1,279 @@
...
+}
+
+void SimengMemInterface::SimengMemHandlers::handle(StandardMem::ReadResp* rsp) {
+    uint64_t id = rsp->getID();
+    auto data = rsp->data;
+    delete rsp;
+
+    // Upon recieving a response from SST the aggregation_map is used to retrieve
+    // the AggregatedReadRequest the recieved SST response is a part of.     
+    auto itr = mem_interface.aggregation_map.find(id);
```
jj16791, line 242: what is `mem_interface`? I assumed `aggregation_map` is a reference to the private member variable defined in 'sst/include/SimengMemInterface.hh'. Why isn't it just `auto itr = aggregation_map.find(id);`?

jj16791: Looking at the header file I kind of get it now. This shows the importance of commenting/"simple code", hopefully, the requested comments in the header file (akin to those seen in existing SimEng header files) will shed light on this. A new user should be able to understand all of this with some prior c++ experience

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,125 @@
...
+using namespace simeng;
+using namespace SST::Interfaces;
+
+namespace SST {
+    namespace SSTSimeng {
+        class SimengMemInterface: public MemoryInterface {
+            public:
+                SimengMemInterface(StandardMem* mem, uint64_t cl, uint64_t max_addr, SST::Output* out);
+                void sendProcessImageToSST(const span<char> image);
+                virtual void requestRead(const MemoryAccessTarget& target,
```
jj16791, line 29: Why are these defined as `virtual`? What is overriding it?

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,125 @@
...
+#include <set>
+#include <type_traits>
+
+#include  "simeng/MemoryInterface.hh"
+#include "simeng/span.hh"
+
+using namespace simeng;
+using namespace SST::Interfaces;
+
+namespace SST {
```
jj16791, line 23: Follow the non-indented formatting seen in existing header files

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,125 @@
...
+    namespace SSTSimeng {
+        class SimengMemInterface: public MemoryInterface {
```
jj16791, line 25: Comment explaining the class similar to existing header files

jj16791 (changes requested):
Goo start. A few requests and clarifications are needed. A general comment is the need to add more comments. The commenting style (DOxygen), frequency, a occurrence should follow that seen in existing SimEng files.

sst/SimengCoreWrapper.cc
FinnWilkinson: Didn't realise that was the case. I agree with using `auto` for lengthy type names, however these look like `uints` or `doubles`. Maybe something to change in `main` to improve clarity?

sst/SimengMemInterface.cc
rahahahat: If a request fails SST breaks and the program terminates.

sst/SimengCoreWrapper.cc
jj16791: The time ones might be long and so probably looked nicer to have all as `auto`

rahahahat: Shall I leave it as is?

sst/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+target_compile_options(sstsimeng PUBLIC -g)
+
+# Define statements generated by sst-config --ELEMENT_CXXFLAGS
+# SIMENG_ENABLE_SST is used as a preprocessing directive and is defined by us,
+# without this definition files in simeng/sst won't compile.
+target_compile_definitions(sstsimeng PUBLIC SIMENG_ENABLE_SST __STDC_FORMAT_MACROS HAVE_CONFIG_H)
+
+# Extra link flags generated by sst-config --ELEMENT_LDFLAGS
+# The commmand had to be replaced as some extra flags added using target_include_directories
+target_link_options(sstsimeng PUBLIC -fno-common -Wl,-undefined -Wl,dynamic_lookup)
```
tom91136, line 30: Use `LINKER:-undefined` instead of `-Wl`, see https://cmake.org/cmake/help/latest/command/target_link_options.html#handling-compiler-driver-differences, it works all the way back to CMake 3.13.

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,218 @@
...
+enum class SimulationMode { Emulation, InOrderPipelined, OutOfOrder };
+
+SimengCoreWrapper::SimengCoreWrapper(SST::ComponentId_t id, SST::Params& params): SST::Component(id) {
+    output.init("SimengCoreWrapper[" + getName() + ":@p:@t]:", 999, 0, SST::Output::STDOUT);
+    clock = registerClock(params.find<std::string>("clock", "1GHz"), new SST::Clock::Handler<SimengCoreWrapper>(this, &SimengCoreWrapper::clockTick));
+
+    executable_path = params.find<std::string>("executable_path", "");
+    executable_args = params.find<std::string>("executable_args", "");
+    config_path = params.find<std::string>("config_path", "");
+    cache_line_width  = params.find<uint64_t>("cache_line_width", "64");
```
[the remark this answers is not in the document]
rahahahat: AFAIK you can't have multiple configs in SST. To run a simulation the command `sst config.py` is used. So i'm sure how we can pass multiple configs. I don't think SST can exeucte python logic so we can programtically import it inside `config.py`. However we can write a python script which generates a config.py per specification.

sst/SimengMemInterface.cc
```diff
@@ -0,0 +1,279 @@
...
+    */
+    if (--aggrReq->aggregateCount <= 0) {
+        mem_interface.aggregatedReadResponses(aggrReq);
+    }
+}
+
+int SimengMemInterface::getCacheLinesNeeded(int size) {
+    if (size < clw) return 1;
+    if (size % clw == 0) return size / clw;
+    return (size / clw) + 1;
```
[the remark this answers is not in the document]
rahahahat: For unsigned and non-negative values the result of integer divisionis truncated towards zero, i.e  it rounds towards the smallest absolute value. I'll fix the function definition to have unsigned int.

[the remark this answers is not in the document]
rahahahat: For unsigned and non-negative values the result of integer divisionis truncated towards zero, i.e  it rounds towards the smallest absolute value. I'll fix the function definition to have unsigned int.

sst/SimengMemInterface.cc
```diff
@@ -0,0 +1,279 @@
...
+}
+bool SimengMemInterface::unsignedOverflow_(uint64_t a, uint64_t b) const {
+    return (a + b) < a || (a + b) < b;
+};
+bool SimengMemInterface::requestSpansMultipleCacheLines(uint64_t addrStart, uint64_t addrEnd) {
+    uint64_t lineDiff = (addrEnd / clw) - (addrStart/clw);
```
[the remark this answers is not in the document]
rahahahat: For unsigned and non-negative values the result of integer divisionis truncated towards zero, i.e  it rounds towards the smallest absolute value.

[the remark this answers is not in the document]
rahahahat: For unsigned and non-negative values the result of integer divisionis truncated towards zero, i.e  it rounds towards the smallest absolute value.

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
rahahahat: For unsigned and non-negative values the result of integer divisionis truncated towards zero, i.e  it rounds towards the smallest absolute value. I'll fix the function definition to have unsigned int.

sst/SimengMemInterface.cc
```diff
@@ -0,0 +1,279 @@
...
+    return lineDiff > 0;
+};
+uint64_t SimengMemInterface::nearestCacheLineEnd(uint64_t addrStart) {
+    return (addrStart / clw) + 1;
```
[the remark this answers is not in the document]
rahahahat: For unsigned and non-negative values the result of integer divisionis truncated towards zero, i.e  it rounds towards the smallest absolute value. I'll fix the function definition to have unsigned int.

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
rahahahat: https://stackoverflow.com/questions/3602827/what-is-the-behavior-of-integer-division

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,218 @@
...
+      modeString = "Emulation";
+    }
+
+    float clockFreq_ = config["Core"]["Clock-Frequency"].as<float>();
+    uint32_t timerFreq_ = config["Core"]["Timer-Frequency"].as<uint32_t>();
+    timer_modulo = (clockFreq_ * 1e9) / (timerFreq_ * 1e6);
+
+    // // Create the process Image
+    std::vector<std::string> commandLine({executable_path, executable_args});
+    process = std::make_unique<simeng::kernel::LinuxProcess>(commandLine, config);
```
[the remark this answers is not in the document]
rahahahat: I'll be handling that fix in a separate PR.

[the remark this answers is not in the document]
rahahahat: I'll be handling that fix in a separate PR.

sst/SimengMemInterface.cc
```diff
@@ -0,0 +1,279 @@
...
+using namespace SST::SSTSimeng;
+
+SimengMemInterface::SimengMemInterface(StandardMem* mem, uint64_t cl, uint64_t max_addr, SST::Output* out): simeng::MemoryInterface() {
+    this->mem = mem;
+    this->clw= cl;
+    this->max_addr_memory = max_addr;
+    this->output = out;
+};
+
+void SimengMemInterface::sendProcessImageToSST(const span<char> image) {
```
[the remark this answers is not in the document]
rahahahat: Yes, will be handing that in a separate PR.

sst/SimengCoreWrapper.cc
jj16791: Depends on the above comments. Does it look messy/harder to read if auto is not used?

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: I didn't mean to suggest we have multiple configs. The additional config file was in addition to the SimEng config file. So we'd have a config file for the core and a config file for the memory subsystem. At the moment, you have SST config params as CLI args, but, if there are many possible params we'd like to target, then the amount of CLI args may become cumbersome. It may be better to pass a config file and extract SST args from that (similar to how we do for SimEng).

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
jj16791: Is this behaviour associated with a specific C++ version i.e C++17/C++20?

sst/SimengMemInterface.cc
jj16791: Is that expected behaviour (SST just breaks)? Does it safely break (i.e. exit with an error code or eqv)?

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: Why is this going to be in a separate PR? Given that SSTxSimEng can easily break without this fix I'd say it should belong in the initial PR.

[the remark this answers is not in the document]
rahahahat: The fix is with SimEng code, nothing to do with SST. The code here will remain the same. The only addition is that once data is copied I can just free it.

Also, I was hoping to get that PR merged in before this one and then rebase and merge. wdyt?

sst/SimengMemInterface.cc
rahahahat: It does break safely with an error code. There is nothing we can do to catch it as it is an internal error are we are running simeng inside SST.

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
rahahahat: I'm not sure if I understand you correctly. These config params aren't passed as CLI args, but rather these are passed inside the `config.py` file. Since `SimengCoreWrapper` is an SST component, we identify it in `config.py` by the tag `sstsimeng.simengcore` and then pass parameters to it. Those params are then exposed to us via the params object provided by SST. The only command I run is `sst config.py`. 

```
// sst/config/config.py
cpu = sst.Component("core", "sstsimeng.simengcore")
cpu.addParams({
    "config_path": "/home/rahat/asimov/SimEng/configs/a64fx.yaml",
    "executable_path": "/home/rahat/asimov/simeng-progs/stream",
    "executable_args": "",
    "clock" : "1GHz",
    "max_addr_memory": 1*1024*1024*1024-1,
    "cache_line_width": "64",
})

```
By passing params like this we can also document then via ELI Macros provided by SST:
```
// sst/include/SimengCoreWrapper.hh

class SimengCoreWrapper : public SST::Component {
 ......
  SST_ELI_DOCUMENT_PARAMS(
        { "config_path", "Path to Simeng YAML config file (string)", "" },
        { "executable_path", "Path to executable binary to be run by SimEng (string)", "" },
        { "executable_args", "argument to be passed to the executable binary (string)", "" },
        { "clock", "Clock rate of the SST clock (string)", "" },
        { "max_addr_memory", "Maximum address that memory can access (int)"},
 )
 ........
}
```
Then when we get all information about our custom SST::Component by using the command `sst-info --libs=sstsimeng`:
```
Filtering output on Element = "sstsimeng"
================================================================================
ELEMENT 0 = sstsimeng ()
Num Components = 1
      Component 0: simengcore
      CATEGORY: PROCESSOR COMPONENT
         NUM STATISTICS = 0
         NUM PORTS = 0
         NUM SUBCOMPONENT SLOTS = 0
         NUM PARAMETERS = 5
            PARAMETER 0 = config_path (Path to Simeng YAML config file (string)) []
            PARAMETER 1 = executable_path (Path to executable binary to be run by SimEng (string)) []
            PARAMETER 2 = executable_args (argument to be passed to the executable binary (string)) []
            PARAMETER 3 = clock (Clock rate of the SST clock (string)) []
            PARAMETER 4 = max_addr_memory (Maximum address that memory can access (int)) [<required>]
    simengcore: Simeng core wrapper for SST
    Using ELI version 0.9.0
    Compiled on: Aug 10 2022 10:36:49, using file: /home/rahat/asimov/SimEng/sst/include/SimengCoreWrapper.hh
Num SubComponents = 0
Num Modules = 0
Num SSTPartitioners = 0

```
Is this not the right way to do it? 
Or worse have I completely misunderstood you again?

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
rahahahat: Same for C++20 Standard (https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/n4849.pdf)
image

sst/SimengCoreWrapper.cc
rahahahat: PTAL. I have removed `auto` declarations wherever possible.

rahahahat: PTAL. I have removed `auto` declarations wherever possible.

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,218 @@
...
+    handlers = new SimengMemInterface::SimengMemHandlers(*data_memory, &output);
+
+    registerAsPrimaryComponent();
+    primaryComponentDoNotEndSim();
+}
+
+SimengCoreWrapper::~SimengCoreWrapper() {}
+
+void SimengCoreWrapper::setup() {
+  mem->setup();
```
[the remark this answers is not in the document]
rahahahat: PTAL

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,218 @@
...
+
+    std::cout << "\nFinished " << iterations << " ticks in " << duration << "ms ("
+            << std::round(khz) << " kHz, " << std::setprecision(2) << mips
+            << " MIPS)" << std::endl;
+
+    delete[] process_memory;
+}
+
+void SimengCoreWrapper::init(unsigned int phase) {
+    mem->init(phase);
```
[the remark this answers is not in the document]
rahahahat: PTAL

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,218 @@
...
+      for (size_t j = 0; j < reservation_station["Ports"].size(); j++) {
+        uint8_t port = reservation_station["Ports"][j].as<uint8_t>();
+        if (rsArrangement.size() < port + 1) {
+          rsArrangement.resize(port + 1);
+        }
+        rsArrangement[port] = {i, reservation_station["Size"].as<uint16_t>()};
+      }
+    }
+
+    modeString = "Out-of-Order";
```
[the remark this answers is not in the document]
rahahahat: PTAL, I've added support for all modes.

sst/SimengMemInterface.cc
```diff
@@ -0,0 +1,279 @@
...
+        std::vector<StandardMem::Request*> rvec1 = splitAggregatedRequest(aggrReq, addrStart, firstFragmentSize);
+        std::vector<StandardMem::Request*> rvec2 = splitAggregatedRequest(aggrReq, cacheLineEndAddr, secondFragmentSize);
+        reqs.insert(reqs.end(), rvec1.begin(), rvec1.end());
+        reqs.insert(reqs.end(), rvec2.begin(), rvec2.end());
+        return reqs;
+    }
+    return splitAggregatedRequest(aggrReq, addrStart, size);
+}
+
+std::vector<StandardMem::Request*> SimengMemInterface::splitAggregatedRequest(
```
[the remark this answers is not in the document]
rahahahat: PTAL, removed the overloaded base instance.

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
rahahahat: PTAL, updated `getCacheLinesNeeded` to `uint64_t`

sst/include/SimengCoreWrapper.hh
```diff
@@ -0,0 +1,109 @@
...
+using namespace simeng;
+
+namespace SST {
+    namespace SSTSimeng {
+        class SimengCoreWrapper: public SST::Component {
+            public:
+                SimengCoreWrapper(SST::ComponentId_t id, SST::Params& params);
+                ~SimengCoreWrapper();
+
+                void setup();
```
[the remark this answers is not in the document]
rahahahat: PTAL

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,125 @@
...
+            public:
+                SimengMemInterface(StandardMem* mem, uint64_t cl, uint64_t max_addr, SST::Output* out);
```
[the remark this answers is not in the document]
rahahahat: PTAL

sst/SimengMemInterface.cc
rahahahat: PTAL

sst/include/SimengMemInterface.hh
rahahahat: PTAL

sst/CMakeLists.txt
rahahahat: PTAL

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
rahahahat: We can support emulation mode and out-of-order mode, but not inorder mode because it requires a FlatMemoryInterface& as a constructor argument. Should inorder core constructor be changed to accept MemoryInterface?

sst/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+# without this definition files in simeng/sst won't compile.
+target_compile_definitions(sstsimeng PUBLIC SIMENG_ENABLE_SST __STDC_FORMAT_MACROS HAVE_CONFIG_H)
+
+# Extra link flags generated by sst-config --ELEMENT_LDFLAGS
+# The commmand had to be replaced as some extra flags added using target_include_directories
+target_link_options(sstsimeng PUBLIC -fno-common LINKER:-undefined,dynamic_lookup)
+
+
+target_link_libraries(sstsimeng PUBLIC libsimeng yaml-cpp dl bfd)
+
```
jj16791, line 34: Too many new lines. Could replace with comments to help separate out regions of this file like above

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: Okay no, this is good, I understand now. So we'd have to alter this config.py file when we want to alter the memory hierarchy we are simulating within SST?

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: Good comments but move to header file to match existing header file formatting.

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: Good comments but move to header file to match existing header file formatting.

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: As far as I was aware, the fix for SST was different from the new memoryImage PR you have. I thought, for the SST integration, we were moving away from constructing a memoryImage on the SimEng side and constructing the memoryImage on the SST side through various write requests.

sst/SimengMemInterface.cc
jj16791: I think we should action a f2f talk about this regarding the mem_interface/friend class setup. I understand that it must all work but better in-code clarity is required.

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
jj16791: Looks good to me. As long as you're confident the rounding down is consistent for C++17 then I'm happy.

sst/include/SimengCoreWrapper.hh
```diff
@@ -0,0 +1,113 @@
...
+using namespace SST;
+using namespace SST::Interfaces;
+using namespace SST::SSTSimeng;
+using namespace simeng;
+
+namespace SST {
+
+namespace SSTSimeng {
+
+class SimengCoreWrapper: public SST::Component {
```
jj16791, line 49: Lacking appropriate commenting

sst/CMakeLists.txt
tom91136: LGTM

FinnWilkinson:

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
rahahahat: Yes only this file.

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
rahahahat: But what about classes in SimEng that require processImage?

sst/SimengMemInterface.cc
rahahahat: yeah sure sounds good.

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.13)
 include(FetchContent)
 set(FETCHCONTENT_QUIET OFF)
-
+set(CMAKE_VERBOSE_MAKEFILE ON)
```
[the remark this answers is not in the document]
rahahahat: This just spits out all compiled commands so I can inspect them, I can remove it if you want. I put this in when I was facing compilation problems with SST.

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: The inorder core needs an update but I think that may come with Dan's RISCV work so leave for now. However, if this PR will only support SST+OoO core then we'll need to prevent the simulation from starting if the passed config file doesn't use the OoO core.

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: What are those classes?

sst/SimengCoreWrapper.cc
jj16791: Which ones?

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
rahahahat: Shall I resolve this comment given that the fix is handled in a separate PR

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
rahahahat: Just a quick note about this, even after the mem-optimisation fix. This needs to be done because the process image still needs to be constructed inside LinuxProcess. After the initial copy is completed I can then free the processImage reference inside SimEng. Unless you want something else?

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,236 @@
+// DO NOT MOVE FROM TOP OF FILE
+#include <sst/core/sst_config.h>
```
jj16791, line 2: This fails the format check but we need to put this at the top of the file apparently. Not sure what the fix will be.

sst/include/SimengCoreWrapper.hh
```diff
@@ -0,0 +1,141 @@
...
+       "Path to executable binary to be run by SimEng (string)", ""},
+      {"executable_args",
+       "argument to be passed to the executable binary (string)", ""},
+      {"clock", "Clock rate of the SST clock (string)", ""},
+      {"max_addr_memory", "Maximum address that memory can access (int)"}, )
+
+ private:
+  // SST properties
+  SST::Output output;
+  TimeConverter* clock;
```
jj16791, line 110: All the below needs comments akin to the existing SimEng header files

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,210 @@
...
+    std::map<uint64_t, std::vector<uint8_t>> response_map;
+    int aggregateCount = 0;
+
+    AggregateReadRequest() : SimengMemoryRequest(), id(0){};
+    AggregateReadRequest(const MemoryAccessTarget& target, const uint64_t id)
+        : SimengMemoryRequest(target), id(id) {}
+  };
+
+ private:
+  SST::Output* output;
```
jj16791, line 137: All the below needs comments akin to the existing SimEng header files

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,210 @@
...
+   */
+  struct SimengMemoryRequest {
+    const MemoryAccessTarget target;
+
+    SimengMemoryRequest() : target(MemoryAccessTarget()){};
+    SimengMemoryRequest(const MemoryAccessTarget& target) : target(target){};
+  };
+
+  /**
+   * Structs AggregatedWriteRequest and AggregatedReadRequest are used to store
```
jj16791, line 106: Split into two separate DOXYGEN comments for these structs. Having it like this will only link this comment to the top struct and the bottom struct won't have anything.

jj16791 (changes requested):
Some comments. Still missing code comments for header file variables/incorrect comments for some functions. The point of the comments/the styling is that it'll work for DOXYGEN in the future so please follow what you can see in the existing header files.

sst/SimengCoreWrapper.cc
jj16791: Resolving as it looks fine.

sst/include/SimengCoreWrapper.hh
```diff
@@ -0,0 +1,143 @@
...
+using namespace SST;
+using namespace SST::Interfaces;
+using namespace SST::SSTSimeng;
+using namespace simeng;
+
+namespace SST {
+
+namespace SSTSimeng {
+
+class SimengCoreWrapper : public SST::Component {
```
jj16791, line 51: As you did with `sst/include/SimengMemInterface.hh`, needs a comment explaining the class similar to existing header files

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,238 @@
...
+  simeng::SpecialFileDirGen SFdir = simeng::SpecialFileDirGen(config);
+  // Create the Special Files directory if indicated to do so in Config
+  if (config["CPU-Info"]["Generate-Special-Dir"].as<std::string>() == "T") {
+    // Remove any current special files dir
+    SFdir.RemoveExistingSFDir();
+    // Create new special files dir
+    SFdir.GenerateSFDir();
+  }
+
+  data_memory->sendProcessImageToSST(processImage);
```
jj16791, line 232: Destroy after use.

sst/config/config.py
```diff
@@ -0,0 +1,61 @@
...
+DEBUG_LEVEL = 0
+
+# Define the simulation components
+cpu = sst.Component("core", "sstsimeng.simengcore")
+cpu.addParams({
+    "config_path": "/home/rahat/asimov/SimEng/configs/a64fx.yaml",
+    "executable_path": "/home/rahat/asimov/simeng-progs/stream",
+    "executable_args": "",
+    "clock" : "1GHz",
+    "max_addr_memory": 1*1024*1024*1024-1,
```
jj16791, line 14: Shouldn't this be implicit from the size of HEAP and STACK defined in the model config?

CMakeLists.txt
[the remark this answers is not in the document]
jj16791: This needs to be removed

sst/SimengMemInterface.cc
```diff
@@ -0,0 +1,287 @@
...
+
+  StandardMem::Request* req = new StandardMem::Write(0, data.size(), data);
+  mem_->sendUntimedData(req);
+  return;
+};
+
+template <
+    typename T,
+    typename std::enable_if<std::is_base_of<
+        SimengMemInterface::SimengMemoryRequest, T>::value>::type* = nullptr>
```
jj16791, line 37: remove redefinition of default argument (already defined in header file) as it causes an error when compiled with Apple Clang i.e. `::type* = nullptr` -> `::type*`.

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,253 @@
...
+  float clockFreq_ = config["Core"]["Clock-Frequency"].as<float>();
+  uint32_t timerFreq_ = config["Core"]["Timer-Frequency"].as<uint32_t>();
+  timerModulo_ = (clockFreq_ * 1e9) / (timerFreq_ * 1e6);
+
+  // Create the process Image
+  std::vector<std::string> commandLine({executablePath_, executableArgs_});
+  process_ =
+      std::make_unique<simeng::kernel::LinuxProcess>(commandLine, config);
+  if (!process_->isValid())
+    output_.fatal(CALL_INFO, 1, 0, "Could not read/parse %s", executablePath_);
```
jj16791, line 165: To fix the AppleClang error `error: cannot pass object of non-trivial type 'std::string' (aka 'basic_string<char, char_traits<char>, allocator<char> >') through variadic method; call will abort at runtime [-Wnon-pod-varargs]`, change `executablePath_` to `executablePath_.c_str()`

jj16791 (changes requested):
Some fixes for AppleClang

jj16791 (commented):
Looks good. Just need to finalise the yaml-cpp issue now.

configs/m1_firestorm-sst.yaml
```diff
@@ -0,0 +1,307 @@
+# M1 Firestorm core
```
rahahahat, line 1: Would it be better to put all sst related config files in the sst folder itself?

configs/tx2-sst.yaml
```diff
@@ -0,0 +1,185 @@
+---
```
rahahahat, line 1: Would it be better to put all sst related config files in the sst folder itself?

src/lib/CoreInstance.cc
```diff
@@ -175,6 +178,9 @@ void CoreInstance::createL1DataMemory(const simeng::MemInterfaceType type) {
 
 void CoreInstance::setL1DataMemory(
     std::shared_ptr<simeng::MemoryInterface> memRef) {
+  assert(setDataMemory_ &&
```
rahahahat, line 181: For some reason this assertion is not firing in my SST Build, I changed the function to:
```
void CoreInstance::setL1DataMemory(
    std::shared_ptr<simeng::MemoryInterface> memRef) {
  std::cout << "Set Data Memory: " << setDataMemory_ << std::endl;
  assert(setDataMemory_ &&
         "setL1DataMemory(...) called but the interface was created by the "
         "CoreInstance class.");
  // Set the L1D cache instance to use
  dataMemory_ = memRef;
  return;
}
```
The output I get is below and the simulation still continues. Can you please confirm that this happens to you as well.
image

I encountered this because the config.py in the sst folder still has the path to `a64fx.yaml` and not to `a64fx-sst.yaml`. So the assertion should've failed however it doesn't.

src/lib/CoreInstance.cc
```diff
@@ -149,6 +149,9 @@ void CoreInstance::createL1InstructionMemory(
 
 void CoreInstance::setL1InstructionMemory(
     std::shared_ptr<simeng::MemoryInterface> memRef) {
+  assert(setInstructionMemory_ &&
```
rahahahat, line 152: Same as the comment for setL1DataMemory.

sst/config/config.py
```diff
@@ -0,0 +1,62 @@
+import sst
+
+DEBUG_L1 = 0
+DEBUG_MEM = 0
+DEBUG_LEVEL = 0
+
+# Define the simulation components
+cpu = sst.Component("core", "sstsimeng.simengcore")
+cpu.addParams({
+    "config_path": "/home/rahat/asimov/SimEng/configs/a64fx.yaml",
```
rahahahat, line 10: change config path to a64fx-sst.config.py

sst/SimengMemInterface.cc
```diff
@@ -200,7 +200,7 @@ void SimengMemInterface::clearCompletedReads() {
 }
 
 bool SimengMemInterface::hasPendingRequests() const {
-  return completedReadRequests_.size() > 0;
+  return aggregationMap_.size() > 0;
```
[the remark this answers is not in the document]
jj16791: Just double checking, every memory request will have a `aggregationMap_`?

sst/SimengCoreWrapper.cc
```diff
@@ -127,52 +117,52 @@ bool SimengCoreWrapper::clockTick(SST::Cycle_t current_cycle) {
   }
 }
 
+std::vector<std::string> SimengCoreWrapper::splitArgs(std::string argString) {
```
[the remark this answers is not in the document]
jj16791: Whilst this is a good idea generally, maybe we should re-evaluate the custom delimiters. If we deem `[` or `]` to be valid arguments then we need to support all possible permutations. It may be worth us outputting what workload and args are being used in the sim. Perhaps inline or below the "Running in x mode"

configs/m1_firestorm-sst.yaml
jj16791: Resolving as it's a duplicate comment that just bloats the PR.

configs/a64fx-sst.yaml
```diff
@@ -0,0 +1,265 @@
+---
```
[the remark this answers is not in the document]
jj16791: As long as it's documented, we could. Adds more complexity but somewhat makes sense. Thoughts @FinnWilkinson @dANW34V3R.

sst/config/config.py
jj16791: Left unresolved just in case we change location. We should also change some of these config params to purposely fail to ensure a new user is running what they mean to.

src/lib/CoreInstance.cc
jj16791: You've built with release mode. No assertions will trigger. Please rebuild and re-try.

src/lib/CoreInstance.cc
jj16791: Comment left on the referenced comment. It looks like user error so resolving this duplicate.

sst/include/SimengCoreWrapper.hh
```diff
@@ -0,0 +1,187 @@
...
+
+  /**
+   * This is the finish lifecycle method present in all SST::Components.
+   * Here it is overriden to finish statistics about the SimEng simulation.
+   */
+  void finish();
+
+  /**
+   * The clockTick is a method present in all SST::Components. This fuction
+   * is called everytime the SST clock ticks. The current clock cylce is passed
```
FinnWilkinson, line 75: typo at end of line "cylce"

jj16791 (approved):

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
jj16791: If not, we'll need a container that supports this so we can correctly provide the functionality of `hasPendingRequests`.

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,248 @@
...
+  void aggregatedReadResponses(AggregateReadRequest* aggrReq);
+
+  /** Get the number of cache lines needed incase the size of a memory request
+   * is larger than cache line width.
+   */
+  int getCacheLinesNeeded(uint64_t size) const;
+  bool unsignedOverflow_(uint64_t a, uint64_t b) const;
+
+  /**
+   * check to see if a request span multiple cache lines. This methods
```
FinnWilkinson, line 230: should be "Check" instead of "check", and "method" not "methods"

FinnWilkinson (changes requested):
A few minor comments, but generally looks good

configs/a64fx-sst.yaml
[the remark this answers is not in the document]
FinnWilkinson: I think seperating the models via "type" may be useful inside the current `configs` directory. Perhaps have standard `a64fx.yaml` and `tx2.yaml` in the root, then have a folder for `sst-models`? This method would also be useful in future for hypothetical models (i.e. RISC-V, AArch64+SME, etc...)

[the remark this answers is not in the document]
jj16791: I second Finn's idea of fragmenting the config file into multi-sub directories

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,193 @@
...
+                  "Simulation complete. Finalising stats....\n");
+
+  auto endTime = std::chrono::high_resolution_clock::now();
+  auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(
+                      endTime - startTime_)
+                      .count();
+  double hz = iterations_ / (static_cast<double>(duration) / 1000.0);
+  double khz = hz / 1000.0;
+  uint64_t retired = core_->getInstructionsRetiredCount();
+  double mips = retired / static_cast<double>(duration) / 1000.0;
```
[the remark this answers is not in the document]
jj16791: Good spot

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,193 @@
...
+void SimengCoreWrapper::finish() {
+  output_.verbose(CALL_INFO, 1, 0,
+                  "Simulation complete. Finalising stats....\n");
+
+  auto endTime = std::chrono::high_resolution_clock::now();
+  auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(
+                      endTime - startTime_)
+                      .count();
+  double hz = iterations_ / (static_cast<double>(duration) / 1000.0);
+  double khz = hz / 1000.0;
```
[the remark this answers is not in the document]
jj16791: An artefact of old `main.cc` logic which this was copied from. Made a change in both files

sst/config/config.py
```diff
@@ -0,0 +1,62 @@
...
+    "executable_path": "/home/rahat/asimov/simeng-progs/stream",
```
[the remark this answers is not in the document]
jj16791: Changed to text that will always error

sst/SimengMemInterface.cc
```diff
@@ -0,0 +1,286 @@
...
+    uint64_t cacheLineEndAddr = nearestCacheLineEnd(addrStart) * clw_;
+    uint64_t firstFragmentSize = cacheLineEndAddr - addrStart;
+    uint64_t secondFragmentSize = size - firstFragmentSize;
+    std::vector<StandardMem::Request*> rvec1 =
+        splitAggregatedRequest(aggrReq, addrStart, firstFragmentSize);
+    std::vector<StandardMem::Request*> rvec2 =
+        splitAggregatedRequest(aggrReq, cacheLineEndAddr, secondFragmentSize);
+    reqs.insert(reqs.end(), rvec1.begin(), rvec1.end());
+    reqs.insert(reqs.end(), rvec2.begin(), rvec2.end());
+    return reqs;
```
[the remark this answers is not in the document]
rahahahat: That case is already handled by splitAggregated request. This just handles the initial overflow (if it exists)

configs/a64fx-sst.yaml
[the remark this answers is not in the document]
rahahahat: Sounds good to me

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
rahahahat: I agree, outputting args is a good idea.  Do you think having an escape character for `[ ]` would solve our issue or will it make it too complicated?

sst/config/config.py
```diff
@@ -0,0 +1,62 @@
+import sst
+
+DEBUG_L1 = 0
+DEBUG_MEM = 0
+DEBUG_LEVEL = 0
+
+# Define the simulation components
+cpu = sst.Component("core", "sstsimeng.simengcore")
+cpu.addParams({
+    "config_path": "<PATH TO CONFIG .YAML FILE>",
```
FinnWilkinson, line 10: Given that SST also has config files, I think it would make things 100% clear if we refer to `config` as the current framework's config (in SimEng code `config` means core model config files, in SST `config` refers to the `.py` file) and then verbosely name the other framework's config files (in SST name a SimEng-config as `simeng_config_path` for example, like-wise in SimEng for SST configs)

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
rahahahat: The comment above says `addrEnd can be multiple cache-lines ahead of addrStart`

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
rahahahat: Any update on this?

sst/config/config.py
rahahahat: I agree. Good point.

FinnWilkinson: I think so yes, although I think just `simeng_config_path` would suffice and would reduce the variable's name length

sst/include/SimengCoreWrapper.hh
```diff
@@ -0,0 +1,187 @@
...
+   */
+  SST_ELI_DOCUMENT_PARAMS(
+      {"config_path", "Path to Simeng YAML config file (string)", ""},
+      {"executable_path",
+       "Path to executable binary to be run by SimEng (string)", ""},
+      {"executable_args",
+       "argument to be passed to the executable binary (string)", ""},
+      {"clock", "Clock rate of the SST clock (string)", ""},
+      {"max_addr_memory", "Maximum address that memory can access (int)"}, )
+
```
rahahahat, line 109: I have forgotten to document `cache_line_width`  via ELI. Can you please add an entry for it? Thanks!

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
jj16791: So, when a write request gets sent to the data mem interface, does it go straight to the SST backend or are there cycles in which is resides within the interface similar to fixed memory interface?

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: Yeh I think so. Then we can have errored states for []] but not [\]]

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
FinnWilkinson: Yup, I see this now. Thanks for clarifying

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
rahahahat: It gets sent straight to SST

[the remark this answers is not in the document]
rahahahat: If simple fix to this would be to write requests to aggregation map and rather than aggregating the response, we just delete it from the Map.

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
rahahahat: Sorry, I don't see how both those are different.

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
jj16791: If it gets fired off to SST straight away then we'll be fine as it's not a pending request

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: updated comment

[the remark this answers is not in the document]
jj16791: I think the following should give us the desired functionality:
```
std::vector<std::string> SimengCoreWrapper::splitArgs(std::string argString) {
  int brCount = 0;
  std::string newArg = "";
  std::vector<std::string> arguments = {};

  // Using a custom delimiter, split the argString string into individual
  // arguments and collate in a vector
  for (int c = 0; c < argString.length(); c++) {
    // Recored starting delimiter
    if (argString.at(c) == '[') {
      brCount++;
    } else if (argString.at(c) == ']') {
      // Ensure ending delimiter has a matching start
      if (brCount <= 0) {
        std::cerr << "Invalid argument format, mismatch on delimiters"
                  << std::endl;
        exit(1);
        return {};
      } else {
        arguments.push_back(newArg);
        newArg = "";
      }
      brCount--;
    } else if (argString.at(c) == '\\') {
      // Identified escape character, read next character as an argument
      // character
      c++;
      newArg += argString.at(c);
    } else {
      newArg += argString.at(c);
    }
  }

  // Ensure all starting delimiters have been matched
  if (brCount != 0) {
    std::cerr << "Invalid argument format, unmatched delimiters" << std::endl;
    exit(1);
    return {};
  }

  return arguments;
}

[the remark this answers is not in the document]
Comment (jj16791) [reply sst/SimengCoreWrapper.cc:120]: Omitting the fix for outputting workload information as it will be contained in an upcoming PR

**Review by jj16791 — CHANGES_REQUESTED** [review]
We need to change all instances of `Simeng` to `SimEng` as the former is incorrect

Do you mean for comments or for variable names, filenames as well?

File: sst/SimengCoreWrapper.cc (hunk 42 above)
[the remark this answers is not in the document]
Comment (rahahahat) [reply sst/SimengCoreWrapper.cc:120]: This looks good! 
Is there any way we could test it? (if we need do)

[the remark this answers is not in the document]
Comment (rahahahat) [reply sst/SimengCoreWrapper.cc:120]: The inner if can be refactored like this:
```
 else if (argString.at(c) == ']') {
      // Ensure ending delimiter has a matching start
      if (brCount <= 0) {
        std::cerr << "Invalid argument format, mismatch on delimiters"
                  << std::endl;
        exit(1);
      } 
      arguments.push_back(newArg);
      newArg = "";
      brCount--;
    }
```

File: sst/SimengMemInterface.cc [hunk 52]
```diff
@@ -0,0 +1,286 @@
...
+  aggrReq->responseMap_.clear();
+  delete aggrReq;
+}
+
+void SimengMemInterface::SimengMemHandlers::handle(
+    StandardMem::WriteResp* rsp) {
+  delete rsp;
+}
+
+void SimengMemInterface::SimengMemHandlers::handle(StandardMem::ReadResp* rsp) {
```
dANW34V3R, line 241: Is overloading necessary here? Would it be more readable to have two separate function names

sst/SimengCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: > This looks good! Is there any way we could test it? (if we need do)

Other than manually, I don't know where we'd test it other than your framework

sst/SimengMemInterface.cc
dANW34V3R: Also not a fan of "handle" as it isn't very descriptive

sst/include/SimengCoreWrapper.hh
```diff
@@ -0,0 +1,188 @@
...
+  SimengCoreWrapper(SST::ComponentId_t id, SST::Params& params);
+  ~SimengCoreWrapper();
+
+  /** SST lifecycle methods (in-order of invocation) overriden from
+   * SST::Component. */
+
+  /**
+   * This is the init lifecycle method present in all SST::Components.
+   * Here it is overriden to include init calls to all other SST::Components
+   * which are contained inside SimengCoreWrapper. It is neccessary call all
```
dANW34V3R, line 54: "neccessary *to* call"

dANW34V3R: And below

sst/include/SimengCoreWrapper.hh
```diff
@@ -0,0 +1,188 @@
...
+   * as an argument by SST. The SimEng core ticks in this method.
+   */
+  bool clockTick(SST::Cycle_t currentCycle);
+
+  /**
+   * This handle event method is registered to StandardMem interface is called
+   * everytime a memory request is forwarded by the interface. This functions
+   * acts a callback and invokes SimengMemHandlers on the memory request.
+   */
+  void handleEvent(StandardMem::Request* ev);
```
dANW34V3R, line 85: Grammar

sst/include/SimengCoreWrapper.hh
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * SST clock for the component register with the custom component
+   * during instantiation using the registerClock method provided
+   * by SST.
+   */
+  TimeConverter* clock_;
+
+  /**
+   * SST::Interfaces::StandardMem interface responsible for convering
```
dANW34V3R, line 134: "converting"

sst/include/SimengCoreWrapper.hh
```diff
@@ -0,0 +1,188 @@
...
+  // Simeng properties
+  /** Reference to the CoreInstance class responsible for creating the core to
+   * be simulated. */
+  std::unique_ptr<simeng::CoreInstance> coreInstance_;
+
+  /** Reference to SimEng core. */
+  std::shared_ptr<simeng::Core> core_;
+
+  /** Path to the YAML configuration file for SimEng. */
+  std::string configPath_;
```
dANW34V3R, line 149: "simeng_config_path"?

sst/include/SimengCoreWrapper.hh
```diff
@@ -0,0 +1,188 @@
...
+  std::vector<std::string> executableArgs_;
+
+  /** The cache line width for SST. */
+  uint64_t cacheLineWidth_;
+
+  /** Maximum address availbale to SimEng for memory purposes. */
+  uint64_t maxAddrMemory_;
+
+  /** Reference to the SimEng LinuxProcess. */
+  std::shared_ptr<simeng::kernel::LinuxProcess> process_;
```
dANW34V3R, line 164: "LinuxProcess_"?

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,213 @@
...
+}
+
+SimengCoreWrapper::~SimengCoreWrapper() {}
+
+void SimengCoreWrapper::setup() {
+  mem_->setup();
+  output_.verbose(CALL_INFO, 1, 0, "Memory setup complete\n");
+}
+
+void SimengCoreWrapper::handleEvent(StandardMem::Request* ev) {
```
dANW34V3R, line 63: What does "ev" stand for?

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,213 @@
...
+  coreInstance_->setL1DataMemory(dataMemory_);
+
+  // Construct core
+  coreInstance_->createCore();
+
+  // Get remaining simulation objects needed to forward simulation
+  core_ = coreInstance_->getCore();
+  instructionMemory_ = coreInstance_->getInstructionMemory();
+
+  // This check ensure that SST has enough memory to store the entire
```
dANW34V3R, line 183: grammar

sst/SimengCoreWrapper.cc
```diff
@@ -0,0 +1,213 @@
...
+  // processImage constructed by SimEng.
+  if (maxAddrMemory_ < coreInstance_->getProcessImageSize()) {
+    output_.verbose(
+        CALL_INFO, 1, 0,
+        "Error: SST backend memory is less than processImage size. Please "
+        "increase the memory allocated to memHierarchy.memBackend and "
+        "ensure it is consistent with \'max_addr_memory\' and "
+        "\'addr_range_end\'. \n");
+    primaryComponentOKToEndSim();
```
dANW34V3R, line 192: Do you need to do this every time before exiting? Just wondering if this is the reason why I still have an sst process running after the simulation ends

dANW34V3R (changes requested):
The main concept looks good. Comments mainly about spelling and grammar as well as code readability

src/lib/CoreInstance.cc
```diff
@@ -149,6 +149,9 @@ void CoreInstance::createL1InstructionMemory(
 
 void CoreInstance::setL1InstructionMemory(
```
dANW34V3R, line 150: Is there a comment anywhere explaining why coreInstance is needed e.g. because of SST. If not it would be useful to explain this design decision e.g. at the top of coreInstance.hh

rahahahat: I think @Jack said documentation will be added later, if i remember correctly

sst/SimengCoreWrapper.cc
rahahahat: shorthand for ev. I can change it something else?

sst/SimengCoreWrapper.cc
rahahahat: Yes! This method signals SST to stop the simulation.

sst/SimengMemInterface.cc
```diff
@@ -0,0 +1,286 @@
...
+  return requests;
+}
+
+std::vector<StandardMem::Request*> SimengMemInterface::splitAggregatedRequest(
+    AggregateReadRequest* aggrReq, uint64_t addrStart, uint64_t size) {
+  std::vector<StandardMem::Request*> requests;
+  std::vector<uint64_t> req_ids;
+  // Get the number of cache-lines needed to read the data requested by the read
+  // request.
+  int cacheLinesNeeded = getCacheLinesNeeded(size);
```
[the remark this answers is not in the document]
rahahahat: Agreed.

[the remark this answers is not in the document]
rahahahat: Although I think the comment should still be there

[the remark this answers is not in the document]
rahahahat: It gives more context in my opinion

sst/SimengMemInterface.cc
rahahahat: The overloaded definitions are necessary as this class inherits from `StandardMem::RequestHandler` which exposes two virtual overloaded instances of `handle(...)`. Instance of `SimengMemHandlers` is registered as a polymorphic base in `SimengCoreWrapper`. For this we don't have any control over the naming convention.

sst/include/SimengCoreWrapper.hh
rahahahat: Should it be capitalised? wouldn't `linuxProcess_` suffice?

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,248 @@
...
+                            SimengMemoryRequest, T>::value>::type* = nullptr>
+  std::vector<StandardMem::Request*> makeSSTRequests(T* aggrReq,
+                                                     uint64_t addrStart,
+                                                     uint64_t addrEnd,
+                                                     uint64_t size);
+
+  /** The overloaded instance of splitAggregatedRequest is used to split an
+   * AggregatedWriteRequest into multiple SST write requests.
+   */
+  std::vector<StandardMem::Request*> splitAggregatedRequest(
```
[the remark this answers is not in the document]
rahahahat: The mechanism for splitting is so similar it won't make sense to have two different methods? Not sure if it does compromise readability as we are splitting an `AggregatedRequest` in both cases.

Also if we don't overload then we need extra branching logic in `makeSSTRequests` which will make the code quite messy, in my opinion.

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,248 @@
...
+      AggregateReadRequest* aggrReq, uint64_t addrStart, uint64_t size);
+
+  /** This method is used to aggregate responses from multiple read request into
+   * one response. */
+  void aggregatedReadResponses(AggregateReadRequest* aggrReq);
+
+  /** Get the number of cache lines needed incase the size of a memory request
+   * is larger than cache line width.
+   */
+  int getCacheLinesNeeded(uint64_t size) const;
```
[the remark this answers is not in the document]
rahahahat: Agreed

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,248 @@
...
+   * SST::StandardMem::Request(s) into SST memory events to be passed
+   * down the memory heirarchy.
+   */
+  StandardMem* mem_;
+
+  /** Counter for clock ticks. */
+  uint64_t tickCounter_ = 0;
+
+  /** The cache line width specified by SST config.py. */
+  uint64_t clw_;
```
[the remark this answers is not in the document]
rahahahat: Agreed.

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,248 @@
...
+   * also be configured to output information like line-number and filename.
+   */
+  SST::Output* output_;
+
+  /**
+   * SST::Interfaces::StandardMem interface responsible for converting
+   * SST::StandardMem::Request(s) into SST memory events to be passed
+   * down the memory heirarchy.
+   */
+  StandardMem* mem_;
```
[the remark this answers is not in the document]
rahahahat: Not sure if we should capitalise member names

sst/include/SimengMemInterface.hh
```diff
@@ -0,0 +1,248 @@
...
+        : SimengMemoryRequest(target), id_(id) {}
+  };
+
+ private:
+  /**
+   * SST defined output class used to output information to standard output.
+   * This class has in-built methods for different levels of severity and can
+   * also be configured to output information like line-number and filename.
+   */
+  SST::Output* output_;
```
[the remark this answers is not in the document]
rahahahat: Good spot. Removed.

sst/include/SimengCoreWrapper.hh
rahahahat: Turns out this isn't being used, so removed it.

sst/include/SimengCoreWrapper.hh
rahahahat: corrected.

sst/include/SimengCoreWrapper.hh
rahahahat: corrected.

sst/include/SimengCoreWrapper.hh
rahahahat: corrected.

sst/include/SimengCoreWrapper.hh
rahahahat: corrected.

sst/SimengCoreWrapper.cc
rahahahat: corrected.

FinnWilkinson (approved):
All looks good to me

sst/SimengCoreWrapper.cc
dANW34V3R: "ensures"

sst/SimengCoreWrapper.cc
dANW34V3R: So if somewhere else within SimEng exit is called does OKToEndSim need to be called?

sst/SimengMemInterface.cc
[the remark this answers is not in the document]
dANW34V3R: Happy with comment staying as well

sst/SimengMemInterface.cc
dANW34V3R: Sure, thought this may be the case

sst/include/SimengMemInterface.hh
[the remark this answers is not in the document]
dANW34V3R: Sure, but something along those lines may help with readability and understanding

sst/SimengCoreWrapper.cc
rahahahat: corrected.

sst/SimengCoreWrapper.cc
rahahahat: No, it doesn't, as that would terminate SST since the simulation runs inside SST.

sst/include/SimengMemInterface.hh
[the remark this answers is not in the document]
rahahahat: corrected.

rahahahat:
> We need to change all instances of `Simeng` to `SimEng` as the former is incorrect

@jj16791 do you mean change it for all variable names, namespaces, filenames etc?

sst/include/SimengCoreWrapper.hh
rahahahat: Shouldn't this be resolved?

sst/include/SimengCoreWrapper.hh
rahahahat: resolve?

sst/SimengCoreWrapper.cc
rahahahat: resolve?

sst/SimengCoreWrapper.cc
rahahahat: resolve?

sst/SimengCoreWrapper.cc
rahahahat: any updates on this?

src/lib/CoreInstance.cc
rahahahat: and updates on this?

sst/include/SimengMemInterface.hh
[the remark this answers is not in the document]
rahahahat: any updates on this?

sst/SimengMemInterface.cc
```diff
@@ -0,0 +1,286 @@
...
+  for (uint64_t id : req_ids) {
+    /*
+        Insert a key-value pair of SST request id and AggregatedReadRequest
+       reference in the aggregation map. These key-value pairs will later be
+       used to store read response data recieved from SST. This models a
+       many-to-one relation between multiple SST requests and a SimEng read
+       request.
+    */
+    aggregationMap_.insert({id, aggrReq});
+  }
```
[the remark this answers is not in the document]
rahahahat: Done.

jj16791 (approved):

sst/include/SimengMemInterface.hh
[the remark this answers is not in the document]
jj16791: Agree with Rahat on this. Resolving.

src/lib/CoreInstance.cc
jj16791: Original comment out of scope.

PR #233, opened 2022-06-25
Additional aarch64 and syscall support

tomhepworth:
Adds some instructions, syscalls, neon helpers. Includes tests unless functionality was omitted. I just implemented the features necessary for one of my benchmarks to compile and run properly. 

Added instructions + tests:
- LD1v16b
- LD2v16b_POST
- CMHSv16i8
- UMAXPv16i8
- UMINPv16i8
- CMEQv8i8
- BICv8i16
- BRK
- ORRv8i8
- FADDv2f32
- FSUBv2f32
- FADDPv2f32
- Vector TBL variants
- Vector UZP<1,2> variants
- Vector REV16,32,64
- MVNI Masking shift variant
- LD2Twov4s

Added Syscalls:
- getrandom
- tgkill (functionlity omitted as simeng only has one thread)
- shutdown (functionality omitted, mimics behaviour on isembard)
- rseq

Added NEON helper functions:
- vecUMaxP
- vecUMinP
- vecTBL
- vecUzp
- vecRev

Additionally, fixed metadata and operands accessing for LD1R instructions


src/include/simeng/Pool.hh
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstddef>
+#include <cstdint>
```
FinnWilkinson, line 4: Why was this library needed if no other changes were made?

src/include/simeng/arch/aarch64/helpers/neon.hh
```diff
@@ -547,6 +547,42 @@ class neonHelp {
     return {out, 256};
   }
 
+  /** Helper function for NEON instructions with the format `umaxp rd, vn, vm`.
+   * T represents the type of operands (e.g. for vn.2d, T = uint64_t).
+   * I represents the number of elements in the output array to be updated (e.g.
+   * for vd.8b I = 8).
```
FinnWilkinson, line 553: If the instruction writes to a vector register, the comment should reflect this throughout. Should change end of first line to `umaxp vd, vn, vm`

src/include/simeng/arch/aarch64/helpers/neon.hh
```diff
@@ -547,6 +547,42 @@ class neonHelp {
...
+    for (int i = 0; i < I; i++) {
+      out[i] = std::max(n[i], m[i]);
+    }
+    return {out, 256};
+  }
+
+  /** Helper function for NEON instructions with the format `uminp rd, vn, vm`.
+   * T represents the type of operands (e.g. for vn.2d, T = uint64_t).
+   * I represents the number of elements in the output array to be updated (e.g.
+   * for vd.8b I = 8).
```
FinnWilkinson, line 571: If the instruction writes to a vector register, the comment should reflect this throughout. Should change end of first line to `uminp vd, vn, vm`

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -586,6 +592,32 @@ bool ExceptionHandler::init() {
...
         break;
       }
+      case 278: {  // getrandom
+        // Write <buflen> random bytes to buf
+        uint64_t bufPtr = registerFileSet.get(R0).get<uint64_t>();
+        size_t buflen = registerFileSet.get(R1).get<size_t>();
+
+        char buf[buflen];
+        for (size_t i = 0; i < buflen; i++) {
+          buf[i] = (uint8_t)rand();
```
FinnWilkinson, line 602: To make this truly random, it should probably be seeded... 
From C++ Reference : `srand (time(NULL));` will initialise a random seed.

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -586,6 +592,32 @@ bool ExceptionHandler::init() {
...
+
+        stateChange = {ChangeType::REPLACEMENT, {R0}, {(uint64_t)buflen}};
+
+        stateChange.memoryAddresses.push_back({bufPtr, (uint8_t)buflen});
+        stateChange.memoryAddressValues.push_back(RegisterValue(buf, buflen));
+
+        break;
+      }
+      case 293: {  // shutdown
+        // TODO: Functionality omitted - returning errno 38 (function not
```
FinnWilkinson, line 613: Comment should match the value that is returned.

src/lib/arch/aarch64/InstructionMetadata.cc
```diff
@@ -87,6 +87,9 @@ InstructionMetadata::InstructionMetadata(const cs_insn& insn)
       // adds incorrectly flags destination as READ
       operands[0].access = CS_AC_WRITE;
       break;
+    case Opcode::AArch64_BICv8i16:
+      operands[0].access = CS_AC_WRITE | CS_AC_READ;
+      break;
```
FinnWilkinson, line 92: If access types are not properly set by Capstone decode, you should define them for all operands, not just the destination.

FinnWilkinson, line 92: If access types are not properly set by Capstone decode, you should define them for all operands, not just the destination.

src/lib/arch/aarch64/InstructionMetadata.cc
```diff
@@ -813,6 +816,11 @@ InstructionMetadata::InstructionMetadata(const cs_insn& insn)
       // For vector arrangment of 32-bit, post_index immediate is 4
       operands[2].imm = 4;
       break;
+    case Opcode::AArch64_LD1Onev16b: {
+      // Fix incorrect access types
+      operands[0].access = CS_AC_WRITE;
+      break;
+    }
```
FinnWilkinson, line 823: If access types are not properly set by Capstone decode, you should define them for all operands, not just the destination.

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -1493,12 +1493,16 @@ void Instruction::execute() {
         return executionNYI();
         break;
       }
-      case Opcode::AArch64_CMEQv8i8: {
-        return executionNYI();
+      case Opcode::AArch64_CMEQv8i8: {  // cmeq vd.8b, vn.8b, vm.8b
+        results[0] = neonHelp::vecCompare<uint16_t, 8>(
```
FinnWilkinson, line 1497: The operand type should be `uint8_t`?

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -1493,12 +1493,16 @@ void Instruction::execute() {
...
+            operands, false,
+            [](uint32_t x, uint32_t y) -> bool { return (x == y); });
```
FinnWilkinson, line 1499: The lambda should take in type `uint8_t` rather than `uint32_t` to match the operand type used by the helper funtion.

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -1493,12 +1493,16 @@ void Instruction::execute() {
...
         break;
       }
-      case Opcode::AArch64_CMEQv8i8rz: {
-        return executionNYI();
+      case Opcode::AArch64_CMEQv8i8rz: {  // cmeq vd.8b, vn.8b, #0
+        results[0] = neonHelp::vecCompare<uint16_t, 8>(
```
FinnWilkinson, line 1503: The operand type should be `uint8_t` not `uint16_t`

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -17137,7 +17145,7 @@ void Instruction::execute() {
         break;
       }
       case Opcode::AArch64_UMINPv16i8: {
```
FinnWilkinson, line 17147: Add switch case comment

test/regression/aarch64/Syscall.cc
```diff
@@ -14,6 +14,61 @@ using Syscall = AArch64RegressionTest;
...
+      mov x1, #8
+      ldr x0, buf
+      mov x8, #278
+      svc #0
+
+      # Move buf addr into x1 to easily read it from the test
+      ldr x1, buf
+
+      .balign 1
+      buf:  .skip   4
```
FinnWilkinson, line 46: Instead of using `buf` could you not use the heap you initialised directly, and store the initial heap address in `x1` to extract the contents afterwards?

test/regression/aarch64/Syscall.cc
```diff
@@ -14,6 +14,61 @@ using Syscall = AArch64RegressionTest;
...
+    // init buffer to record the output into
+    buffers[i] = (uint8_t*)calloc(8, sizeof(uint8_t));
+
+    RUN_AARCH64(R"(
+      # Get heap address
+      mov x0, 0
+      mov x8, 214
+      svc #0
+      
+      # getrandom(buf * = [a], buflen = 4, no flags)
```
FinnWilkinson, line 36: Comment says buf length is 4, but looks like it is 8 from the rest of the test

FinnWilkinson, line 36: Comment says buf length is 4, but looks like it is 8 from the rest of the test

test/regression/aarch64/instructions/load.cc
```diff
@@ -180,6 +180,60 @@ TEST_P(InstLoad, ld1_tworeg) {  // 128-bit
...
+    # save heap address after post index
+    mov x11, x0
+
+  )");
+
+  CHECK_NEON(0, uint8_t,
+             {0xFF, 0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88, 0x99,
+              0xAA, 0xBB, 0xCC, 0xDD, 0xEE});
+
+  CHECK_NEON(0, uint8_t,
```
FinnWilkinson, line 229: This test should check register `1` not `0`

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -1663,8 +1667,10 @@ void Instruction::execute() {
         return executionNYI();
         break;
       }
-      case Opcode::AArch64_CMHSv16i8: {
-        return executionNYI();
+      case Opcode::AArch64_CMHSv16i8: {  // cmhs vd.16b, vn.16b, vm.16b
+        results[0] = neonHelp::vecCompare<uint8_t, 16>(
+            operands, false,
+            [](uint8_t x, uint8_t y) -> bool { return (x >= y); });
```
FinnWilkinson, line 1673: I think Comparisons should be signed unless specified otherwise in the ISA profile

test/regression/aarch64/instructions/neon.cc
```diff
@@ -578,6 +622,32 @@ TEST_P(InstNeon, cmeq) {
...
+    ldr q0, [x0]
+    ldr q1, [x0, #16]
+    cmeq v2.8b, v0.8b, v1.8b
+    cmeq v3.8b, v0.8b, 0
+  )");
+
+  CHECK_NEON(2, uint16_t,
+             {0xFFFF, 0xFFFF, 0xFFFF, 0x0000, 0xFFFF, 0xFFFF, 0x0000, 0xFFFF});
+  CHECK_NEON(3, uint16_t,
+             {0xFFFF, 0x0000, 0x0000, 0xFFFF, 0x0000, 0x0000, 0xFFFF, 0x0000});
```
FinnWilkinson, line 649: Should also add some tests for comparing negatives.

test/regression/aarch64/Syscall.cc
tomhepworth: Seeded with clock() instead of time(NULL) as the resolution of time() is too slow (seconds) to give different random numbers on sequential syscalls

configs/m1_firestorm.yaml
```diff
@@ -0,0 +1,352 @@
+---
```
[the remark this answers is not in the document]
tomhepworth: Removed because I think it probably shouldn't be included until it is at least semi-accurate

src/lib/arch/aarch64/ExceptionHandler.cc
tomhepworth: Seeded with clock() because the resolution of time() is too large to give different random bytes on sequential syscalls

src/include/simeng/Pool.hh
tomhepworth: My simeng wont build without it :'( I will remove it from the PR though

tomhepworth:
@jj16791 @FinnWilkinson 

I have noticed a difference in how I have implemented the LD1 instructions compared to some that were already in simeng

In `instruction_address.cc`:
I use `operands[1]` to access the `[xn]` argument:
```
      case Opcode::AArch64_LD1Onev16b: {  // ld1 {vt.16b}, [xn]
        setMemoryAddresses({{operands[1].get<uint64_t>(), 16}});
        break;
      }
```
But older instructions are using operands[0]
```
      case Opcode::AArch64_LD1Twov16b: {  // ld1 {vt1.16b, vt2.16b}, [xn]
        uint64_t base = operands[0].get<uint64_t>();
        setMemoryAddresses({{base, 16}, {base + 16, 16}});
        break;
      }
```

`cstool -sd arm64 0070404C` gives:
```
 0  00 70 40 4c  ld1	{v0.16b}, [x0]
	ID: 143 (ld1)
	op_count: 2
		operands[0].type: REG = v0
		operands[0].access: READ | WRITE
			Vector Arrangement Specifier: 0x2
		operands[1].type: MEM
			operands[1].mem.base: REG = x0
		operands[1].access: READ | WRITE
	Registers read: v0 x0
	Registers modified: v0
	Groups: neon 
```

showing operands[1] as containing the address to load from.

Is this a bug with the older instructions, or is mine wrong? Have I misunderstood something?

tomhepworth:
> @jj16791 @FinnWilkinson
> 
> I have noticed a difference in how I have implemented the LD1 instructions compared to some that were already in simeng
> 
> In `instruction_address.cc`: I use `operands[1]` to access the `[xn]` argument:
> 
> ```
>       case Opcode::AArch64_LD1Onev16b: {  // ld1 {vt.16b}, [xn]
>         setMemoryAddresses({{operands[1].get<uint64_t>(), 16}});
>         break;
>       }
> ```
> 
> But older instructions are using operands[0]
> 
> ```
>       case Opcode::AArch64_LD1Twov16b: {  // ld1 {vt1.16b, vt2.16b}, [xn]
>         uint64_t base = operands[0].get<uint64_t>();
>         setMemoryAddresses({{base, 16}, {base + 16, 16}});
>         break;
>       }
> ```
> 
> `cstool -sd arm64 0070404C` gives:
> 
> ```
>  0  00 70 40 4c  ld1	{v0.16b}, [x0]
> 	ID: 143 (ld1)
> 	op_count: 2
> 		operands[0].type: REG = v0
> 		operands[0].access: READ | WRITE
> 			Vector Arrangement Specifier: 0x2
> 		operands[1].type: MEM
> 			operands[1].mem.base: REG = x0
> 		operands[1].access: READ | WRITE
> 	Registers read: v0 x0
> 	Registers modified: v0
> 	Groups: neon 
> ```
> 
> showing operands[1] as containing the address to load from.
> 
> Is this a bug with the older instructions, or is mine wrong? Have I misunderstood something?

Also curiously, capstone crashes when I try to disassemble ld1 with two vector arguments:

image

tomhepworth:
> > @jj16791 @FinnWilkinson
> > I have noticed a difference in how I have implemented the LD1 instructions compared to some that were already in simeng
> > In `instruction_address.cc`: I use `operands[1]` to access the `[xn]` argument:
> > ```
> >       case Opcode::AArch64_LD1Onev16b: {  // ld1 {vt.16b}, [xn]
> >         setMemoryAddresses({{operands[1].get<uint64_t>(), 16}});
> >         break;
> >       }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But older instructions are using operands[0]
> > ```
> >       case Opcode::AArch64_LD1Twov16b: {  // ld1 {vt1.16b, vt2.16b}, [xn]
> >         uint64_t base = operands[0].get<uint64_t>();
> >         setMemoryAddresses({{base, 16}, {base + 16, 16}});
> >         break;
> >       }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `cstool -sd arm64 0070404C` gives:
> > ```
> >  0  00 70 40 4c  ld1	{v0.16b}, [x0]
> > 	ID: 143 (ld1)
> > 	op_count: 2
> > 		operands[0].type: REG = v0
> > 		operands[0].access: READ | WRITE
> > 			Vector Arrangement Specifier: 0x2
> > 		operands[1].type: MEM
> > 			operands[1].mem.base: REG = x0
> > 		operands[1].access: READ | WRITE
> > 	Registers read: v0 x0
> > 	Registers modified: v0
> > 	Groups: neon 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > showing operands[1] as containing the address to load from.
> > Is this a bug with the older instructions, or is mine wrong? Have I misunderstood something?
> 
> So the CStool output shows all operands, hence the memory operand will be `[1]`. But in `Instruction_address.cc` the `operands` array is just the source operands rather than all operands; so here `[xn]` will be `operands[0]` as `{v0.16b}` is a destination operand rather than a source operand.
> 
> TLDR : the older implementation is correct, in this case you should be accessing `operands[0]`.

Still a bit confused because my tests pass with operands[1], and I have noticed other old implementations using operands[1] eg:

```
      case Opcode::AArch64_LD1Rv4s: {  // ld1r {vt.4s}, [xn]
        setMemoryAddresses({{operands[1].get<uint64_t>(), 16}});
        break;
      }
      case Opcode::AArch64_LD1Rv4s_POST: {  // ld1r {vt.4s}, [xn], #imm
        setMemoryAddresses({{operands[1].get<uint64_t>(), 16}});
        break;
      }
```

Does this mean my metadata is wrong? There is currently no entry in instruction_metadata for my instructions, as I have not noticed incorrect behaviour.

If I add the following to instruction_metadata.cc...
```
   case Opcode::AArch64_LD1Onev16b:
      operands[0].access = CS_AC_WRITE;
      operands[1].access = CS_AC_READ;
      break;
    case Opcode::AArch64_LD1Onev16b_POST:
      operands[0].access = CS_AC_WRITE;
      operands[1].access = CS_AC_READ | CS_AC_WRITE;
```
...and adjust my code to use operands[0] for the [xn] argument, my tests pass.

Which is preferable here?

FinnWilkinson:
> > > @jj16791 @FinnWilkinson
> > > I have noticed a difference in how I have implemented the LD1 instructions compared to some that were already in simeng
> > > In `instruction_address.cc`: I use `operands[1]` to access the `[xn]` argument:
> > > ```
> > >       case Opcode::AArch64_LD1Onev16b: {  // ld1 {vt.16b}, [xn]
> > >         setMemoryAddresses({{operands[1].get<uint64_t>(), 16}});
> > >         break;
> > >       }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > But older instructions are using operands[0]
> > > ```
> > >       case Opcode::AArch64_LD1Twov16b: {  // ld1 {vt1.16b, vt2.16b}, [xn]
> > >         uint64_t base = operands[0].get<uint64_t>();
> > >         setMemoryAddresses({{base, 16}, {base + 16, 16}});
> > >         break;
> > >       }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > `cstool -sd arm64 0070404C` gives:
> > > ```
> > >  0  00 70 40 4c  ld1	{v0.16b}, [x0]
> > > 	ID: 143 (ld1)
> > > 	op_count: 2
> > > 		operands[0].type: REG = v0
> > > 		operands[0].access: READ | WRITE
> > > 			Vector Arrangement Specifier: 0x2
> > > 		operands[1].type: MEM
> > > 			operands[1].mem.base: REG = x0
> > > 		operands[1].access: READ | WRITE
> > > 	Registers read: v0 x0
> > > 	Registers modified: v0
> > > 	Groups: neon 
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > showing operands[1] as containing the address to load from.
> > > Is this a bug with the older instructions, or is mine wrong? Have I misunderstood something?
> > 
> > 
> > So the CStool output shows all operands, hence the memory operand will be `[1]`. But in `Instruction_address.cc` the `operands` array is just the source operands rather than all operands; so here `[xn]` will be `operands[0]` as `{v0.16b}` is a destination operand rather than a source operand.
> > TLDR : the older implementation is correct, in this case you should be accessing `operands[0]`.
> 
> Still a bit confused because my tests pass with operands[1], and I have noticed other old implementations using operands[1] eg:
> 
> ```
>       case Opcode::AArch64_LD1Rv4s: {  // ld1r {vt.4s}, [xn]
>         setMemoryAddresses({{operands[1].get<uint64_t>(), 16}});
>         break;
>       }
>       case Opcode::AArch64_LD1Rv4s_POST: {  // ld1r {vt.4s}, [xn], #imm
>         setMemoryAddresses({{operands[1].get<uint64_t>(), 16}});
>         break;
>       }
> ```
> 
> Does this mean my metadata is wrong? There is currently no entry in instruction_metadata for my instructions, as I have not noticed incorrect behaviour.
> 
> If I add the following to instruction_metadata.cc...
> 
> ```
>    case Opcode::AArch64_LD1Onev16b:
>       operands[0].access = CS_AC_WRITE;
>       operands[1].access = CS_AC_READ;
>       break;
>     case Opcode::AArch64_LD1Onev16b_POST:
>       operands[0].access = CS_AC_WRITE;
>       operands[1].access = CS_AC_READ | CS_AC_WRITE;
> ```
> 
> ...and adjust my code to use operands[0] for the [xn] argument, my tests pass.
> 
> Which is preferable here?

Looking in the Capstone instruction access map the LD1Rv4s and LD1Rv4s_POST instructions have incorrect destination register access rights (both are Read & Write instead of just Write).

It would be preferable to add cases to instruction_metadata for both your new instructions and the old ones you've found to fix the incorrect access rights.

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -586,6 +592,37 @@ bool ExceptionHandler::init() {
         stateChange = {ChangeType::REPLACEMENT, {R0}, {0ull}};
         break;
       }
+      case 278: {  // getrandom
+        // TODO: support flags argument
```
FinnWilkinson, line 596: Ideally the flags should be supported before this PR is merged

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -7881,11 +7889,11 @@ void Instruction::execute() {
         break;
       }
       case Opcode::AArch64_LD1Rv2d_POST: {  // ld1r {vt.2d}, [xn], #imm
-        // LOAD
+        // LO0D
```
FinnWilkinson, line 7892: Typo

tomhepworth: How did I manage this... must have been a Vim moment

src/lib/arch/aarch64/ExceptionHandler.cc
tomhepworth: From the man pages, the flags are:
```
GRND_RANDOM
              If this bit is set, then random bytes are drawn from the
              random source (i.e., the same source as the /dev/random
              device) instead of the urandom source.  The random source
              is limited based on the entropy that can be obtained from
              environmental noise.  If the number of available bytes in
              the random source is less than requested in buflen, the
              call returns just the available random bytes.  If no
              random bytes are available, the behavior depends on the
              presence of GRND_NONBLOCK in the flags argument.

       GRND_NONBLOCK
              By default, when reading from the random source,
              getrandom() blocks if no random bytes are available, and
              when reading from the urandom source, it blocks if the
              entropy pool has not yet been initialized.  If the
              GRND_NONBLOCK flag is set, then getrandom() does not block
              in these cases, but instead immediately returns -1 with
              [errno](https://man7.org/linux/man-pages/man3/errno.3.html) set to EAGAIN.
```
Do we want to simulate this blocking or the EAGAIN signal? If not then perhaps it is safe to ignore the flags after all?

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -994,7 +994,7 @@ void Instruction::execute() {
         break;
       }
       case Opcode::AArch64_BRK: {
```
BraveSirRobbo, line 996: add descriptor comment ?

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -994,7 +994,7 @@ void Instruction::execute() {
...
-        return executionNYI();
+        // TODO: Generate breakpoint exception
```
jj16791, line 997: I fear that having this as a TODO comment rather than an NYI exception could cause issues in the future.

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -6527,7 +6536,8 @@ void Instruction::execute() {
         break;
       }
       case Opcode::AArch64_FSUBv2f32: {
```
BraveSirRobbo, line 6538: add descriptor comment

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -32,6 +32,11 @@ std::enable_if_t<std::is_integral_v<T> && std::is_unsigned_v<T>, T> shiftValue(
       amount &= mask;
       return (value >> amount) | (value << ((-amount) & mask));
     }
+    case ARM64_SFT_MSL: {
```
jj16791, line 35: Does this have a use in one of the new test cases?

test/regression/aarch64/instructions/neon.cc
```diff
@@ -368,6 +368,50 @@ TEST_P(InstNeon, and) {
 }
```
jj16791, line 368: I working under the assumption that each new instruction has an associated set of tests

jj16791 (changes requested):
A few comments. Hard to tell from face value if all new instructions have an associated test, they will have to for the PR to go through.

FinnWilkinson:

BraveSirRobbo (dismissed):
Please run clang-format on all changed files as there seem to be some missing formatting

I've fixed some small formatting issues in two commits b1d1201 and 0270c09

And I've included a few other comments

src/lib/arch/aarch64/Instruction_address.cc
```diff
@@ -415,6 +423,9 @@ span<const MemoryAccessTarget> Instruction::generateAddresses() {
         setMemoryAddresses(std::move(addresses));
         break;
       }
+      case Opcode::AArch64_LD2Twov4s: {  // ld2 {vt1.4s, vt2.4s}, [xn]
```
jj16791, line 426: Remove curl brackets if using fallthrough?

src/lib/arch/aarch64/Instruction_address.cc
```diff
@@ -415,6 +423,9 @@ span<const MemoryAccessTarget> Instruction::generateAddresses() {
...
+        [[fallthrough]];
+      }
```
jj16791, line 428: Remove curl brackets if using fallthrough?

jj16791:
#rerun tests

src/include/simeng/arch/aarch64/Instruction.hh
tomhepworth: Yeah there is an MSL option in eg `MVNI <Vd>.<T>, #<imm8>, MSL #<amount>` which I have tested.

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -4,6 +4,7 @@
 
 #include <iomanip>
 #include <iostream>
+#include <ostream>
```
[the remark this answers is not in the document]
tomhepworth: Removed

src/lib/arch/aarch64/Instruction_execute.cc
tomhepworth: I have added a new exception type and thrown it here, but it remains unhandled. I would be unsure about the best way to handle the exception so perhaps we can chat about it?

test/regression/aarch64/instructions/neon.cc
tomhepworth: Yes they should do, apart from the BRK instruction for reasons I mentioned in the other comment

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -538,6 +545,14 @@ bool ExceptionHandler::init() {
             ChangeType::REPLACEMENT, {R0}, {static_cast<uint64_t>(result)}};
         break;
       }
+      case 210: {  // shutdown
+        // TODO: Functionality omitted - returns -38 (errno 38, function not
+        // implemented) is to mimic the behaviour on isambard and avoid an
```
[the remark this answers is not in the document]
tomhepworth: I have replaced the comment but please check if it is satisfactory.

tomhepworth:
Merge went wrong so force pushed to revert it.

src/lib/arch/aarch64/InstructionMetadata.cc
```diff
@@ -2000,6 +2083,18 @@ void InstructionMetadata::revertAliasing() {
...
+          opcode == Opcode::AArch64_REV64v2i32 ||
+          opcode == Opcode::AArch64_REV64v4i16 ||
+          opcode == Opcode::AArch64_REV64v4i32 ||
+          opcode == Opcode::AArch64_REV64v8i16 ||
+          opcode == Opcode::AArch64_REV64v8i8) {
+        operandCount = 2;
+        operands[0].access = CS_AC_WRITE;
+        operands[1].access = CS_AC_READ;
+        return;
+      }
```
FinnWilkinson, line 2097: I am confused by this alias addition. If the access rights are not correct then add it to the main switch statement instead of the revertalias function.

If they do infact alias to something weird and this is required, then please make this clear in the comment

FinnWilkinson (changes requested):
Some comments to clarify or changes to make, but mostly good. You also need to sort out our clang-format as it keeps failing the Jenkins pipeline, and you need to resolve the merge conflic also (rebasing to dev should fix this, and is good to do before merging a PR anyway)

src/include/simeng/arch/aarch64/helpers/neon.hh
```diff
@@ -738,6 +774,150 @@ class neonHelp {
...
+   * Returns correctly formatted RegisterValue. */
+  template <typename T, int V, int I>
+  static RegisterValue vecRev(
+      std::array<RegisterValue, Instruction::MAX_SOURCE_REGISTERS>& operands) {
+    const T* source = operands[0].getAsVector<T>();
+    int element_size = (sizeof(T) * 8);
+    int datasize = I * element_size;
+    int container_size = V;
+    int n_containers = datasize / container_size;
+    int elements_per_container = container_size / element_size;
```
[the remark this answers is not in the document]
tomhepworth: If valid inputs are given then they should always be whole. Should I add some assertions to ensure inputs are valid?

src/include/simeng/arch/aarch64/helpers/neon.hh
```diff
@@ -738,6 +774,150 @@ class neonHelp {
...
+
+    int element = 0;
+    int rev_element;
+    T out[16 / sizeof(T)] = {0};
+    for (int c = 0; c < n_containers; c++) {
+      rev_element = element + elements_per_container - 1;
```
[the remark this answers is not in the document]
tomhepworth: As this line is selecting an index I felt it was more readable this way. Also it would break the for loop which follows.

src/include/simeng/arch/aarch64/helpers/neon.hh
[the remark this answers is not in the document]
FinnWilkinson: I don't think it is necessary, as there wouldn't be a valid decoding with non-valid parameters

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -586,6 +599,34 @@ bool ExceptionHandler::init() {
...
+
+        char buf[buflen];
+        for (size_t i = 0; i < buflen; i++) {
+          buf[i] = (uint8_t)rand();
+        }
+
+        stateChange = {ChangeType::REPLACEMENT, {R0}, {(uint64_t)buflen}};
+
+        stateChange.memoryAddresses.push_back({bufPtr, (uint8_t)buflen});
+        stateChange.memoryAddressValues.push_back(RegisterValue(buf, buflen));
```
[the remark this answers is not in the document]
jj16791: It shouldn't matter. We've used the 128-bit restriction prior to mimic the multi-cycle latency of the syscalls called many times

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -538,6 +543,14 @@ bool ExceptionHandler::init() {
             ChangeType::REPLACEMENT, {R0}, {static_cast<uint64_t>(result)}};
         break;
       }
+      case 210: {  // shutdown
+        // TODO: Functionality omitted - returns -38 (errno 38, function not
+        // implemented) The targetted linux kernel does not implement this
```
jj16791, line 548: Just reword to:

TODO: Functionality omitted - returns -38 (errno 38, function not implemented). The targetted linux kernel does not implement this syscall.

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -993,8 +993,11 @@ void Instruction::execute() {
         return executionNYI();
         break;
       }
-      case Opcode::AArch64_BRK: {
-        return executionNYI();
+      case Opcode::AArch64_BRK: {  // brk #imm16
```
jj16791, line 996: We need a case in the exception handler to support this, namely just a "unsupported breakpoint exception" printout

jj16791 (changes requested):
Only a couple of new comments

tomhepworth:
#rerun tests

src/lib/arch/aarch64/ExceptionHandler.cc
jj16791, line 548: Fullstop after ")" before capitalised "The"

jj16791 (changes requested):
Typo

tomhepworth:
#rerun tests

jj16791 (approved):

FinnWilkinson (approved):

Issue #238, opened 2022-07-30
Add Predicate pattern recognition to other SVE instructions


FinnWilkinson:
Currently only PTRUE has named predicant pattern recognition, but other instructions can also include said optional pattern. This needs adding into the execution behaviour of the relevant instructions / helper functions to prevent any minor bugs from occuring.

The `sveGetPattern` function in `include/simeng/arch/aarch64/helpers/auxiliaryFunctions.hh` which is used to extract the correct pattern will require modification as it will currently only work correctly for PTRUE.

Issue #237, opened 2022-07-01
Move Virtual Counter Timer and Processor Cycle Counter from main.cc


FinnWilkinson:
These system registers (CNTVCT, CNTPCT) are AArch64 specific and so not all ISAs will need to update these registers (RISC-V for example will not update them).

They are currently updated via function calls in `main.cc`. They should instead be updated by the core each tick by calling some counter-update function in the architecture - therefore allowing non-AArch64 ISAs to alter or remove this functionality.

PR #240, opened 2022-08-03
Minor Feature Fixes

FinnWilkinson:
This PR addresses issues #238 and #237; Making Processor and Virtual counter timer logic architecture agnostic and ensuring all relevant SVE instructions have predicate pattern recognition.

Also, the specialFiles generation directory has been changed from the project source to the project build directory.


test/regression/aarch64/AArch64RegressionTest.hh
```diff
@@ -6,6 +6,7 @@
 
 #define AARCH64_CONFIG                                                         \
   ("{Core: {Simulation-Mode: emulation, Clock-Frequency: 2.5, "                \
+   "Timer-Frequency: 100, "                                                    \
```
jj16791, line 9: Could we have this formatted correctly (i.e. have ends of multi-line string more-or-less aligned)

src/include/simeng/arch/aarch64/helpers/auxiliaryFunctions.hh
```diff
@@ -265,16 +265,22 @@ class AuxFunc {
   static uint16_t sveGetPattern(const std::string operandStr,
                                 const uint8_t esize, const uint16_t VL_) {
     const uint16_t elements = VL_ / esize;
-    // If not pattern then same as ALL
-    if (operandStr.find(",") == std::string::npos) return elements;
+    const std::vector<std::string> patterns = {
```
jj16791, line 268: What happens when we get a pattern we don't support?

jj16791 (changes requested):
A couple of changes/clarifications needed

test/regression/aarch64/AArch64RegressionTest.hh
FinnWilkinson: I tried my best to do this, but clang format would not recognise it as a `define` and produced incorrect code. I will try again to format it correctly

src/include/simeng/arch/aarch64/helpers/auxiliaryFunctions.hh
FinnWilkinson: All possible predicate patterns for armv8.4 and armv9.2 are included in this vector. However, if a pattern that wasn't supported was used then this function will treat it as an `all` (i.e. same as no pattern).

jj16791: Is it worth having a debug assert clause to avoid issues with unsupported patterns causing functional errors?

jj16791: So if we found a pattern in the operandStr that doesn't have an associated clause in the if-else structure below, then raise trigger an `assert` clause? It would be better if this were baked into the runtime (i.e. raise an NYI exec with some sort of extra clause to say it's the pattern that's not simply rather than the opcode). Additionally, how much work is it to have an else if clause for each pattern?

FinnWilkinson: I’m not sure I fully understand. For each pattern in the vector there is an associated if clause. Are you saying we should have a default clause to catch any other patterns? Or are you saying we should catch a pattern that isn’t in the patterns vector? As if the latter is the case I’m not sure it is possible

jj16791: Ah, some of the lines were omitted from the preview. Didn't see that all patterns have been supported. If all the patterns within the string vector contain all patterns supported under SVE then we should be good regarding functional issue due to lacking pattern support right?

FinnWilkinson: Yeah, the only case that a pattern is not recognised is if more are introduced in a new ISA release

jj16791: Then maybe a default case which raises a Debug assert would be worth it in those cases (i.e. just above the return 0 line)

jj16791: something like `assert(false && "Unknown predicate pattern")`

FinnWilkinson: Can do yes. Although I don’t think anything would trigger it. I should have clarified that the pattern string is initialised to “all” to accommodate for the fact that the pattern is optional, and so if no pattern in the vector is seen it assumes this is the case

jj16791: Ah I see this issue now. We may have to just keep a close eye on this then and update the vector as needed.

jj16791: I'll mark this as resolved but we need to remember this as a possible future issue.

jj16791 (approved):
All looks good to me.

BraveSirRobbo (commented):
one query on the 50x nops, but also - failures on the Jenkins pipeline, is that a problem ?

test/regression/aarch64/SystemRegisters.cc
```diff
@@ -71,6 +71,64 @@ TEST_P(SystemRegister, sysreg_access) {
   EXPECT_EQ(getSystemRegister(0xde82), 42);
 }
 
+TEST_P(SystemRegister, counter_timers) {
+  // Ensure that the VCT is incremented at correct rate : once per ((2.5 * 1e9)
+  // / (100 * 1e6)) cycles (i.e. once per 25 cycles).
+  RUN_AARCH64(R"(
```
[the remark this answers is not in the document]
FinnWilkinson: The 50 `nop` instructions ensures 50 cycles for emulation (and a couple more for OoO due to  pushing the instructions through the whole pipeline) which ensures that the virtual counter timer is incremented just twice.

A loop could be used instead however... It was just easier to implement a bunch of `nop`s. I will look into changing it into a loop

tomhepworth (approved):

src/include/simeng/arch/aarch64/helpers/auxiliaryFunctions.hh
```diff
@@ -265,16 +265,22 @@ class AuxFunc {
...
+        "pow2", "vl1",  "vl2",  "vl3",   "vl4",   "vl5",  "vl6",  "vl7", "vl8",
+        "vl16", "vl32", "vl64", "vl128", "vl256", "mul3", "mul4", "all"};
 
-    // Get pattern string
-    std::string pattern(operandStr.substr(operandStr.find(",") + 2));
-    if (pattern.find(',') != std::string::npos) {
-      pattern.erase(pattern.find(","));
+    // If not pattern then same as ALL
```
FinnWilkinson, line 272: Comment should be changed to be more easily read - something like "If no pattern present in operandStr then same behaviour as ALL"

FinnWilkinson (commented):
Minor changes needed, along with existing issues pointed out by other reviewers

rahahahat (approved):

PR #242, opened 2022-08-09
Added comments explaining ELF File structure to Elf.cc file.

rahahahat:
Reference: https://man7.org/linux/man-pages/man5/elf.5.html


FinnWilkinson:

src/lib/Elf.cc
```diff
@@ -61,12 +141,27 @@ Elf::Elf(std::string path) {
     file.read(reinterpret_cast<char*>(&(header.fileSize)), fieldBytes);
     file.read(reinterpret_cast<char*>(&(header.memorySize)), fieldBytes);
 
+    // To construct the process we look for the largest virtual address and
+    // add it to the memory size of the header. This way we obtain a very
+    // large array which can hold data at large virtual address. 
+    // However, this way we end up creating a sparse array, in which most 
+    // of the entries are unused. Also SimEng internally treats these
+    // virtual address as physical addresses to index into this large array. 
+    // TODO: This mechanism needs to be improved.
```
jj16791, line 150: What would the alternative be? If we don't respect the spatial difference between these ELF sections, we may encounter issues relating to relative offsets (e.g. something stored in the .data section). We could have a translator within the memory interfaces but, from the experience of trying, it has a significant enough performance overhead

jj16791 (changes requested):
Added a comment regarding the `// TODO: This mechanism needs to be improved.`

src/lib/Elf.cc
rahahahat: I left this comment after some consultation with Finn. Afaik we discussed an implementation of a TLB. 
Sorry I didn't know that it had already been tried. In that case, I will remove this comment. But out of curiosity what sort translator did you implement? and why would we encounter issues with relative offsetting if we translate each address even though we copy all LOAD sections?
This might be silly but would it be possible to replace vaddr with translated addrs during the decode phase, to save additional translations during execution/branching/load/stores?

jj16791: It stacked all the segments on top of each other essentially. Then when access it at L1D/L1I cache, it would translate into the correct region and then use a relative offset from that region base to get the correct data

FinnWilkinson (approved):
Approve this PR based on the resolution of outstanding comment

jj16791 (approved):

PR #243, opened 2022-08-11
Process Image memory optimisation for SimEng [WIP]

rahahahat:
In `Elf.hh`, SimEng parses ELF binaries and extracts the initialized data as `processImage_`. This `processImage_` is returned as `span<char>` and used in `LinuxProcess.hh` to create the actual process image. In `LinuxProcess.hh` the size of the image is first byte-aligned and then incremented by configuration variables: `HEAP_SIZE` and `STACK_SIZE`, because of this a `char *` of the new size is initialized and the `processImage` obtained from `Elf.hh` is copied for the first time.

Then in `main.cc` the newly size extended `processImage` is returned as a `span<char>` from `LinuxProcess.hh`. Here it is again copied into another `char*` array so that it can be passed to other classes. the previous copy of the `processImage` present in `LinuxProcess.hh` remains persistent until the destructor for `LinuxProcess.hh` is called, which only happens when `main.cc` terminates.

Furthermore, for SST builds this processImage will need to copied for the third time to SST's memory model, leaving two persistent copies behind.

This PR introduces a possible solution to this problem by holding an original reference `char * processImage_` inside `LinuxProcess.hh/cc` and construction/resizing the same pointer through `malloc` and `realloc` calls. This original reference is then passed to every other object which requires it.


FinnWilkinson:
Please fix Jenkins assertion error

src/include/simeng/Elf.hh
```diff
@@ -25,6 +25,7 @@ struct ElfHeader {
 class Elf {
  public:
   Elf(std::string path);
```
jj16791, line 27: Why are we keeping this constructor?

src/lib/kernel/LinuxProcess.cc
```diff
@@ -1,6 +1,7 @@
 #include "simeng/kernel/LinuxProcess.hh"
 
 #include <cassert>
+#include <iostream>
```
jj16791, line 4: What needs this library?

src/lib/Elf.cc
```diff
@@ -81,12 +81,86 @@ Elf::Elf(std::string path) {
   file.close();
 }
 
-Elf::~Elf() {
-  if (isValid_) {
-    delete[] processImage_;
+Elf::Elf(std::string path, std::vector<char>& imageVec) {
```
jj16791, line 84: I assume apart from the use of `imageVec`, the functionality/code is identical to the original ELF constructor?

src/tools/simeng/main.cc
```diff
@@ -202,10 +202,10 @@ int main(int argc, char** argv) {
   }
 
   // Read the process image and copy to memory
```
jj16791, line 204: Reword comment to better suit the new functionality i.e. we're not copying anymore but rather getting the `processMemory` pointer to point to the image vector

jj16791 (changes requested):
A few comments and it fails on all CI test pipelines.

FinnWilkinson:

src/include/simeng/Elf.hh
rahahahat: I have left most of the old code, comments, etc because I wanted you and Finn you approve the new methodology first. Will be tidying it up soon

src/lib/Elf.cc
rahahahat: Yes.

jj16791:
@mutalibmohammed Do you have any tools already in use which would help us understand the improved memory overhead?

src/include/simeng/kernel/LinuxProcess.hh
```diff
@@ -60,6 +62,9 @@ class LinuxProcess {
   /** Get the process image. */
   const span<char> getProcessImage() const;
 
+  /** Get the process image vector */
```
FinnWilkinson, line 65: Minor preference to end header file function comments with a full stop

src/include/simeng/kernel/LinuxProcess.hh
```diff
@@ -100,6 +105,9 @@ class LinuxProcess {
   /** The process image. */
   char* processImage_;
 
+  /** The process image as a vector */
```
FinnWilkinson, line 108: Minor preference to end header file function comments with a full stop

src/lib/Elf.cc
```diff
@@ -66,7 +66,8 @@ Elf::Elf(std::string path) {
     }
   }
 
-  processImage_ = new char[processImageSize_];
+  *imagePointer = (char*)malloc(processImageSize_ * sizeof(char));
+  processImage_ = *imagePointer;
```
jj16791, line 70: Shouldn't we be getting rid of this copy as well?

src/include/simeng/kernel/LinuxProcess.hh
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <vector>
```
jj16791, line 3: Don't think this is needed

src/lib/kernel/LinuxProcess.cc
```diff
@@ -2,6 +2,7 @@
 
 #include <cassert>
 #include <cstring>
+#include <iostream>
```
jj16791, line 5: Do we need this to be included? Possibly left over form debugging

src/lib/kernel/LinuxProcess.cc
```diff
@@ -22,18 +23,17 @@ LinuxProcess::LinuxProcess(const std::vector<std::string>& commandLine,
...
   }
   isValid_ = true;
 
   entryPoint_ = elf.getEntryPoint();
 
-  span<char> elfProcessImage = elf.getProcessImage();
-
   // Align heap start to a 32-byte boundary
-  heapStart_ = alignToBoundary(elfProcessImage.size(), 32);
+  heapStart_ = alignToBoundary(elf.getProcessImage().size(), 32);
```
jj16791, line 36: Could change this to a processImageSize_ getter instead

src/lib/kernel/LinuxProcess.cc
```diff
@@ -76,7 +83,7 @@ LinuxProcess::LinuxProcess(span<char> instructions, YAML::Node config)
 
 LinuxProcess::~LinuxProcess() {
   if (isValid_) {
-    delete[] processImage_;
+    free(processImage_);
```
jj16791, line 86: Have we considered a smart pointer approach with `std::shared_ptr<char*> processImage_`

jj16791: I'd argue it's a desirable approach for a C++17 project which already makes heavy use of them

src/tools/simeng/main.cc
jj16791: Comment persist after file change

jj16791 (changes requested):
Comment persist and some additional comments/possible changes

.jenkins/run.sh
```diff
@@ -4,4 +4,4 @@ pwd
 
 source .jenkins/build_test_run.sh
 
-run
+run
```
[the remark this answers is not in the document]
rahahahat: This change doesn't exist, However GitHub still shows it. I fixed it quite a while ago. If you take a look at the file in the branch it is identical to one in master

src/lib/kernel/LinuxProcess.cc
rahahahat: PTAL

src/lib/kernel/LinuxProcess.cc
rahahahat: PTAL

src/include/simeng/kernel/LinuxProcess.hh
rahahahat: Done

src/lib/Elf.cc
rahahahat: Removed. PTAL.

src/include/simeng/kernel/LinuxProcess.hh
rahahahat: Done.

src/lib/kernel/LinuxProcess.cc
rahahahat: It is now needed for `std:cerr`

test/regression/RegressionTest.cc
```diff
@@ -53,23 +50,29 @@ void RegressionTest::run(const char* source, const char* triple) {
...
-  // Allocate memory for the process and copy the full process image to it
-  simeng::span<char> processImage = process_->getProcessImage();
-  processMemorySize_ = processImage.size();
-  if (processMemory_) delete[] processMemory_;
-  processMemory_ = new char[processMemorySize_];
-  std::copy(processImage.begin(), processImage.end(), processMemory_);
-
-  // Create memory interfaces for instruction and data access
-  simeng::FlatMemoryInterface instructionMemory(processMemory_,
+  processMemorySize_ = process_->getProcessImageSize();
```
FinnWilkinson, line 53: May be useful to keep old comment above this line `// Allocate memory for the process and copy the full process image to it`

test/regression/RegressionTest.cc
```diff
@@ -53,23 +50,29 @@ void RegressionTest::run(const char* source, const char* triple) {
...
+  simeng::FlatMemoryInterface instructionMemory(procImgForInstrMem.get(),
                                                 processMemorySize_);
+
+  std::shared_ptr<char> procImgForFlatDataMem = process_->getProcessImage();
   std::unique_ptr<simeng::FlatMemoryInterface> flatDataMemory =
-      std::make_unique<simeng::FlatMemoryInterface>(processMemory_,
+      std::make_unique<simeng::FlatMemoryInterface>(procImgForFlatDataMem.get(),
                                                     processMemorySize_);
+
+  std::shared_ptr<char> procImgForFixedDataMem = process_->getProcessImage();
```
FinnWilkinson, line 72: Rather than having 3 shared pointers which all point to the same process image, why can't we use a single `processImage` shared pointer and pass this to each MemoryInterface constructor?

FinnWilkinson (changes requested):
A few minor points for discussion in RegressionTest.cc

test/regression/RegressionTest.cc
rahahahat: Thanks for pointing this out, I was confused with shared_ptr ownership so I copied it 3 times, which shouldn't be the case. Since the scope which contains these objects retains the ownership. I have fixed it now.

test/regression/RegressionTest.cc
rahahahat: But because of the optimisation, no extra memory allocation takes place. Only once while creating the LinuxProcess object?

rahahahat: PTAL.

test/regression/RegressionTest.cc
rahahahat: PTAL

FinnWilkinson (approved):
LGTM

jj16791 (approved):

dANW34V3R (approved):
LGTM

PR #244, opened 2022-08-30
GDB support - SE-123

BraveSirRobbo:
This adds GDB Support !
With this, GDB can now connect to SimEng via its remote debugging feature.

Please note that support for runtime flags for verbose output and for changing the port is currently unfinished.  Support for this requires support for GDB's "run" and "start".  I attempted this, but wasn't able to do it within the time of the internship.

I have rebased this to dev, after originally branching off main

Instructions for use (also included in docs/sphinx/developer/test/index.rst):

1. Ensure that you have gdb-multiarch installed by typing ``gdb-multiarch -v``, otherwise install it using your favourite package manager. (N.B. if you are running this on a machine that matches the architecture that SimEng is running, then a default gdb install should suffice - just replace instances of ``gdb-multiarch`` in these instructions with ``gdb``.)

2. When building SimEng, include the parameter ``-DSIMENG_ENABLE_GDB=ON``, and otherwise continue to build and install SimEng as you would normally.

3. Run SimEng as you would normally, including a config file and the binary you would like to debug.  The console's output should say ``Started listening on port 2424``.

4. In another console window, make sure you are in the same folder as the binary you are debugging, and run ``gdb-multiarch <binary name>``.  This should give you a ``(gdb)`` prompt after having read in the symbols from your binary.

5. In the GDB shell, type ``set architecture aarch64``. (N.B. this step is not needed if you are aready running on an aarch64 processor).

6. In the GDB shell, type ``target remote localhost:2424``.  After a small delay, the SimEng console should print ``Connection to GDB client established, debugging in progress`` and the gdb shell should print the memory address of ``_start``, and provide you with a ``(gdb)`` prompt.
 
7. Continue to use GDB as you would normally, noting what SimEng currently supports:

* Setting breakpoints (but not watchpoints)
* Continuing until a breakpoint or the end of the program
* Stepping through, including one instruction at a time
* Reading registers (currently only general purpose registers) and memory
* The TUI, showing registers, assembly or source code (if applicable)
* N.B. ``run``, ``start``, arguments are not currently supported

8. Once SimEng has finished running your binary, use ctrl+c to kill the SimEng instance.  At this point, GDB should close its connection to SimEng.  Simply re-run SimEng and continue in GDB from the ``target remote localhost:2424`` command.


src/tools/simeng/main.cc
```diff
@@ -87,6 +89,11 @@ int main(int argc, char** argv) {
     mode = SimulationMode::OutOfOrder;
   }
 
+#ifdef GDB_ENABLED
+  std::cout << "GDB Enabled\n";
+  mode = SimulationMode::Emulation;
+#endif
```
FinnWilkinson, line 95: This `#ifdef GDB_ENABLED`  clause could encapsulate the original SimulationMode setting logic as currently it could be doing this logic redundantly. i.e. add an else clause to your `#ifdef` which surrounds the setting of `SimulationMode` from the config file.

src/tools/simeng/main.cc
```diff
@@ -30,6 +31,7 @@ enum class SimulationMode { Emulation, InOrderPipelined, OutOfOrder };
 
 float clockFreq_;
 uint32_t timerFreq_;
+bool gdb = 0;
```
FinnWilkinson, line 34: This variable seems redundant

src/include/simeng/GDBStub.hh
```diff
@@ -0,0 +1,98 @@
...
+  /** Convert uint64_t into RSP compliant hex - a series of 8 hex bytes. */
+  std::string decToRSP(uint64_t dec) const;
+
+  /** Compute an RSP compliant checksum from a packet (sum of chars % 256, in a
+   * 2 digit hex). */
+  std::string computeChecksum(std::string packet) const;
+
+  /** Add RSP compliant start and end characters, followed by a checksum. */
+  std::string generateReply(std::string packet) const;
+  ;
```
FinnWilkinson, line 56: Extra `;`

src/include/simeng/GDBStub.hh
```diff
@@ -0,0 +1,98 @@
...
+   * breakpoints. */
+  std::string handleContinue();
+
+  /** Create a socket and listen on the port number provided.
+   * Socket handling code taken from:
+   * https://ncona.com/2019/04/building-a-simple-server-with-cpp/. */
+  int openSocket(int port) const;
+
+  /** Boolean for if the runtime -v verbose flag has been set. */
+  bool verbose_ = 0;
```
FinnWilkinson, line 85: Probably preferable to initialise to `false` rather than 0?

src/include/simeng/GDBStub.hh
```diff
@@ -0,0 +1,98 @@
...
+
+  /** The core used for the simulation.  Currently only supports the
+   * emulation core*/
```
FinnWilkinson, line 88: Add space and `.` at end of second line

src/tools/simeng/gdb/GDBStub.cc
```diff
@@ -0,0 +1,331 @@
+#include "simeng/GDBStub.hh"
+
+namespace simeng {
+
+GDBStub::GDBStub(simeng::Core& core, simeng::MemoryInterface& dataMemory)
+    : core_(core), dataMemory_(dataMemory) {}
+
+int GDBStub::run() {
+  // verbose_ = 1; // uncomment this if you'd like the GDBStub to run verbosely
```
FinnWilkinson, line 9: Is there a way to handle verbose runs using a runtime or cmake argument rather than having to manually un-comment code?

src/tools/simeng/gdb/GDBStub.cc
```diff
@@ -0,0 +1,331 @@
...
+        std::cout << RESET;
+      }
+      std::string checksum = packet_match[2].str();
+      std::string expected = computeChecksum(packet);
+      if (verbose_ && checksum != expected) {
+        std::cout << RED << "   Checksum failed\n"
+                  << "   Expected checksum: " << expected
+                  << "   , but received: " << checksum << std::endl
+                  << RESET;
+        // TODO: take action
```
FinnWilkinson, line 53: What should be the expected action for a failed checksum? i.e. abort and exit entirely?

src/tools/simeng/gdb/GDBStub.cc
```diff
@@ -0,0 +1,331 @@
...
+      } else if (packet[0] == 'Z' ||
+                 packet[0] == 'z') {  // create/remove breakpoint
+        std::regex break_regex(
+            "^([zZ])([0-4]),([0-9a-f]+),([0-9a-f]+)");  // format:
+                                                        // z/Z,type,address,kind
+        std::smatch break_match;
+        regex_match(packet, break_match, break_regex);
+        std::string type = break_match[2];
+        std::string address = break_match[3];
+        // std::string kind = break_match[4]; // kind not used
```
FinnWilkinson, line 94: If un-used then could remove this comment entirely, unless it provides some benefit to currently un-supported features?

src/tools/simeng/gdb/GDBStub.cc
```diff
@@ -0,0 +1,331 @@
...
+
+  while (!hitBreakpoint) {
+    core_.tick();
+    dataMemory_.tick();
+    for (std::string breakpoint : breakpoints_) {
+      if (hexToInt(breakpoint) == core_.getProgramCounter()) hitBreakpoint = 1;
+    }
+  }
+
+  return "S05";
```
FinnWilkinson, line 282: May be worth defining this string using `#define` as it is used in a few places

FinnWilkinson (changes requested):
Generally very good, just a few minor comments and clarifications

src/tools/simeng/gdb/GDBStub.cc
jj16791: @BraveSirRobbo we could have an additional optional config section for GDB and pass the config file through to the GDBStub to set such options? We wouldn't have to recompile then. I'd imagine, in time, we won't need such functionality as the implementation won't need debugging/fixing.

BraveSirRobbo: Supporting GDB's run/start commands would be able to support runtime flags, but I haven't managed to implement that (yet?).

I agree with this being a good simple alternative in the meanwhile though - would you like me to take a look at it ? I haven't dealt with the config files yet so I don't know how hard it would be for me

FinnWilkinson: If it is an option to pass runtime flags within GDB with the start/run command then I think that would be preferable for this verbose mode. So I think concentrating on getting them to work would be best (if possible)

src/tools/simeng/gdb/GDBStub.cc
BraveSirRobbo: it looks like breakpoint kinds could be used for thumb mode - is the line worth keeping in for usefulness later, perhaps with a thumb mention ?

src/tools/simeng/gdb/GDBStub.cc
BraveSirRobbo: I'm using it 3 times - what's the benefit of using a `#define` for it, and what would be a useful name for it ?

FinnWilkinson: If it needed changing in the future then it ensures that the value is changed in all places it is used. Although that may not be relevant to this exact case where this string probably won't change, it is good practice to do so.

As for a name, whatever "S05" represents within GDB

src/tools/simeng/gdb/GDBStub.cc
FinnWilkinson: Keep the comment, but it may be useful to include what it can be used for in the handover docs

src/tools/simeng/gdb/GDBStub.cc
BraveSirRobbo: this was easy enough to find out - respond with a "-" packet - fixed in a commit

src/tools/simeng/main.cc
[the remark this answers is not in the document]
BraveSirRobbo: nothing - removed in a later commit

src/tools/simeng/main.cc
BraveSirRobbo: thanks - added in a later commit

src/tools/simeng/main.cc
BraveSirRobbo: it is, fixed in a later commit

src/include/simeng/GDBStub.hh
BraveSirRobbo: thanks, fixed in a later commit

src/include/simeng/GDBStub.hh
BraveSirRobbo: thanks, fixed this and other similar examples in a later commit

src/include/simeng/GDBStub.hh
BraveSirRobbo: thanks, fixed this and other similar examples in a later commit

src/tools/simeng/gdb/GDBStub.cc
BraveSirRobbo: thanks, will do

src/tools/simeng/gdb/GDBStub.cc
BraveSirRobbo: cool - i'll do that.

I'm also using "OK" as a common response to GDB, should i `#define` that too ?

FinnWilkinson: Could be useful yes

src/tools/simeng/gdb/GDBStub.cc
```diff
@@ -0,0 +1,375 @@
...
+        regex_match(packet, break_match, break_regex);
+        std::string type = break_match[2];
+        std::string address = break_match[3];
+        // std::string kind = break_match[4]; // kind not used
+        if (packet[0] == 'Z')
+          handleCreateBreakpoint(type, address);
+        else
+          handleRemoveBreakpoint(type, address);
+        response += generateReply(GDB_OK);
+      } else if (packet[0] == 'M') {  // write memory
```
jj16791, line 110: Request from PR author, "remove the lines in GDBStub.cc that check for 'M' and 'G' due to lacking support".

src/tools/simeng/gdb/GDBStub.cc
```diff
@@ -0,0 +1,375 @@
...
+  // verbosely
+
+  int connection = openSocket(2424);  // change this if the port is blocked
+
+  std::cout << "[SimEng:GDBStub] Connection to GDB client established, "
+               "debugging in progress"
+            << std::endl
+            << std::endl;
+
+  char buffer[10000];
```
FinnWilkinson, line 19: Why has 10,000 been chosen? Is there a need for a buffer to be ~10KiB?

src/tools/simeng/gdb/GDBStub.cc
```diff
@@ -0,0 +1,375 @@
...
+
+  while (true) {
+    read(connection, buffer, 10000);
+
+    // First parentheses = package, second = checksum
+    std::regex packet_regex("^\\+*\\$([^#]*)#([0-9a-f]{2})");
```
FinnWilkinson, line 25: Would improve code readability if there was a brief explanation of the RegEx used

src/tools/simeng/gdb/GDBStub.cc
```diff
@@ -0,0 +1,375 @@
...
+    if (regex_match(bufferString, packet_match, packet_regex)) {
+      std::string packet = packet_match[1].str();
+      if (verbose_) {
+        std::cout << CYAN;
+        if (packet_match.size() == 3)
+          std::cout << "[SimEng:GDBStub] <- Received packet: " << packet
+                    << std::endl;
+        else
+          std::cout << RED << "[SimEng:GDBStub] <- Unknown message"
+                    << std::endl;  // TODO: take action
```
FinnWilkinson, line 51: If there is an error then the `Unknown Message` output should arguably be displayed whether verbose mode is enabled or disabled. (Unless it is common for `unknown message` errors?)

src/tools/simeng/gdb/GDBStub.cc
```diff
@@ -0,0 +1,375 @@
...
+  for (uint16_t i = 0; i < 32;
+       i++) {  // general purpose registers 0-31, 31 = stack pointer
+    uint64_t value = registers.get({0, i}).get<uint64_t>();
+    output += decToRSP(value);
+  }
+
+  output += decToRSP(core_.getProgramCounter());  // program counter
+
+  output += byteToHex(
+      registers.get({3, 0}).get<uint8_t>());  // NZCV (first 4 bits of cpsr)
```
FinnWilkinson, line 230: If possible, should use `NZCV` Register Type defined in AArch64/Instruction.hh

FinnWilkinson (changes requested):
Some changes needed to get this to 100%. Also, the GDBstub should be modified to always be ISA agnostic and then have architecture specific code for each ISA supported - similar to how the SimEng project does this.

src/tools/simeng/gdb/GDBStub.cc
BraveSirRobbo: The only reason I remember was just that the buffer would be long enough.  10,000 is excessive here.  The longest incoming message I see in current normal usage is 156 characters.  I expect this to rise when run/start is supported, but I suggest it could be 300 or so for now

src/tools/simeng/gdb/GDBStub.cc
BraveSirRobbo: I have included some explanations for this and other regexes (and added slightly tighter logic) in [this commit](https://github.com/UoB-HPC/SimEng/pull/244/commits/ec4a185f36b67eeffa59ae11d73c2830529a3c16)

src/tools/simeng/gdb/GDBStub.cc
BraveSirRobbo: I believe an ``unknown message`` should only occur if something changes or is wrong on the GDB side of things, as communication over TCP should receive the messages correctly.  This check is only a simple check on the message format.  I would be happy to remove this whole ``else`` statement if you are, or have it changed to an error through SimEng's ``ExceptionHandler``

jj16791:
Closing PR as the content does not reflect a stable implementation nor a minimal viable product.

jj16791:
Closing PR as the content does not reflect a stable implementation nor a minimal viable product.

PR #245, opened 2022-09-01
M1 Firestorm model

tomhepworth:
Contains my best performing config fire (m1_firestorm6.yaml from my documentation on branch `m1-dev`) and a port allocator class that aims to approximate the load balancing of the firestorm core's front-end.

### DIFF
--- a/src/include/simeng/pipeline/M1PortAllocator.hh
+++ b/src/include/simeng/pipeline/M1PortAllocator.hh
@@ -1,0 +1,63 @@
+#pragma once
+
+#include <cstdint>
+#include <vector>
+
+#include "simeng/pipeline/PortAllocator.hh"
+
+namespace simeng {
+namespace pipeline {
+
+/** A load-balancing port allocator implementation. Maintains demand weightings
+ * for each port, and allocates instructions to the suitable port with the
+ * lowest weighting. */
+class M1PortAllocator : public PortAllocator {
+ public:
+  /** Construct a load-balancing port allocator, providing a port arrangement
+   * specification. Each element of the port arrangement should represent a
+   * port, and contain a list of the instruction groups that port supports and
+   * a port type which denotes the matching requirements of said instruction
+   * groups. */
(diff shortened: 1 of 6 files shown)


src/include/simeng/pipeline/M1PortAllocator.hh
```diff
@@ -0,0 +1,63 @@
...
+  /** Allocate the lowest weighted port available for the specified instruction
+   * group. Returns the allocated port, and increases the weight of the port.
+   */
+  uint8_t allocate(const std::vector<uint8_t>& ports) override;
+
+  /** Decrease the weight for the specified port. */
+  void issued(uint8_t port) override;
+
+  /** Decrease the weight for the specified port. */
+  void deallocate(uint8_t port) override;
```
FinnWilkinson, line 33: Do these functions do the same thing? Or just duplicated comments?

src/lib/pipeline/M1PortAllocator.cc
```diff
@@ -0,0 +1,70 @@
...
+  return bestPort;
+}
+
+void M1PortAllocator::issued(uint8_t port) {
+  assert(weights[port] > 0);
+  weights[port]--;
+}
+
+void M1PortAllocator::deallocate(uint8_t port) { issued(port); };
+
```
FinnWilkinson, line 62: What is the need for both functions if one simply calls the other?

FinnWilkinson (changes requested):
Bit of confusion over `issued` and `deallocate` functions in the M1 port allocator

jj16791 (approved):

src/include/simeng/pipeline/M1PortAllocator.hh
FinnWilkinson: IGNORE - Clarified by Jack

src/lib/pipeline/M1PortAllocator.cc
FinnWilkinson: IGNORE - Clarified by Jack

FinnWilkinson (approved):

PR #246, opened 2022-09-02
Updated jenkins armclang build script to use new armclang22 compiler

jj16791:
The old armclang20 compiler we were using has an associated license which has now expired. The most recent armclang compilers don't need a compiler so the Isambard support team has installed the armclang22 compiler which we'll now use for our Jenkins CI.

I've added everyone with active or upcoming PRs as reviewers so you're all aware of this. You'll need to rebase `dev` onto your branches to pass the Jenkins armclang build pipeline.

### DIFF
--- a/.jenkins/build_arm22.sh
+++ b/.jenkins/build_arm22.sh
@@ -6,14 +6,13 @@ source .jenkins/build_test_run.sh
 checkout
 
 ## Setup environment
-module load cdt/20.03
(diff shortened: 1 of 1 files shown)


jj16791:
#rerun tests

FinnWilkinson (approved):

dANW34V3R (approved):

mutalibmohammed (approved):

PR #247, opened 2022-09-12
New core creation

jj16791:
A new CoreInstance class which creates the core and its associated memory interfaces. `main.cc` now creates an instance of this class and uses it to generate the SimEng simulation objects.

This change was influenced by the need to remove any dependencies SST had on yaml-cpp due to AppleClang having issues linking them. Through this class, the new SST core wrapper can now use the new CoreINstance class to create all SimEng objects without the need to link against the yaml-ccp external project.

### DIFF
--- a/configs/a64fx.yaml
+++ b/configs/a64fx.yaml
@@ -241,7 +241,7 @@ Latencies:
 CPU-Info:
   # Set Generate-Special-Dir to 'T' to generate the special files directory, or to 'F' to not.
   # (Not generating the special files directory may require the user to copy over files manually)
-  Generate-Special-Dir: T
+  Generate-Special-Dir: true
   # Core-Count MUST be 1 as multi-core is not supported at this time. (A64FX true value is 48)
   Core-Count: 1
   # Socket-Count MUST be 1 as multi-socket simulations are not supported at this time. (A64FX true value is 1)
--- a/configs/a64fx.yaml
+++ b/configs/a64fx.yaml
@@ -37,7 +37,11 @@ Branch-Predictor:
   Global-History-Length: 11
   RAS-entries: 8
   Fallback-Static-Predictor: "Always-Taken"
-L1-Cache:
+Data-Memory:
+  Interface-Type: Fixed
+Instruction-Memory:
+  Interface-Type: Flat
+LSQ-L1-Interface:
   Access-Latency: 5
   Exclusive: True
   Load-Bandwidth: 128
--- a/configs/a64fx.yaml
+++ b/configs/a64fx.yaml
@@ -37,9 +37,9 @@ Branch-Predictor:
   Global-History-Length: 11
   RAS-entries: 8
   Fallback-Static-Predictor: "Always-Taken"
-Data-Memory:
+L1-Data-Memory:
   Interface-Type: Fixed
-Instruction-Memory:
+L1-Instruction-Memory:
   Interface-Type: Flat
 LSQ-L1-Interface:
   Access-Latency: 5
@@ -243,9 +243,9 @@ Latencies:
 # CPU-Info mainly used to generate a replica of the special (or system) file directory 
 # structure
 CPU-Info:
-  # Set Generate-Special-Dir to 'T' to generate the special files directory, or to 'F' to not.
+  # Set Generate-Special-Dir to True to generate the special files directory, or to False to not.
   # (Not generating the special files directory may require the user to copy over files manually)
-  Generate-Special-Dir: true
+  Generate-Special-Dir: True
   # Core-Count MUST be 1 as multi-core is not supported at this time. (A64FX true value is 48)
   Core-Count: 1
   # Socket-Count MUST be 1 as multi-socket simulations are not supported at this time. (A64FX true value is 1)
--- a/src/include/simeng/CoreInstance.hh
+++ b/src/include/simeng/CoreInstance.hh
@@ -35,9 +35,6 @@ class CoreInstance {
   /** Constructor with a config file path. */
   CoreInstance(std::string configPath);
 
-  /** Constructor with a pre-constructed config file. */
-  CoreInstance(YAML::Node config);
-
   ~CoreInstance();
 
   /** Construct the SimEng linux process object from an executable ELF file and
--- a/src/include/simeng/CoreInstance.hh
+++ b/src/include/simeng/CoreInstance.hh
@@ -84,10 +84,10 @@ class CoreInstance {
   std::unique_ptr<simeng::kernel::LinuxProcess> process_ = nullptr;
 
   /** The size of the process memory. */
-  size_t processMemorySize_;
+  uint64_t processMemorySize_;
 
   /** The process memory space. */
-  char* processMemory_;
+  std::shared_ptr<char> processMemory_;
 
   /** The SimEng kernel object. */
   simeng::kernel::Linux kernel_;
--- a/src/include/simeng/CoreInstance.hh
+++ b/src/include/simeng/CoreInstance.hh
@@ -43,13 +43,15 @@ enum class SimulationMode { Emulation, InOrderPipelined, OutOfOrder };
 /** A class to create a SimEng core instance from a supplied config. */
 class CoreInstance {
  public:
-  /** Default constructor with command line arguments but no passed
+  /** Default constructor with an executable and its arguments but no model
    * configuration. */
-  CoreInstance(int argc, char** argv);
+  CoreInstance(std::string executablePath,
+               std::vector<std::string> executableArgs);
 
-  /** Constructor with with command line arguments and a configuration file
-   * path. */
-  CoreInstance(int argc, char** argv, std::string configPath);
+  /** Constructor with an executable, its arguments, and a model configuration.
+   */
+  CoreInstance(std::string configPath, std::string executablePath,
+               std::vector<std::string> executableArgs);
 
   ~CoreInstance();
 
@@ -87,7 +89,8 @@ class CoreInstance {
  private:
   /** Generate the appropriate simulation objects as parameterised by the
    * configuration.*/
-  void generateCoreModel(int argc, char** argv);
+  void generateCoreModel(std::string executablePath,
+                         std::vector<std::string> executableArgs);
 
   /** Extract simulation mode from config file. */
   void setSimulationMode();
@@ -95,7 +98,8 @@ class CoreInstance {
   /** Construct the SimEng linux process object from command line arguments.
    * Empty command line arguments denote the usage of hardcoded
    * instructions held in the hex_ array. */
-  void createProcess(int argc, char** argv);
+  void createProcess(std::string executablePath,
+                     std::vector<std::string> executableArgs);
 
   /** Construct the process memory from the generated process_ object. */
   void createProcessMemory();
--- a/src/include/simeng/CoreInstance.hh
+++ b/src/include/simeng/CoreInstance.hh
@@ -22,10 +22,8 @@
 #include "simeng/pipeline/BalancedPortAllocator.hh"
 #include "yaml-cpp/yaml.h"
 
-// Out-of-order test; counts down from 1024*1024, with an independent `orr`
-// at the start of each branch. With an instruction latency of 2 or greater,
-// the `orr` at the start of the next loop should issue/execute while the
-// preceding branch is waiting on the result from the `subs`.
+// Instruction set used when no executable is provided; counts down from
+// 1024*1024, with an independent `orr` at the start of each branch.
 uint32_t hex_[] = {
     0x320C03E0,  // orr w0, wzr, #1048576
     0x320003E1,  // orr w0, wzr, #1
@@ -67,6 +65,9 @@ class CoreInstance {
 
   /** Getter for the set simulation mode. */
   const SimulationMode getSimulationMode() const;
+
+  /** Getter for the set simulation mode in a string format. */
+  const std::string getSimulationModeString() const;
 
   /** Getter for the create core object. */
   std::shared_ptr<simeng::Core> getCore() const;
@@ -123,9 +124,6 @@ class CoreInstance {
   /** The SimEng Linux kernel object. */
   simeng::kernel::Linux kernel_;
 
-  /** Whether or not the createCore() function must be manually called. */
-  bool manualCreateCore_ = false;
-
   /** Whether or not the dataMemory_ must be set manually. */
   bool setDataMemory_ = false;
 
@@ -146,6 +144,10 @@ class CoreInstance {
 
   /** The simulation mode in use, defaulting to emulation. */
   SimulationMode mode_ = SimulationMode::Emulation;
+
+  /** A string format for the simulation mode in use, defaulting to emulation.
+   */
+  std::string modeString_ = "Emulation";
 
   /** Reference to the SimEng data memory object. */
   std::shared_ptr<simeng::MemoryInterface> dataMemory_ = nullptr;
--- a/src/include/simeng/CoreInstance.hh
+++ b/src/include/simeng/CoreInstance.hh
@@ -22,7 +22,7 @@
 #include "simeng/pipeline/BalancedPortAllocator.hh"
 #include "yaml-cpp/yaml.h"
 
-// Instruction set used when no executable is provided; counts down from
+// Program used when no executable is provided; counts down from
 // 1024*1024, with an independent `orr` at the start of each branch.
 uint32_t hex_[] = {
     0x320C03E0,  // orr w0, wzr, #1048576
--- a/src/include/simeng/CoreInstance.hh
+++ b/src/include/simeng/CoreInstance.hh
@@ -67,6 +67,12 @@ class CoreInstance {
   /** Getter for the set simulation mode. */
   const SimulationMode getSimulationMode() const;
 
+  /** Getter for a shared pointer to the created process image. */
+  std::shared_ptr<char> getProcessImage() const;
+
+  /** Getter for the size of the created process image. */
+  const uint64_t getProcessImageSize() const;
+
  private:
   /** Extract simulation mode from config file. */
   void setSimulationMode();
--- a/src/include/simeng/CoreInstance.hh
+++ b/src/include/simeng/CoreInstance.hh
@@ -24,6 +24,7 @@
 
 namespace simeng {
 
+/** The available modes of simulation. */
 enum class SimulationMode { Emulation, InOrderPipelined, OutOfOrder };
 
 /** A class to create a SimEng core instance from a supplied config. */
@@ -58,7 +59,8 @@ class CoreInstance {
   /** Set the SimEng L1 data cache memory. */
   void setL1DataMemory(std::shared_ptr<simeng::MemoryInterface> memRef);
 
-  /** Construct the core and all its associated simulation objects. */
+  /** Construct the core and all its associated simulation objects after the
+   * process and memory interfaces have been instantiated. */
   std::shared_ptr<simeng::Core> createCore();
 
   /** Construct the special file directory. */
@@ -92,7 +94,7 @@ class CoreInstance {
   /** The process memory space. */
   std::shared_ptr<char> processMemory_;
 
-  /** The SimEng kernel object. */
+  /** The SimEng Linux kernel object. */
   simeng::kernel::Linux kernel_;
 
   /** Reference to the SimEng architecture object. */
--- a/src/include/simeng/CoreInstance.hh
+++ b/src/include/simeng/CoreInstance.hh
@@ -1,0 +1,129 @@
+#pragma once
+
+#include <string>
+
+#include "simeng/AlwaysNotTakenPredictor.hh"
+#include "simeng/Core.hh"
+#include "simeng/Elf.hh"
+#include "simeng/FixedLatencyMemoryInterface.hh"
+#include "simeng/FlatMemoryInterface.hh"
+#include "simeng/GenericPredictor.hh"
+#include "simeng/ModelConfig.hh"
+#include "simeng/SpecialFileDirGen.hh"
+#include "simeng/arch/Architecture.hh"
+#include "simeng/arch/aarch64/Architecture.hh"
+#include "simeng/arch/aarch64/Instruction.hh"
+#include "simeng/arch/aarch64/MicroDecoder.hh"
+#include "simeng/kernel/Linux.hh"
+#include "simeng/models/emulation/Core.hh"
+#include "simeng/models/inorder/Core.hh"
+#include "simeng/models/outoforder/Core.hh"
+#include "simeng/pipeline/A64FXPortAllocator.hh"
+#include "simeng/pipeline/BalancedPortAllocator.hh"
+#include "yaml-cpp/yaml.h"
+
+namespace simeng {
+
+enum class SimulationMode { Emulation, InOrderPipelined, OutOfOrder };
+
+/** A class to create a SimEng core instance from a supplied config. */
+class CoreInstance {
+ public:
+  /** Default constructor with a no passed configuration. */
+  CoreInstance();
+
+  /** Constructor with a config file path. */
+  CoreInstance(std::string configPath);
+
+  /** Constructor with a pre-constructed config file. */
+  CoreInstance(YAML::Node config);
+
+  ~CoreInstance();
+
+  /** Construct the SimEng linux process object from an executable ELF file and
+   * its command-line arguments. */
+  void createProcess(const std::vector<std::string>& commandLine);
+
+  /** Construct the SimEng linux process object from an span of instructions. */
+  void createProcess(span<char> instructions);
+
+  /** Construct the SimEng L1 instruction cache memory. */
+  std::shared_ptr<simeng::MemoryInterface> createL1InstructionMemory(
+      const simeng::MemInterfaceType type);
+
+  /** Set the SimEng L1 instruction cache memory. */
+  void setL1InstructionMemory(std::shared_ptr<simeng::MemoryInterface> memRef);
+
+  /** Construct the SimEng L1 data cache memory. */
+  std::shared_ptr<simeng::MemoryInterface> createL1DataMemory(
+      const simeng::MemInterfaceType type);
+
+  /** Set the SimEng L1 data cache memory. */
+  void setL1DataMemory(std::shared_ptr<simeng::MemoryInterface> memRef);
+
+  /** Construct the core and all its associated simulation objects. */
+  std::shared_ptr<simeng::Core> createCore();
+
+  /** Construct the special file directory. */
+  void createSpecialFileDirectory();
+
+  /** Getter for the set CPU clock speed. */
+  const float getClockFrequency() const;
+
+  /** Getter for the set virtual counter timer frequency. */
+  const uint32_t getTimerFrequency() const;
+
+  /** Getter for the set simulation mode. */
+  const SimulationMode getSimulationMode() const;
+
+ private:
+  /** Extract simulation mode from config file. */
+  void setSimulationMode();
+
+  /** Construct the process memory from the generated process_ object. */
+  void createProcessMemory();
+
+  /** The config file describing the modelled core to be created. */
+  YAML::Node config_;
+
+  /** Reference to the SimEng linux process object. */
+  std::unique_ptr<simeng::kernel::LinuxProcess> process_ = nullptr;
+
+  /** The size of the process memory. */
+  size_t processMemorySize_;
+
+  /** The process memory space. */
+  char* processMemory_;
+
+  /** The SimEng kernel object. */
+  simeng::kernel::Linux kernel_;
+
+  /** Reference to the SimEng architecture object. */
+  std::unique_ptr<simeng::arch::Architecture> arch_ = nullptr;
+
+  /** Reference to the SimEng branch predictor object. */
+  std::unique_ptr<simeng::BranchPredictor> predictor_ = nullptr;
+
+  /** Reference to the SimEng port allocator object. */
+  std::unique_ptr<simeng::pipeline::PortAllocator> portAllocator_ = nullptr;
+
+  /** Reference to the SimEng core object. */
+  std::shared_ptr<simeng::Core> core_ = nullptr;
+
+  /** The simulation mode in use, defaulting to emualtion. */
+  SimulationMode mode_ = SimulationMode::Emulation;
+
+  /** Reference to the SimEng instruction memory object. */
+  std::shared_ptr<simeng::MemoryInterface> instructionMemory_ = nullptr;
+
+  /** Reference to the SimEng data memory object. */
+  std::shared_ptr<simeng::MemoryInterface> dataMemory_ = nullptr;
+
+  /** CPU clock speed. */
+  float clockFreq_ = 0;
+
+  /** Virtual counter timer frequency. */
+  uint32_t timerFreq_ = 0;
+};
+
+}  // namespace simeng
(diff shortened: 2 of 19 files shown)


dANW34V3R (commented):
A couple of suggested changes. Will review again once #243 goes through

src/include/simeng/CoreInstance.hh
```diff
@@ -0,0 +1,129 @@
...
+class CoreInstance {
+ public:
+  /** Default constructor with a no passed configuration. */
+  CoreInstance();
+
+  /** Constructor with a config file path. */
+  CoreInstance(std::string configPath);
+
+  /** Constructor with a pre-constructed config file. */
+  CoreInstance(YAML::Node config);
```
rahahahat, line 39: What is the use case for this constructor? If the configPath is provided then CoreInstance::CoreInstance(std::string configPath) will handle that case. If it isn't the default config is loaded in CoreInstance::CoreInstance(). Is this constructor used anywhere?

I'm not sure about this, so please correct me if I'm wrong but wouldn't declaring a public constructor with the YAML::Node argument make it necessary for every library/executable which uses CoreInstance to also link yaml-cpp explicitly? In which case it would still require us to link yaml-cpp to the SST build?

src/tools/simeng/main.cc
```diff
@@ -197,107 +162,46 @@ int main(int argc, char** argv) {
...
-  // Extract number of ports
-  for (size_t i = 0; i < config_ports.size(); i++) {
-    auto config_groups = config_ports[i]["Instruction-Group-Support"];
-    // Extract number of groups in port
-    for (size_t j = 0; j < config_groups.size(); j++) {
-      portArrangement[i].push_back(config_groups[j].as<uint16_t>());
-    }
+  // Get simualtion mode and data memory interface type
+  simeng::SimulationMode mode = coreInstance->getSimulationMode();
+  simeng::MemInterfaceType L1Dtype = simeng::MemInterfaceType::Flat;
```
rahahahat, line 171: Why can't we have MemInterfaceType member variable inside `CoreInstance` which gets populated in the Constructor, given that we already get `SimulationMode` using the `getSimulationMode` getter? And expose it (if necessary) via a getter?

src/tools/simeng/main.cc
```diff
@@ -197,107 +162,46 @@ int main(int argc, char** argv) {
...
-      }
-      rsArrangement[port] = {i, reservation_station["Size"].as<uint16_t>()};
-    }
+  if (mode == simeng::SimulationMode::InOrderPipelined) {
+    modeString = "In-Order Pipelined";
   }
 
-  int iterations = 0;
+  // Create memory interfaces
+  std::shared_ptr<simeng::MemoryInterface> dataMemory =
```
rahahahat, line 182: Also since we already do have a `setL1DataMemory function` which allows us to set any memory interface shouldn't`createMemoryL1DataMemory` create a memory interface associated with a mode? e.g OoO -> Fixed

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,240 @@
...
+
+void CoreInstance::setL1DataMemory(
+    std::shared_ptr<simeng::MemoryInterface> memRef) {
+  // Set the L1D cache instance to use
+  dataMemory_ = memRef;
+  return;
+}
+
+std::shared_ptr<simeng::Core> CoreInstance::createCore() {
+  // Ensure all appropriate creation functions have been called
```
rahahahat, line 148: Are the prerequisites for `createCore` documented anywhere, so that the developer knows what order to call these functions in?

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,240 @@
...
+    std::cerr << "Incompatible instruction memory interface type requested"
+              << std::endl;
+    exit(1);
+  }
+
+  // Create a L1I cache instance based on type supplied
+  if (type == simeng::MemInterfaceType::Flat) {
+    instructionMemory_ = std::make_shared<simeng::FlatMemoryInterface>(
+        processMemory_, processMemorySize_);
+  } else if (type == simeng::MemInterfaceType::Fixed) {
```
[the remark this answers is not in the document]
jj16791: It was kept for future compatibility once we consider an in-house cache solution

src/include/simeng/CoreInstance.hh
jj16791: The comments made are correct. Forgot to push commit which removed this constructor. It was initially used by the regression test suite but I found it was better to forgo the use of the CoreInstance class and use the prior core creation method.

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,240 @@
...
+}
+
+std::shared_ptr<simeng::MemoryInterface> CoreInstance::createL1DataMemory(
+    const simeng::MemInterfaceType type) {
+  // Currently, if the core in use is emulation or in-order, only a flat data
+  // memory interface can be used
+  if (mode_ == SimulationMode::Emulation ||
+      mode_ == SimulationMode::InOrderPipelined) {
+    if (type != simeng::MemInterfaceType::Flat) {
+      std::cerr << "Incompatible instruction memory interface type requested"
```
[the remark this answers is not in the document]
jj16791: Error messages altered.

src/lib/CoreInstance.cc
jj16791: No as of yet. It's too early in the development pipeline of this new feature to consider writing documentation.

src/tools/simeng/main.cc
jj16791: This would reduce the flexibility of the possible configurations. Whilst we currently always use a set memory interface type for each simulation mode, there's nothing to say we wouldn't use differing/new interface types in the future.

src/tools/simeng/main.cc
jj16791: I believe https://github.com/UoB-HPC/SimEng/pull/247#discussion_r970681853 answers this query.

src/tools/simeng/main.cc
```diff
@@ -182,109 +150,42 @@ int main(int argc, char** argv) {
...
-  for (size_t i = 0; i < config["Reservation-Stations"].size(); i++) {
-    auto reservation_station = config["Reservation-Stations"][i];
-    for (size_t j = 0; j < reservation_station["Ports"].size(); j++) {
-      uint8_t port = reservation_station["Ports"][j].as<uint8_t>();
-      if (rsArrangement.size() < port + 1) {
-        rsArrangement.resize(port + 1);
-      }
-      rsArrangement[port] = {i, reservation_station["Size"].as<uint16_t>()};
-    }
+  if (mode == simeng::SimulationMode::InOrderPipelined) {
```
jj16791, line 165: else if

jj16791 (commented):
Need to add getters for process image shared pointer and its size

src/include/simeng/CoreInstance.hh
```diff
@@ -0,0 +1,114 @@
...
+  std::unique_ptr<simeng::kernel::LinuxProcess> process_ = nullptr;
+
+  /** The size of the process memory. */
+  uint64_t processMemorySize_;
+
+  /** The process memory space. */
+  std::shared_ptr<char> processMemory_;
+
+  /** The SimEng kernel object. */
+  simeng::kernel::Linux kernel_;
```
FinnWilkinson, line 96: Have comment as `The SimEng *Linux* kernel object` instead?

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,229 @@
...
+
+  // Create a L1I cache instance based on type supplied
+  if (type == simeng::MemInterfaceType::Flat) {
+    instructionMemory_ = std::make_shared<simeng::FlatMemoryInterface>(
+        processMemory_.get(), processMemorySize_);
+  } else if (type == simeng::MemInterfaceType::Fixed) {
+    instructionMemory_ = std::make_shared<simeng::FixedLatencyMemoryInterface>(
+        processMemory_.get(), processMemorySize_,
+        config_["L1-Cache"]["Access-Latency"].as<uint16_t>());
+  }
```
FinnWilkinson, line 93: No need for this `else if` as type must be Flat to get to this point

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,229 @@
...
+  if (instructionMemory_ == nullptr) {
+    std::cerr << "Instruction memory not instantiated or set" << std::endl;
+    exit(1);
+  } else if (dataMemory_ == nullptr) {
+    std::cerr << "Data memory not instantiated or set" << std::endl;
+    exit(1);
+  } else if (process_ == nullptr) {
+    std::cerr << "Process not instantiated or set" << std::endl;
+    exit(1);
+  }
```
FinnWilkinson, line 155: I'm trying to reason about why this is needed over just using the object's constructor.

Sureley we could pass the  process's path string to the constructor and memory types can be inferred from the config file rather than having to make so many function calls? And as for default program, this could be defined in a header file to access as needed (if no string is passed to constructor).

src/tools/simeng/main.cc
```diff
@@ -182,109 +150,42 @@ int main(int argc, char** argv) {
...
-    // Remove any current special files dir
-    SFdir.RemoveExistingSFDir();
-    // Create new special files dir
-    SFdir.GenerateSFDir();
-  }
+  // Create core
+  std::shared_ptr<simeng::Core> core = coreInstance->createCore();
 
+  // Create Special Files directory if indicated to do so in Config
+  coreInstance->createSpecialFileDirectory();
```
FinnWilkinson, line 178: Could the coreInstance not call this itself?

FinnWilkinson (commented):
Some confusion over current design decisions

src/include/simeng/CoreInstance.hh
```diff
@@ -0,0 +1,114 @@
...
+#include "simeng/models/emulation/Core.hh"
+#include "simeng/models/inorder/Core.hh"
+#include "simeng/models/outoforder/Core.hh"
+#include "simeng/pipeline/A64FXPortAllocator.hh"
+#include "simeng/pipeline/BalancedPortAllocator.hh"
+#include "yaml-cpp/yaml.h"
+
+namespace simeng {
+
+enum class SimulationMode { Emulation, InOrderPipelined, OutOfOrder };
```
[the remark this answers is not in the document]
jj16791: Is there a set format for enum/enum class DOXYGEN comments. We don't have any comments on any of these in the project yet.

src/include/simeng/CoreInstance.hh
```diff
@@ -0,0 +1,114 @@
...
+
+  /** Construct the SimEng L1 data cache memory. */
+  std::shared_ptr<simeng::MemoryInterface> createL1DataMemory(
+      const simeng::MemInterfaceType type);
+
+  /** Set the SimEng L1 data cache memory. */
+  void setL1DataMemory(std::shared_ptr<simeng::MemoryInterface> memRef);
+
+  /** Construct the core and all its associated simulation objects. */
+  std::shared_ptr<simeng::Core> createCore();
```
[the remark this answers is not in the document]
jj16791: Altered comment.

src/include/simeng/CoreInstance.hh
```diff
@@ -0,0 +1,114 @@
...
+  /** Set the SimEng L1 instruction cache memory. */
+  void setL1InstructionMemory(std::shared_ptr<simeng::MemoryInterface> memRef);
+
+  /** Construct the SimEng L1 data cache memory. */
+  std::shared_ptr<simeng::MemoryInterface> createL1DataMemory(
+      const simeng::MemInterfaceType type);
+
+  /** Set the SimEng L1 data cache memory. */
+  void setL1DataMemory(std::shared_ptr<simeng::MemoryInterface> memRef);
+
```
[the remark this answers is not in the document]
jj16791: SimEng with SST has SST create the L1D cache as an SST mem interface hence the need for a setter as well as a "constructor"

src/include/simeng/MemoryInterface.hh
```diff
@@ -5,6 +5,8 @@
 
 namespace simeng {
 
+enum class MemInterfaceType { Flat, Fixed };
+
```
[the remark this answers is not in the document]
jj16791: see https://github.com/UoB-HPC/SimEng/pull/247#discussion_r970887480

src/tools/simeng/main.cc
jj16791: I can't remember which one of the interns had this, but they needed to not create the special file directory for some reason. Implemented it for scenarios equivalent to that.

jj16791: It may have been something to do with it being created in the default install directory which they don't have access to so maybe we could remove it as that's no the case anymore. Not sure if you can recall the issue better?

src/include/simeng/CoreInstance.hh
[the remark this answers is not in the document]
FinnWilkinson: From : https://doxygen.nl/manual/docblocks.html

    /** An enum type. 
     *  The documentation block cannot be put after the enum! 
     */
    enum EnumType
    {
      int EVal1,     /**< enum value 1 */
      int EVal2      /**< enum value 2 */
    };

src/lib/CoreInstance.cc
FinnWilkinson: After previous comment (https://github.com/UoB-HPC/SimEng/pull/247#discussion_r970890787) I understand that this decision comes from SST compatability. Setting as resolved, no clarification needed. #

src/tools/simeng/main.cc
FinnWilkinson: This functionality can be set inside the configuration file already

FinnWilkinson: And yes, as the creation directory has been changed there should not be any access issues anymore

src/lib/CoreInstance.cc
jj16791: So are you suggesting we have extra config file params for defining the memory hierarchy which can be used to configure the memory objects within CoreInstance? Not a bad plan and would work well with future in-house cache solutions.

For the process image, we could do something equivalent to that. May be better to keep two separate constructors to avoid complications between an incorrect executable path of `""` and a way to denote 'use hex span'.

src/include/simeng/CoreInstance.hh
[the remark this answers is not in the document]
jj16791: Okay good, we can add those. I'll add them to our other enum/enum class objects and just couple them in with this PR? Or do you recon another PR for those comments?

src/tools/simeng/main.cc
jj16791: I'll move out of its own function then.

src/include/simeng/CoreInstance.hh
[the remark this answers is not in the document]
FinnWilkinson: Does this mean that SimEng can only currently use SST for L1D cache? i.e. no L2 etc?

src/include/simeng/CoreInstance.hh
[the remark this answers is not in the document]
FinnWilkinson: This PR is fine

src/lib/CoreInstance.cc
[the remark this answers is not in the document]
dANW34V3R: A comment would be good here explaining that then

src/lib/CoreInstance.cc
jj16791: Resolving due to the significant changes made in the latest commits. Please re-review.

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,304 @@
...
+              << std::endl;
+    exit(1);
+  }
+
+  // Construct architecture object
+  arch_ =
+      std::make_unique<simeng::arch::aarch64::Architecture>(kernel_, config_);
+
+  // Construct branch predictor object
+  predictor_ = std::make_unique<simeng::GenericPredictor>(config_);
```
FinnWilkinson, line 197: Given there are multiple branch predictor classes available, could a config option be introduced to define which one is used?

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,304 @@
...
+  auto config_ports = config_["Ports"];
+  std::vector<std::vector<uint16_t>> portArrangement(config_ports.size());
+  for (size_t i = 0; i < config_ports.size(); i++) {
+    auto config_groups = config_ports[i]["Instruction-Group-Support"];
+    // Read groups in associated port
+    for (size_t j = 0; j < config_groups.size(); j++) {
+      portArrangement[i].push_back(config_groups[j].as<uint16_t>());
+    }
+  }
+  portAllocator_ = std::make_unique<simeng::pipeline::BalancedPortAllocator>(
```
FinnWilkinson, line 209: Given there are multiple port allocator classes available, could a config option be introduced to define which one is used?

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,304 @@
...
+        *arch_, *predictor_, *portAllocator_, rsArrangement, config_);
+  }
+
+  createSpecialFileDirectory();
+
+  return;
+}
+
+void CoreInstance::createSpecialFileDirectory() {
+  simeng::SpecialFileDirGen SFdir = simeng::SpecialFileDirGen(config_);
```
FinnWilkinson, line 252: This can be done inside the if clause - no point constructing an object that may not be used

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,304 @@
...
+    SFdir.GenerateSFDir();
+  }
+
+  return;
+}
+
+const SimulationMode CoreInstance::getSimulationMode() const { return mode_; }
+
+std::shared_ptr<simeng::Core> CoreInstance::getCore() const {
+  if (manualCreateCore_ && core_ == nullptr) {
```
FinnWilkinson, line 267: Again, parentheses around the nullptr check may help readability

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,304 @@
...
+                 "`External` type, they must be set manually and then core's "
+                 "creation must be called manually."
+              << std::endl;
+    exit(1);
+  }
+  return core_;
+}
+
+std::shared_ptr<simeng::MemoryInterface> CoreInstance::getDataMemory() const {
+  if (setDataMemory_) {
```
FinnWilkinson, line 280: It seems this clause will always trigger if external dataMemory is to be used? Which I presume is incorrect?

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,304 @@
...
+    std::cerr << "`External` data memory object not set." << std::endl;
+    exit(1);
+  }
+  return dataMemory_;
+}
+
+std::shared_ptr<simeng::MemoryInterface> CoreInstance::getInstructionMemory()
+    const {
+  if (setInstructionMemory_) {
```
FinnWilkinson, line 289: It seems this clause will always trigger if external instructionMemory is to be used? Which I presume is incorrect?

src/tools/simeng/main.cc
```diff
@@ -58,233 +41,39 @@ int main(int argc, char** argv) {
...
-    // };
-
-    // Some arbitrary values to sort
-    // std::vector<int> memoryValues = {9,  6, 7, 20, 5,   0,  80, 2,  1,  6,
-    //                                  17, 4, 3, 22, 117, 11, 4,  12, 10, 18};
-    // memcpy(memory, memoryValues.data(), memoryValues.size() * sizeof(int));
-
-    process = std::make_unique<simeng::kernel::LinuxProcess>(
-        simeng::span<char>(reinterpret_cast<char*>(hex), sizeof(hex)), config);
+    coreInstance = std::make_unique<simeng::CoreInstance>(argc - 1, argv);
```
FinnWilkinson, line 50: Some comments about why `argc - 2` and `argv + 2` are performed may be useful, as it is not clear what is passed to each constructor

configs/a64fx.yaml
```diff
@@ -241,7 +245,7 @@ Latencies:
 CPU-Info:
   # Set Generate-Special-Dir to 'T' to generate the special files directory, or to 'F' to not.
   # (Not generating the special files directory may require the user to copy over files manually)
-  Generate-Special-Dir: T
+  Generate-Special-Dir: true
```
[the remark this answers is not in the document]
jj16791: Doesn't seem to matter from a YAML-CPP perspective but can change for consistency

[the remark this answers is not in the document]
FinnWilkinson: Just noticed the comment above - this should be updated also

configs/a64fx.yaml
```diff
@@ -37,7 +37,11 @@ Branch-Predictor:
   Global-History-Length: 11
   RAS-entries: 8
   Fallback-Static-Predictor: "Always-Taken"
-L1-Cache:
+Data-Memory:
+  Interface-Type: Fixed
+Instruction-Memory:
+  Interface-Type: Flat
```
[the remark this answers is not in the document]
jj16791: I guess for now no. With SST we may not have an L1 cache for example. May find a better name in the future with a more complex memory hierarchy (e.g. first-level-cache).

I'll change to L1 for now as it's a better representation.

src/lib/ModelConfig.cc
```diff
@@ -135,8 +135,39 @@ void ModelConfig::validate() {
   }
   subFields.clear();
 
-  // L1-Cache
-  root = "L1-Cache";
+  // Instruction Memory
+  root = "Instruction-Memory";
+  subFields = {"Interface-Type"};
+  nodeChecker<std::string>(
+      configFile_[root][subFields[0]], root + " " + subFields[0],
```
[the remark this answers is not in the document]
jj16791: Not quite sure what you mean.

There are two Interface-Type fields in each config file, hence, we need a prefix to tell the user which one it's referencing. Hopefully that answers your query.

src/lib/CoreInstance.cc
jj16791: We could. I think such functionality would be better suited to an upcoming PR that targets Branch Predication/speculation in general.

src/lib/CoreInstance.cc
jj16791: Similar to https://github.com/UoB-HPC/SimEng/pull/247#discussion_r972132103, good idea but out of scope.

src/lib/ModelConfig.cc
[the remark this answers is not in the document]
FinnWilkinson: If you look towards another nodeChecker<string> call, i.e. simulation mode, the second argument is `subFields[0]` rather than `root + " " + subFields[0]`. Why is it different for the interface types?

configs/a64fx.yaml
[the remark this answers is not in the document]
jj16791: Good spot.

src/lib/ModelConfig.cc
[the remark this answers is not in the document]
jj16791: > There are two Interface-Type fields in each config file, hence, we need a prefix to tell the user which one it's referencing. Hopefully that answers your query.

We need the extra root prefix to distinguish the two config params. Taking your example of "simulation mode", there only exists one "simulation mode" param and thus doesn't need the extra prefix.

src/lib/CoreInstance.cc
jj16791: Yep, forgot to check for it being equal to `nullptr`.

src/lib/ModelConfig.cc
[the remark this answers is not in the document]
FinnWilkinson: Makes sense, thanks for clarification

FinnWilkinson (approved):
LGTM

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,304 @@
+#include "simeng/CoreInstance.hh"
+
+namespace simeng {
+
+CoreInstance::CoreInstance(int argc, char** argv) {
+  config_ = YAML::Load(DEFAULT_CONFIG);
+  generateCoreModel(argc, argv);
+}
+
+CoreInstance::CoreInstance(int argc, char** argv, std::string configPath) {
```
rahahahat, line 10: nit: I think the values of argc and argv should be the same as what is passed to main because it is a bit confusing to reason about how `argv[0]` ([here](https://github.com/UoB-HPC/SimEng/pull/247/files#diff-560f14653baa5d07d791071914951617cde676e28ce946ad75f46ffc2ad115dfR82)) in `createProcess` is the executable path and not `./simeng`  unless you look at `main.cc` where the core constructor is invoked as `core(argc-2, argv+2, configPath)`. I think the `if else` clause [here](https://github.com/UoB-HPC/SimEng/pull/247/files#diff-8b47da0924c9ef7fd9025a5091b26a82a6399866b1eee24da80829fb5538870bL65) can be handled inside `generateCoreModel` and then vector a can be passed to `createProcess` in which index 0 is the path to the executable binary,

dANW34V3R (commented):
A couple of minor comments

src/tools/simeng/main.cc
```diff
@@ -44,144 +44,33 @@ int main(int argc, char** argv) {
   // Create the instance of the core to be simulated
   std::unique_ptr<simeng::CoreInstance> coreInstance;
   if (argc > 1) {
-    coreInstance = std::make_unique<simeng::CoreInstance>(argv[1]);
+    coreInstance =
+        std::make_unique<simeng::CoreInstance>(argc - 2, argv + 2, argv[1]);
```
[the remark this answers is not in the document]
dANW34V3R: Already changed

src/include/simeng/CoreInstance.hh
```diff
@@ -0,0 +1,157 @@
...
+#include "simeng/models/inorder/Core.hh"
+#include "simeng/models/outoforder/Core.hh"
+#include "simeng/pipeline/A64FXPortAllocator.hh"
+#include "simeng/pipeline/BalancedPortAllocator.hh"
+#include "yaml-cpp/yaml.h"
+
+// Out-of-order test; counts down from 1024*1024, with an independent `orr`
+// at the start of each branch. With an instruction latency of 2 or greater,
+// the `orr` at the start of the next loop should issue/execute while the
+// preceding branch is waiting on the result from the `subs`.
```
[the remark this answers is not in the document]
jj16791: Poor comment wording, meant to act as the default instruction span input. Will change.

src/tools/simeng/main.cc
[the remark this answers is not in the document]
jj16791: Based on old code, please review the most up-to-date code

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,304 @@
...
+void CoreInstance::createProcess(int argc, char** argv) {
+  // Check for passed executable
+  std::string executablePath = "";
+  if (argc != 0) {
+    executablePath = std::string(argv[0]);
+  }
+
+  if (executablePath.length() > 0) {
+    // Attempt to create the process image from the specified command-line
+    std::vector<std::string> commandLine(argv, argv + argc);
```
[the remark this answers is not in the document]
jj16791: Arguments are set to be `config path` `executable` `executable args{}`. No plan to change this.

src/lib/CoreInstance.cc
dANW34V3R: I agree

src/lib/CoreInstance.cc
```diff
@@ -0,0 +1,304 @@
...
+        processMemory_.get(), processMemorySize_,
+        config_["LSQ-L1-Interface"]["Access-Latency"].as<uint16_t>());
+  } else {
+    std::cerr << "Unsupported memory interface type used in "
+                 "createL1InstructionMemory()."
+              << std::endl;
+    exit(1);
+  }
+
+  return;
```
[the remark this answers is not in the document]
jj16791: Added to all void funcs for consistency within the class.

src/lib/CoreInstance.cc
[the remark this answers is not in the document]
dANW34V3R: If that is the case, should this be changed to argv + 2 with appropriate comment. +argc is a bit confusing

src/lib/CoreInstance.cc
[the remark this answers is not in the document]
dANW34V3R: I think the redundant check helps with readability. You would hope this is optimised out by the compiler so no performance benefits of removing

src/lib/CoreInstance.cc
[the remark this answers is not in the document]
jj16791: No. That would create incorrect behaviour with more than one executable argument. There can be many arguments to an executable hence argc isn't fixed (`{}` denoted a set in the previous comment)

src/lib/CoreInstance.cc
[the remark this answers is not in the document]
jj16791: Siding with Rahat on this one. I argue that the accompanying error comment within the body of the if clause provides enough context for the redundant check to be removed.

src/lib/CoreInstance.cc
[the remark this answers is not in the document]
dANW34V3R: Ah ok, misunderstood the vector constructor as well. Think this has stemmed from this [comment ](https://github.com/UoB-HPC/SimEng/pull/247#discussion_r972170389) and got confused about what was actually happening e.g. this is generating a vector of arguments to pass to the executable, not just getting the executable path string. Some better naming/comment may have helped

src/lib/CoreInstance.cc
jj16791: This issue with this is when calling from outside of main, you'll have to "Spoof" the values to match what would have come from main. Maybe a renaming of variables from argc/argv to something more generic (e.g. numCLIArgs, CLIArgs)?

rahahahat: So instead why can we simplify the `CoreInstance` constructor to be like:
`CoreInstance(std::string configPath, std::string executablePath, std::vector<std::string> cliArgs)`and then in a `main.cc` context that will be handled externally like it was done in the older version of  `main.cc`.
`if configPath == ""` then load default config (inside constructor) and the same for `executablePath`
This promotes readability and is more convenient for "main.cc" and "non-main.cc" context.
Wdyt?

jj16791: Okay, I'm on board with this. Will make the change.

src/lib/CoreInstance.cc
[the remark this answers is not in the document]
jj16791: No changes are deemed necessary here. Resolving for you to move PR forward.

FinnWilkinson (approved):
More recent updates all look good

rahahahat:
Looks good to me!

rahahahat (approved):

dANW34V3R (approved):
New changes look good. I might change the final comment at a later date

src/include/simeng/CoreInstance.hh
```diff
@@ -0,0 +1,159 @@
...
+#include "simeng/arch/aarch64/MicroDecoder.hh"
+#include "simeng/kernel/Linux.hh"
+#include "simeng/models/emulation/Core.hh"
+#include "simeng/models/inorder/Core.hh"
+#include "simeng/models/outoforder/Core.hh"
+#include "simeng/pipeline/A64FXPortAllocator.hh"
+#include "simeng/pipeline/BalancedPortAllocator.hh"
+#include "yaml-cpp/yaml.h"
+
+// Instruction set used when no executable is provided; counts down from
```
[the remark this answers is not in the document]
jj16791: Good idea.

PR #249, opened 2022-09-21
Output prefix

jj16791:
This PR adds prefixes to all SimEng outputs to help distinguish between simulated workload outputs and the outputs from the framework. Additionally, furthering outputs at the start of simulation have been added to display the workload and config file used during simulation.


FinnWilkinson (approved):
All looks good, but there seems to be duplicate code from other PRs (changes to khz calculation for one example)

rahahahat (approved):
All looks good to me.

PR #250, opened 2022-10-07
Custom Testing Framework for SST -  SE-121

rahahahat:
This PR Implements a custom SST testing framework which exposes two main testing macros (`TEST_GROUP` and `TEST_CASE`) along with multiple evaluation macros. The framework runs sst with a config file specified by the `TEST_GROUP` and captures the stdout of the child process, injecting it to individual test cases and allowing the user to define multiple checks on that captured output through the use of evaluation macros. The testing framework also exposes a parser, which parses output of the sst simulation and turns the output into structs, which can then be used by the tests. Currently, there are two formats:
1.  `[SimEng:SSTDebug] MemRead-read-<type=request|response>-<request ID>-cycle-<cycle count>-data-<value>`. The entire output is parsed into: 
    ``` 
    struct ParsedMemRead {
      uint64_t reqId_;
      uint64_t startCycle_;
      uint64_t endCycle_;
      uint64_t data_;
      uint64_t numReqs_;
    };
    ```
2. `[SimEng::SSTDebug] OutputLine-<text>`. This format is parsed and `<text>` is extracted by the parser.

The testing framework adds 2 new compile definitions to the initial CMake configuration command:
- `-DSIMENG_ENABLE_SST_TESTS`: This definition enables SST Tests only if `-DSIMENG_ENABLE_TESTS` is defined.
- `-DSST_TESTS_MODEL_CONFIG_PATH`: This definition is optional, using this compile definition the user can specify a SimEng model config. However, if not provided. All tests will by default use `a64fx-sst.yaml` model configuration. 

There are 2 sst configuration files added in the testing framework: `fastL1ForBinaries.py` and `fastL1withParams_config.py`.
Arguments can be passed to these sst configuration files via test cases using a key value format i.e `key=value`.
Currently, 6 argument types can be passed by test cases and are replaced in the configuration files, ultimately to be passed to `SimEngCoreWrapper`:
- `execBin` : This is the path to an executable executed by SSTSimEng. Example: `execBin=/ab/cd/exec`.
- `args`: These are all the args passed to the executable. Example: `args=10 20 50`.
- `withSrc`: This signals if test will passed a string of instructions to be assembled by LLVM. Example: `withSrc=<True|False>`.
- `source` This contains all instructions to be assembled by LLVM. Example: `source=mov x1, #0`.
- `heap`: This contains comma separated values which will be used to initialise the heap. Example: `heap=10,20,30`.
- `clw`: This contains the cache line width in bytes: Example: `clw=64`.

For binaries for validation tests have been included: `stream_triad`, `stream_triad_sve`, `cachebw_static` , `cachebw_static_sve`


jj16791:
rerun tests

CMakeLists.txt
```diff
@@ -205,6 +206,7 @@ if(SIMENG_ENABLE_TESTS)
   enable_testing()
 
   add_subdirectory(test)
+
```
jj16791, line 209: remove newline

jj16791 (changes requested):
Good project structure and in-code comments

sst/test/include/framework/macros/eval.hh
```diff
@@ -0,0 +1,41 @@
...
+#define EXPECT_NEQ(A, B) GENERIC_EXPECT(A, !=, B, STRINGIFY(EXPECT_NEQ(A, B)))
+
+#define STR_EQ(A, B) \
+  GENERIC_STR_MATCHER(A, compareEqual(), B, STRINGIFY(STR_EQ(A, B)))
+#define STR_NOT_EQ(A, B) \
+  GENERIC_STR_MATCHER(A, compareNotEqual(), B, STRINGIFY(STR_NOT_EQ(A, B)))
+#define STR_STARTS_WITH(A, B) \
+  GENERIC_STR_MATCHER(A, startsWith(), B, STRINGIFY(STR_STARTSWITH(A, B)))
+#define STR_CONTAINS(A, B) \
+  GENERIC_STR_MATCHER(A, contains(), B, STRINGIFY(STR_CONTAINS(A, B)))
```
FinnWilkinson, line 41: Could be useful to have comments on these macro definitions to describe what they are doing

sst/test/include/framework/macros/group.hh
```diff
@@ -0,0 +1,55 @@
...
+
+#define CREATE_TEST_CASE_G(ClassName, TestName, TestCaseName, ...)            \
+  void TestName(std::string capturedStdout);                                  \
+  REGISTER_TC_G(ClassName, TestName, CONCAT(ptr_, __COUNTER__), TestCaseName, \
+                __VA_ARGS__)                                                  \
+  void TestName(std::string capturedStdout)
+
+#define TEST_CASE(ClassName, TestCaseName, ...)                              \
+  CREATE_TEST_CASE_G(ClassName, CREATE_TEST_NAME_G(ClassName), TestCaseName, \
+                     __VA_ARGS__)
```
FinnWilkinson, line 55: Could be useful to have comments on these macro definitions to describe what they are doing

sst/test/include/framework/macros/util.hh
```diff
@@ -0,0 +1,4 @@
+#pragma once
+
+#define CONCAT(A, B) A##B
+#define STRINGIFY(STR) #STR
```
FinnWilkinson, line 4: For the sake of consistency, Could be useful to have comments on these macro definitions to describe what they are doing

sst/test/include/framework/process.hh
```diff
@@ -0,0 +1,254 @@
...
+#include "framework/output.hh"
+
+/**
+ * This struct holds all information related to any
+ * errors/exceptions that could happen during execution of the binary specified
+ * by a TEST_GROUP. A process exception is thrown upon failure and handled
+ * inside the exception handler.
+ */
+struct ProcessException {
+  /** Custon error string specifying the type of error. */
```
FinnWilkinson, line 22: \* Custom

sst/test/include/framework/process.hh
```diff
@@ -0,0 +1,254 @@
...
+    args.erase(args.begin());
+    defaultCliArgs_ = args;
+  };
+  Process() {}
+
+  /**
+   * This method is used to run the SST executable with a config file and cli
+   * arguments. This method takes in additional cli arguments which replace
+   * default ones. If no additional cli arguments are given the default ones are
+   * used.
```
FinnWilkinson, line 78: Probably better to not use `cli` in comments

sst/test/include/framework/process.hh
```diff
@@ -0,0 +1,254 @@
...
+  void runExecAndCaptureStdout(
+      std::vector<std::string> newArgs = std::vector<std::string>{}) {
+    stderrCapture_ = "";
+    stdoutCapture_ = "";
+
+    // variable used to capture the status of the child process.
+    int status;
+    // int arrays defined for stdout and stderr pipes.
+    int stdout_pipes[2];
+    int stderr_pipes[2];
```
FinnWilkinson, line 89: Why do there need to be two of each pipe?

sst/test/include/framework/registry.hh
```diff
@@ -0,0 +1,102 @@
...
+
+  /** Method creates and returns the singleton instance of Registry. */
+  static Registry* getInstance() {
+    /** Static instance of the singleton Registry class. */
+    static Registry* instance_;
+    if (instance_ == NULL) {
+      instance_ = new Registry();
+    }
+    return instance_;
+  }
```
FinnWilkinson, line 32: What is the advantage of this function over a traditional constructor?

sst/test/include/framework/registry.hh
```diff
@@ -0,0 +1,102 @@
...
+   * all Factory(s).
+   */
+  static RegistryMap* getMap() {
+    /** Static instance of the map used to store all Factory(s). */
+    static RegistryMap* map;
+    if (!map) {
+      map = new RegistryMap;
+    }
+    return map;
+  }
```
FinnWilkinson, line 66: Again, I'm confused over the form of this function. Is there a reason why the `Registry` class can't use a public constructor, then have privatly defined `instance_` and `map_` variables with getters?

FinnWilkinson (changes requested):
Generally good, but some points of clarification needed

sst/test/sstconfigs/config.py
```diff
@@ -0,0 +1,2 @@
+import sst
+import sys
```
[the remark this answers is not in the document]
rahahahat: Yes, you're right.

sst/test/include/sstsimengtest.hh
```diff
@@ -0,0 +1,21 @@
...
+    for (auto itrr = fvec.begin(); itrr != fvec.end(); itrr++) { \
+      std::unique_ptr<Runner> rn = (*itrr)();                    \
+      rn->run();                                                 \
+    }                                                            \
+  }                                                              \
+  auto& stats = Stats::getInstance();                            \
+  stats->printStats();                                           \
+  if (stats->getFailureCount() > 0) {                            \
+    exit(EXIT_FAILURE);                                          \
+  }
```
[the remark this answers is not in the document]
rahahahat: Not really! I just thought it was a bit cleaner this way. Can remove it if you want.

sst/test/include/framework/expression.hh
```diff
@@ -0,0 +1,249 @@
+#pragma once
+
+#include <stdexcept>
+#include <string>
+#include <type_traits>
+
+/**
+ * SFINAE type trait which only allows the template arguments to be a
```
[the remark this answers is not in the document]
rahahahat: Substitution failure is not an error.
https://en.cppreference.com/w/cpp/language/sfinae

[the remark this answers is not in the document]
rahahahat: Shall I add a comment somewhere with a link to SFINAE concepts?

sst/test/include/sstsimengtest.hh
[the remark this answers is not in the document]
FinnWilkinson: I personally don't see the advantage, and setting off the test suite is something main can / should do. Welcome to other opinions on this though from you and the others

sst/test/include/framework/expression.hh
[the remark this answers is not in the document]
FinnWilkinson: I think it would be useful yes

sst/test/include/framework/registry.hh
rahahahat: There are 2 reasons for this:
1) A static variable map is required because the TEST_GROUP Factory(s) must be registered to the registry to create a test group during runtime. To achieve this, the registration mechanism has to be implemented in a static context, otherwise, we cannot register TEST_GROUP(s) without instantiating them manually i.e by calling the constructor. This would require us to call the constructor of all classes extending `Group` manually (which is undesirable)
2) Since execution in a static context is necessary, we have to deal with something called the static initialisation fiasco. Basically, there is no way to guarantee the initialisation order in a static context, so it could happen that when invoking a static variable via a static function it might not have been initialised (hence leading to a segfault), to ensure that we always access an initialised static variable, the getter pattern is required.

sst/test/include/framework/registry.hh
rahahahat: Referring to the response left for this [comment](https://github.com/UoB-HPC/SimEng/pull/250#discussion_r991511710). 
Doing this always ensures we always have a singleton, without having to ensure that manually by passing references and ensuring nothing is copied, this way makes it more manageable and cleaner.

sst/test/include/framework/process.hh
rahahahat: We need one pipe to redirect the STDOUT and one pipe to redirect the STDOUT. Each pipe has two ends: a read end and a write end. The read end is used to read data from the pipe and the write end is used to write data. Also, the pipe syscall requires an int array of size 2: ` int pipe(int pipefd[2]);`

CMakeLists.txt
rahahahat: Done.

sst/test/include/framework/context.hh
```diff
@@ -0,0 +1,61 @@
...
+  TestFunc tfn_;
+  /** The source of the test case. */
+  SourceInfo tsinfo_;
+
+ public:
+  /** Constructor used to a TestContext by the TEST_CASE macro. */
+  TestContext(TestFunc fn, const SourceInfo& info, std::string tname) {
+    tsinfo_ = info;
+    tname_ = tname;
+    tfn_ = std::function<void(std::string)>(fn);
```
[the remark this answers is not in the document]
rahahahat: Good spot, it isn't needed. I've removed it.

sst/test/include/framework/expression.hh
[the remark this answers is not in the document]
rahahahat: Added a definition and link to docs.

sst/test/include/framework/expression.hh
```diff
@@ -0,0 +1,249 @@
...
+ * or std::string.
+ */
+template <typename... Ts>
+using IsString = typename std::enable_if<std::conjunction<std::disjunction<
+    std::is_same<char*, typename std::decay_t<Ts>>,
+    std::is_same<const char*, typename std::decay_t<Ts>>,
+    std::is_same<std::string, typename std::decay_t<Ts>>>...>::value>::type;
+
+/**
+ * This struct represent the evaluation of expression and all details related
```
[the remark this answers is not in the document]
rahahahat: Done.

sst/test/include/framework/expression.hh
```diff
@@ -0,0 +1,249 @@
...
+   * This is a virtual method which converts a Unary expression represented by a
+   * class extending BaseExpr into ExprEval.
+   */
+  virtual auto makeUnaryExprEval() -> ExprEval {
+    return ExprEval{false, false, "", 0};
+  }
+};
+
+/**
+ * This class represent the Lhs of any expression which can evaluated to a
```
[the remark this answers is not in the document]
rahahahat: Done.

sst/test/include/framework/expression.hh
```diff
@@ -0,0 +1,249 @@
...
+  /** The left hand side of an expression. */
+  Lhs lhs_;
+
+ public:
+  explicit LhsExpr(Lhs lhs) : lhs_(lhs){};
+
+  /**
+   * This operator overload enables the execution of the 'greater than'
+   * operation on the rhs value without having to define its template type. The
+   * templated type is filtered by a SFINAE type trait which only allows
```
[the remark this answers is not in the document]
rahahahat: Done.

sst/test/include/framework/expression.hh
```diff
@@ -0,0 +1,249 @@
...
+  template <typename Ts, typename = IsString<Ts>>
+  auto operator<<(Ts const& arg) -> bool {
+    throw std::invalid_argument(
+        "String comparisons should be done with STR macros.");
+    return false;
+  }
+};
+
+/**
+ * This class represent the Lhs and rhs of any expression containing strings
```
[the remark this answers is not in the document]
rahahahat: Done.

sst/test/include/framework/macros/eval.hh
rahahahat: Please take a look.

sst/test/include/framework/macros/util.hh
rahahahat: Please take a look.

sst/test/include/framework/process.hh
rahahahat: Changed.

sst/test/include/sstsimengtest.hh
```diff
@@ -0,0 +1,5 @@
+#include "framework/macros/eval.hh"
+#include "framework/macros/group.hh"
+#include "framework/registry.hh"
+#include "framework/runner.hh"
+#include "framework/stats.hh"
```
FinnWilkinson, line 5: Adding to the previous point on this, these headers (IMO) can be added into main.cc also, with this file being removed all together

sst/test/include/framework/expression.hh
```diff
@@ -0,0 +1,255 @@
...
+   * allows the template argument to be a fundamental type.
+   */
+  template <typename Rhs, typename = IsFundamentalType<Rhs>>
+  auto operator>=(Rhs const& rhs) -> ExprEval {
+    return ExprEval{static_cast<bool>(lhs_ >= rhs), true};
+  };
+
+  /**
+   * This operator overload enables the execution of the 'less than or equal'
+   * operation on the rhs value without having to define its template type. The
```
FinnWilkinson, line 122: Could you go through the file (/ project) and capitalise LHS and RHS please? Stylictic overkill I know...

sst/test/include/framework/macros/eval.hh
FinnWilkinson: Looks good, thanks

sst/test/include/framework/process.hh
FinnWilkinson: Makes sense. Thanks for the explanation

sst/test/include/framework/registry.hh
FinnWilkinson: Makes sense, thanks for explaining

sst/test/main.cc
```diff
@@ -0,0 +1,18 @@
...
+      std::unique_ptr<Runner> rn = (*itrr)();
+      rn->run();
+    }
+  }
+  auto& stats = Stats::getInstance();
+  stats->printStats();
+  if (stats->getFailureCount() > 0) {
+    exit(EXIT_FAILURE);
+  }
+}
```
FinnWilkinson, line 18: Final `return 0;` (or equivalent) needed at end on main function

sst/test/include/sstsimengtest.hh
rahahahat: If remove this file, then everything we have define a testcase we would have to import both a `macro/group.h` and `macro/eval.h`. I think unifies the 'include statements needed. However, happy to remove it.

sst/test/include/framework/expression.hh
rahahahat: Done.

sst/test/main.cc
rahahahat: Done.

sst/test/include/sstsimengtest.hh
FinnWilkinson: I don't think it is too much overhead to add an additional header to each new test file. But again, its a very minor thing so you can leave it as is if you prefer

src/include/simeng/CoreInstance.hh
```diff
@@ -113,6 +118,15 @@ class CoreInstance {
   /** Construct the special file directory. */
   void createSpecialFileDirectory();
 
+  /** Whether or not the source has been assembled by LLVM. */
+  bool assembledSource_ = false;
+
+  /** Reference to source assembled by LLVM. */
+  char* source_ = nullptr;
+
+  /** Size of the source code assembled by LLVM*/
```
jj16791, line 127: LLVM*/ -> LLVM. */

jj16791 (changes requested):
Looking good. Will be able to comment more once in use.

sst/test/include/sstsimengtest.hh
rahahahat: added extra file framework/parser.hh which is needed by each test, I vote to keep the single header as the inclusion of parser now means, every test file now needs to include 3 files.

src/include/simeng/CoreInstance.hh
rahahahat: Done.

sst/test/src/tg0_llvm_assemble.cc
```diff
@@ -0,0 +1,13 @@
...
+
+TEST_GROUP(TG0, "SSTSimEng_correctly_assembles_instructions_using_LLVM",
+           "fastL1WithParams_config.py", "src",
+           R"(
+    mov x1, #1
+    mov x1, #1
+    mov x1, #1
+  )")
+TEST_CASE(TG0, "Test_asssembly_of_simple_instructions") {
+  // If this test doesn't throw an error, means LLVM succesfully assemble
```
rahahahat, line 11: Typo.

sst/include/Assemble.hh
```diff
@@ -0,0 +1,36 @@
...
+  uint8_t* code_ = nullptr;
+
+  /** The size of the assembled flat binary in bytes. */
+  size_t codeSize_ = 0;
+
+  /** Assemble test source to a flat binary for the given triple. */
+  void assemble(const char* source, const char* triple);
+
+ public:
+  /** Constructor for Assemble class which takes in source code. */
```
rahahahat, line 24: Assembler*

src/include/simeng/CoreInstance.hh
```diff
@@ -53,6 +53,11 @@ class CoreInstance {
   CoreInstance(std::string configPath, std::string executablePath,
                std::vector<std::string> executableArgs);
 
+  /** CoreInstance with an source assembled by LLVM and a model configuration.
```
rahahahat, line 56: With source code assembled*

sst/test/src/tg5_benchmarks.cc
```diff
@@ -0,0 +1,39 @@
...
+
+TEST_CASE(TG5, "Running_stream_triad_sve", appendBinDirPath("stream_t_sve")) {
+  Parser p = Parser(capturedStdout);
+  std::vector<std::string> outputLines = p.getOutputLines();
+  STR_CONTAINS(outputLines[0], "Solution Validates");
+}
+
+// Very basic test cases for cachebw. Since cachebw requires arg size to
+// accurately hit cache levels, More benchmark tests should be added as and when
+// more cache models are added. This test case will to test whether cachebw
```
rahahahat, line 19: Typo: These test cases will only*

sst/test/include/framework/parser.hh
```diff
@@ -0,0 +1,119 @@
...
+          out.push_back(itr->second);
+        } else {
+          ParsedMemRead* p = new ParsedMemRead();
+          p->reqId_ = id;
+          p->startCycle_ = std::stoull(splitStr[5]);
+          p->numReqs_ = std::stoull(splitStr[7]);
+          pmap.insert(std::pair<uint64_t, ParsedMemRead*>(id, p));
+        }
+      }
+      if (splitStr[0] == "[SimEng:SSTDebug:OutputLine]") {
```
FinnWilkinson, line 69: Unsure what was decided in the end regarding the fidelity of the output prefix. I think it was [SimEng:LowestLevelComponent] rather than [SimEng:component:foo:bar] etc.

@jj16791 input on this would be appreciated

sst/test/src/tg0_llvm_assemble.cc
```diff
@@ -0,0 +1,13 @@
+#include "sstsimengtest.hh"
+
+TEST_GROUP(TG0, "SSTSimEng_correctly_assembles_instructions_using_LLVM",
+           "fastL1WithParams_config.py", "src",
+           R"(
+    mov x1, #1
+    mov x1, #1
+    mov x1, #1
+  )")
```
FinnWilkinson, line 9: Fro this test, it may be a good idea to have a few different instructions i.e. mov, arithmetic, fp, simd, sve, ld, str just so we can be sure that all the right extensions are being assembled correctly

sst/test/src/tg1_load_store.cc
```diff
@@ -0,0 +1,192 @@
...
+           "fastL1WithParams_config.py", "src", R"( mov x1, #1 )");
+
+TEST_CASE(TG1, "load_of_different_size_from_memory_64bits", "src", R"(
+    # Get heap address
+    mov x0, 0
+    mov x8, 214
+    svc #0
+    ldr x1, [x0]
+    )",
+          "348709988") {
```
FinnWilkinson, line 13: For all of these tests, where are the heap values being initialised?

sst/test/src/tg1_load_store.cc
```diff
@@ -0,0 +1,192 @@
...
+  Parser p = Parser(capturedStdout);
+  std::vector<ParsedMemRead*> reads = p.getParsedMemReads();
+  // skip first two parsed requests as those will be caused by heap address
+  // retrieval into x0.
+  EXPECT_EQ(reads[2]->data_, (uint64_t)20);
+  EXPECT_EQ(reads[3]->data_, (uint64_t)40);
+  EXPECT_EQ(reads[4]->data_, (uint64_t)50);
+}
+
+TEST_CASE(TG1, "store_than_load_64bit", "src", R"(
```
FinnWilkinson, line 142: should be "store_then_load_64bit"

sst/test/src/tg1_load_store.cc
```diff
@@ -0,0 +1,192 @@
...
+    str x1, [x0]
+    ldr x2, [x0]
+    )",
+          "") {
+  Parser p = Parser(capturedStdout);
+  std::vector<ParsedMemRead*> reads = p.getParsedMemReads();
+  EXPECT_EQ(reads[0]->data_, (uint64_t)2048);
+}
+
+TEST_CASE(TG1, "store_than_load_32bit", "src", R"(
```
FinnWilkinson, line 155: should be "store_then_load_32bit"

sst/test/src/tg1_load_store.cc
```diff
@@ -0,0 +1,192 @@
...
+    str w1, [x0]
+    ldr w2, [x0]
+    )",
+          "") {
+  Parser p = Parser(capturedStdout);
+  std::vector<ParsedMemRead*> reads = p.getParsedMemReads();
+  EXPECT_EQ(reads[0]->data_, (uint64_t)256);
+}
+
+TEST_CASE(TG1, "store_than_load_16bit", "src", R"(
```
FinnWilkinson, line 168: should be "store_then_load_16bit"

sst/test/src/tg1_load_store.cc
```diff
@@ -0,0 +1,192 @@
...
+    strh w1, [x0]
+    ldrh w2, [x0]
+    )",
+          "") {
+  Parser p = Parser(capturedStdout);
+  std::vector<ParsedMemRead*> reads = p.getParsedMemReads();
+  EXPECT_EQ(reads[0]->data_, (uint64_t)64);
+}
+
+TEST_CASE(TG1, "store_than_load_8bit", "src", R"(
```
FinnWilkinson, line 181: should be "store_then_load_8bit"

sst/test/src/tg2_cache_access.cc
```diff
@@ -0,0 +1,301 @@
...
+  // retrieval into x0.
+  uint64_t firstLoadCycleLatency = reads[1]->endCycle_ - reads[1]->startCycle_;
+  uint64_t firstLoadData = reads[1]->data_;
+  uint64_t lastLoadCycleLatency = reads[2]->endCycle_ - reads[2]->startCycle_;
+  uint64_t lastLoadData = reads[2]->data_;
+
+  EXPECT_LT(lastLoadCycleLatency, firstLoadCycleLatency);
+  EXPECT_EQ(1024, firstLoadData);
+  EXPECT_EQ(1024, lastLoadData);
+}
```
FinnWilkinson, line 138: Do these two tests not do essentially the same thing? i.e. couldn't the first test's functionality be verified by the second test?

sst/test/src/tg2_cache_access.cc
```diff
@@ -0,0 +1,301 @@
...
+    str x1, [x2, #16]
+    str x1, [x2, #24]
+    str x1, [x2, #32]
+
+    # Multiple instructions to wait for all stores to execute.
+    # Can't be replaced by loop to ensure that caching was
+    # established by store. This sequence is larger to ensure all
+    # store instructions have been executed before we execute load
+    # instructions.
+    mov x4, #0
```
FinnWilkinson, line 168: I wonder if you could use a loop but force an exception to be invoked (i.e. syscall) just before the branch call to ensure a pipeline flush? If it isn't dependant on anything then it should be executed near the start of every loop

sst/test/src/tg4_request_misaligned.cc
```diff
@@ -0,0 +1,424 @@
+#include "sstsimengtest.hh"
+
+TEST_GROUP(TG4, "SSTSimEng_handles_misaligned_memory_requests",
```
FinnWilkinson, line 3: Why isn't 8-bit cache line tested in this group, where as in request_split tests a cache line width of 8-bits is used?

FinnWilkinson (changes requested):

sst/CMakeLists.txt
```diff
@@ -31,6 +31,18 @@ target_link_options(sstsimeng PUBLIC -fno-common LINKER:-undefined,dynamic_looku
 # Linking yaml and libsimeng.so libraries to sstsimeng.so
 target_link_libraries(sstsimeng PUBLIC libsimeng)
 
+if (SIMENG_ENABLE_TESTS)
+  if(SIMENG_ENABLE_SST_TESTS)
+    target_compile_definitions(sstsimeng PUBLIC SIMENG_ENABLE_TESTS SIMENG_ENABLE_SST_TESTS -DSIMENG_LLVM_VERSION=12)
```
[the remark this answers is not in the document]
rahahahat: Good spot, I've made the change.

sst/SimEngMemInterface.cc
```diff
@@ -216,6 +230,22 @@ void SimEngMemInterface::aggregatedReadResponses(
...
+// Format:
+// [SimEng:SSTDebug:MemRead]-read-<type=request|response>-<request ID>
+// -cycle-<cycle count>-data-<value>
+#ifdef SIMENG_ENABLE_SST_TESTS
+  if (debug_) {
+    std::cout << "[SimEng:SSTDebug:MemRead]"
+              << "-read-response-" << aggrReq->id_ << "-cycle-" << tickCounter_
+              << "-data-" << resp << std::endl;
+  }
+#endif
```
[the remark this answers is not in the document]
rahahahat: debug output is meaningful only when the testing mode is enabled. However, debug output is not always needed when testing mode is enabled, i.e for benchmarks. This debug output is solely for the purpose of getting information from the interface so that the testing framework can use it. 
Any other debug information related to SST components can be directly requested from SST components via config params.

sst/test/src/tg4_request_misaligned.cc
rahahahat: The reason there is no 8-bit group is that addresses are always byte aligned, so that means each incremental address will be a separate cache line. i.e `ld x5 #2` and `ld x5 #3` will be on cache line 2 (Start Addr: 2*8 = 16) and 3 (Start Addr: 3*8 = 24) respectively. This means that all load and store instructions will start at the start of a cache line. Cases when requests are larger than 8-bit cache line are already handled in request split.

sst/test/src/tg2_cache_access.cc
rahahahat: In theory, yes but there is a small difference: The first test checks if load requests are cached i.e if something was loaded from the initialised process image, without having been stored first. The second test checks if the value being stored is cached simultaneously (write-through)
The load address in the second test is the same as the store address, whereas in the first test the address for all load requests is the heap address.

sst/test/src/tg1_load_store.cc
rahahahat: The values get initialised in the final process image starting from the heap address before being sent to SST, so they won't be in the cache.

sst/SimEngMemInterface.cc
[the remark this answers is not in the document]
FinnWilkinson: OK I see. But my original question remains - if `debug_` can only be set when testing mode is enabled, why the need for the `#ifdef`? Or can `debug_` be used as a flag outside of the testing context?

sst/test/src/tg0_llvm_assemble.cc
FinnWilkinson: Although it is copied from regressionTest.hh, I still think it is worth while to ensure that this version of the implementation is done correctly. Plus, it will ensure that passing an instruction string to the CoreWrapper works as expected for a range of instruction types.

sst/test/src/tg1_load_store.cc
FinnWilkinson: Sorry, I meant outside of the test. How / where is `348709988` being put into the heap? Where is this process image made?

sst/test/src/tg2_cache_access.cc
FinnWilkinson: I see, thanks for explaining

sst/test/src/tg4_request_misaligned.cc
FinnWilkinson: Understood. Thanks

sst/SimEngCoreWrapper.cc
```diff
@@ -255,7 +284,22 @@ void SimEngCoreWrapper::fabricateSimEngCore() {
...
+#ifdef SIMENG_ENABLE_SST_TESTS
+  if (heapStr_ != "") {
+    std::vector<uint8_t> initialHeapData;
+    std::vector<uint64_t> heapVals = splitHeapStr();
+    uint64_t heapSize = heapVals.size() * 8;
+    initialHeapData.resize(heapSize);
+    uint64_t* heap = reinterpret_cast<uint64_t*>(initialHeapData.data());
+    for (size_t x = 0; x < heapVals.size(); x++) {
+      heap[x] = heapVals[x];
+    }
```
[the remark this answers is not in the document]
rahahahat: This mechanism is similar to what is done in all tests in `regression/aarch64/instructions`.
The need for casting arises as SST stores data in chunks of `uint8_t` bytes and initial heap data is of type `uint64_t`

**Quoting from [cppreference](https://en.cppreference.com/w/cpp/language/reinterpret_cast)**
> A value of any integral or enumeration type can be converted to a pointer type. A pointer converted to an integer of sufficient size and back to the same pointer type is guaranteed to have its original value, otherwise, the resulting pointer cannot be dereferenced safely (the round-trip conversion in the opposite direction is not guaranteed; the same pointer may have multiple integer representations) The null pointer constant [NULL](https://en.cppreference.com/w/cpp/types/NULL) or integer zero is not guaranteed to yield the null pointer value of the target type; [static_cast](https://en.cppreference.com/w/cpp/language/static_cast) or [implicit conversion](https://en.cppreference.com/w/cpp/language/implicit_cast) should be used for this purpose.

As far as the second point is concerned, heap and initialHeapData will always have the same pointer. After the cast is complete `uint64_t` values are then placed in  `std::vector<uint8_t>initialHeapData` using a casted `uint32_t*` start address.

sst/test/src/tg2_cache_access.cc
rahahahat: Even if a flush is caused before the branch, the problem lies more with the branch prediction mechanism, an incorrect prediction will lead to the execution of the last load instruction prematurely i.e before the first request completes. Since we can't flush SST that request will still propagate and the result will not be used, however, it will still be cached in SST.

sst/test/src/tg2_cache_access.cc
rahahahat: There is a small typo in the first test case, the comment for the first sequence of instructions is misleading. Will change it

sst/SimEngCoreWrapper.cc
```diff
@@ -223,15 +234,33 @@ std::vector<std::string> SimEngCoreWrapper::splitArgs(std::string strArgs) {
...
+          simengConfigPath_);
+    } else {
+      coreInstance_ = std::make_unique<simeng::CoreInstance>(
+          simengConfigPath_, executablePath_, executableArgs_);
+    }
+  } else {
+    coreInstance_ = std::make_unique<simeng::CoreInstance>(executablePath_,
+                                                           executableArgs_);
+  }
+#else
```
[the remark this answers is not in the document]
rahahahat: Will do. Thanks for pointing this out

sst/test/src/tg2_cache_access.cc
FinnWilkinson: My thinking was that the exception caused will essentially prevent the prediction from going through as the syscall will stall and flush the pipeline until it has finished executing - if the loop is constructed correctly. Then by the time the load instruction is fetched, the branch should have executed and the actual path would be known.

Something as simple as the following may work:
```
mov x0, xzr
mov x1, #100
add x0, x1
cmp x0, x1
mrs x10, CNTVCT_el0
b.lt -0x12

ldr x1, [x0]
```

Additionally, you can take advantage of the default branch behaviour - if taken is always the first guess then the above code will work regardless. If it is not-taken then you can get creative with jumping over the final load then jumping back to it based on a taken condition.

May be worth a small trial to at least see the behaviour it exhibits? Would be good if there was a way that cut down on this "padding code".

sst/test/src/tg1_load_store.cc
FinnWilkinson: I just realised I was mis-understanding what was being done yesterday.......... This is clear now, thanks for the explanation!

sst/Assemble.cc
```diff
@@ -0,0 +1,161 @@
...
+#include "llvm/Support/SourceMgr.h"
+#include "llvm/Support/TargetRegistry.h"
+#include "llvm/Support/TargetSelect.h"
+#endif
+
+using namespace SST::SSTSimEng;
+
+#define ASSERT(expr, errStr)                                             \
+  if (!(expr)) {                                                         \
+    std::cerr << "Error occured while assembling source through LLVM:\n" \
```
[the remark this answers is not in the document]
rahahahat: Please Check

sst/test/src/tg2_cache_access.cc
rahahahat: Please check

sst/test/src/tg0_llvm_assemble.cc
rahahahat: Please check

sst/test/src/tg2_cache_access.cc
```diff
@@ -86,40 +66,22 @@ TEST_CASE(TG2, "load_after_store_on_same_address_should_be_return_from_cache",
...
+    loop:
+    cmp x5, #256
+    b.eq end
+    add x5, x5, #1
+    b loop
 
     # last load
-    ldr x4, [x2]
+    end:
+    # Adding dependency to for preventing load due to branch misprediction
```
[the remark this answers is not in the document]
rahahahat: Done!

sst/test/src/tg2_cache_access.cc
```diff
@@ -13,44 +13,24 @@ TEST_CASE(TG2, "cache_access_of_load_to_same_address", "src",
...
+    # Buffer loop
+    loop:
+    cmp x5, #256
+    b.eq end
+    add x5, x5, #1
+    b loop
+
+    # Last Load
+    end:
+    # Adding dependency to for preventing any load due to branch misprediction
```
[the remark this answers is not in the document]
rahahahat: Done!

sst/test/include/framework/parser.hh
jj16791: For this, I think it would be best as just `[SimEng:SSTDebug]` for the sake of consistency

jj16791 (changes requested):
Various comments. Overall structure looks good.

sst/test/include/framework/parser.hh
rahahahat: There are various types of parsed outputs i.e `[SimEng::SSTDebug:MemRead]` and `[SimEng::SSTDebug:OutputLine]`. The first one parses information related to a read memory request into struct, the second records the entire line into string vector. Happy to replace the second `:` with a hypen i.e `[SimEng:SSTDebug-MemRead]` and [SimEng:SSTDebug-OutputLine]`

FinnWilkinson: I think the prefix should be describing the file / component producing the message, and output types such as "OutputLine" or "MemRead" should be part of the main message body. Keeps the consistency of prefixes without making them bloated, and keeps the output descriptive

sst/test/src/tg5_benchmarks.cc
```diff
@@ -0,0 +1,39 @@
...
+  STR_CONTAINS(outputLines[0], "Solution Validates");
+}
+
+TEST_CASE(TG5, "Running_stream_triad_sve", appendBinDirPath("stream_t_sve")) {
+  Parser p = Parser(capturedStdout);
+  std::vector<std::string> outputLines = p.getOutputLines();
+  STR_CONTAINS(outputLines[0], "Solution Validates");
+}
+
+// Very basic test cases for cachebw. Since cachebw requires arg size to
```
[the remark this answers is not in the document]
rahahahat: This test just verifies that SSTSimeng is able to run cachebw and cachebw_sve

sst/SimEngCoreWrapper.cc
```diff
@@ -100,12 +109,14 @@ void SimEngCoreWrapper::init(unsigned int phase) {
 bool SimEngCoreWrapper::clockTick(SST::Cycle_t current_cycle) {
   // Tick the core and memory interfaces until the program has halted
   if (!core_->hasHalted() || dataMemory_->hasPendingRequests()) {
+    // Ticking dataMemory just increases cycle count. Tick first to ensure
```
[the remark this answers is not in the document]
rahahahat: This was intentional, as It changes the format from main, in main `data memory->tick()` is after `core->tick()`. Since the core ticks before data memory in SSTSimeng, the cycle count inside SimEngMemInterface isn't accurate as a result clock cycle calculations are off by 1.

sst/SimEngCoreWrapper.cc
```diff
@@ -58,6 +64,9 @@ SimEngCoreWrapper::~SimEngCoreWrapper() {}
 void SimEngCoreWrapper::setup() {
   sstMem_->setup();
   output_.verbose(CALL_INFO, 1, 0, "Memory setup complete\n");
+  // Run Simulation
+  std::cout << "[SimEng] Starting...\n" << std::endl;
```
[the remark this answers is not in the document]
rahahahat: Done.

sst/test/include/framework/parser.hh
jj16791: I second 

> I think the prefix should be describing the file / component producing the message, and output types such as "OutputLine" or "MemRead" should be part of the main message body. Keeps the consistency of prefixes without making them bloated, and keeps the output descriptive

I second this. It should just be `[SimEng:SSTDebug]` and any additional information should be encoded within the remainder of the output. It's then the job of the parser to dead with each message differently.

sst/SimEngCoreWrapper.cc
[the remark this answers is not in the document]
jj16791: Format as in comment format, so it could just be:
```
    // Tick data memory
    dataMemory_->tick();

    // Tick the core
    core_->tick();

    // Tick instruction memory
    instructionMemory_->tick();
```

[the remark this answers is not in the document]
jj16791: All information regarding the body of `tick()` should be coupled with its declaration/definition (but none needed as it's self-explanatory)

sst/test/src/tg5_benchmarks.cc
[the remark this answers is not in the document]
jj16791: Okay sounds good. Worth raising a Jira task to ensure this isn't forgotten

sst/test/src/tg5_benchmarks.cc
```diff
@@ -0,0 +1,39 @@
...
+TEST_CASE(TG5, "Running_stream_triad", appendBinDirPath("stream_t")) {
+  Parser p = Parser(capturedStdout);
+  std::vector<std::string> outputLines = p.getOutputLines();
+  STR_CONTAINS(outputLines[0], "Solution Validates");
+}
+
+TEST_CASE(TG5, "Running_stream_triad_sve", appendBinDirPath("stream_t_sve")) {
+  Parser p = Parser(capturedStdout);
+  std::vector<std::string> outputLines = p.getOutputLines();
+  STR_CONTAINS(outputLines[0], "Solution Validates");
```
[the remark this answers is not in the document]
jj16791: Resolving as this was indirectly answered in https://github.com/UoB-HPC/SimEng/pull/250#discussion_r1005825545

sst/SimEngCoreWrapper.cc
```diff
@@ -223,15 +234,35 @@ std::vector<std::string> SimEngCoreWrapper::splitArgs(std::string strArgs) {
 
 void SimEngCoreWrapper::fabricateSimEngCore() {
   output_.verbose(CALL_INFO, 1, 0, "Setting up SimEng Core\n");
-
   // Create the instance of the core to be simulated
+#ifdef SIMENG_ENABLE_SST_TESTS
```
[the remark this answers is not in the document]
rahahahat: Good spot! I have changed this. Please check

sst/SimEngCoreWrapper.cc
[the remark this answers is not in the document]
rahahahat: I have corrected this, please check

sst/SimEngCoreWrapper.cc
```diff
@@ -223,11 +234,22 @@ std::vector<std::string> SimEngCoreWrapper::splitArgs(std::string strArgs) {
 
 void SimEngCoreWrapper::fabricateSimEngCore() {
   output_.verbose(CALL_INFO, 1, 0, "Setting up SimEng Core\n");
-
-  // Create the instance of the core to be simulated
   if (simengConfigPath_ != "") {
+#ifdef SIMENG_ENABLE_SST_TESTS
```
jj16791, line 238: Is there a case where we'd want to run SST tests with the default config path and an executable or a config path and a string of instructions?

sst/test/sstconfigs/fastL1WithParams_config.py
```diff
@@ -0,0 +1,75 @@
...
+heap = ""
+clw = "8"
+if len(sys.argv) > 3:
+    heap = sys.argv[3]
+if len(sys.argv) > 4:
+    clw = sys.argv[4]
+
+cpu = sst.Component("core", "sstsimeng.simengcore")
+cpu.addParams({
+    "simeng_config_path": os.getcwd() + "/../../../configs/sst-cores/a64fx-sst.yaml",
```
jj16791, line 18: We've spoken about this offline but thought I'd place this comment here so it's trackable. We should try and make this path more agnostic to the user to avoid having to set the path manually

jj16791 (changes requested):
A few comments

sst/SimEngCoreWrapper.cc
rahahahat: Not really no. Although it would be quite beneficial to fallback to a default config if one is provided in the config.py file. That would solve our current problem with paths in config.py. What do you think? The framework would run even without a config being provided.

sst/test/sstconfigs/fastL1WithParams_config.py
rahahahat: Addressed in the most recent change, please review.

sst/SimEngCoreWrapper.cc
rahahahat: Addressed in the most recent changes, please review.

sst/SimEngCoreWrapper.cc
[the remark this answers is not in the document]
rahahahat: Please check!

sst/SimEngCoreWrapper.cc
rahahahat: I don't think it would add any value we want to test for specific configurations in my opinion

rahahahat: All tests default to a64fx as tx2 and m1 don't support sve instructions.

sst/test/include/framework/parser.hh
rahahahat: This has been fixed. Please check.

sst/SimEngMemInterface.cc
```diff
@@ -168,6 +169,21 @@ void SimEngMemInterface::requestRead(const MemoryAccessTarget& target,
   AggregateReadRequest* aggrReq = new AggregateReadRequest(target, requestId);
   std::vector<StandardMem::Request*> requests =
       makeSSTRequests<AggregateReadRequest>(aggrReq, addrStart, addrEnd, size);
+  uint64_t id = requestId;
+#ifdef SIMENG_ENABLE_SST_TESTS
+  testReqCounter_++;
```
jj16791, line 174: Should we not just be leveraging the standard requestID source (i.e. the sequence ID of the instruction which generates this request)?

sst/SimEngMemInterface.cc
```diff
@@ -216,6 +233,24 @@ void SimEngMemInterface::aggregatedReadResponses(
...
+  for (int x = mergedData.size() - 1; x >= 0; x--) {
+    resp = (resp << 8) | mergedData[x];
+  }
+  // SST output data parsed by the testing framework.
+  // Format:
+  // [SimEng:SSTDebug] MemRead-read-<type=request|response>-<request ID>
+  // -cycle-<cycle count>-data-<value>
+  uint64_t id = aggrReq->id_;
+#ifdef SIMENG_ENABLE_SST_TESTS
+  id = aggrReq->testId;
```
jj16791, line 246: Why do we need to change the id?

sst/SimEngMemInterface.cc
jj16791: Given we've restricted the archetype usage to only OoO

sst/SimEngCoreWrapper.cc
```diff
@@ -223,16 +235,53 @@ std::vector<std::string> SimEngCoreWrapper::splitArgs(std::string strArgs) {
...
+          a64fxConfigPath, executablePath_, executableArgs_);
+    }
+#else
     coreInstance_ = std::make_unique<simeng::CoreInstance>(executablePath_,
                                                            executableArgs_);
+#endif
+  }
+  if (coreInstance_->getSimulationMode() !=
+      simeng::SimulationMode::OutOfOrder) {
+    std::cerr << "[SimEng] SimEng only supports out-of-order cores with SST "
```
jj16791, line 280: To be consistent with the docs, maybe we should refer to the OoO sim mode as an archetype not a core https://uob-hpc.github.io/SimEng/developer/models/index.html#archetypes

sst/SimEngCoreWrapper.cc
jj16791: It may be worth exploring a set of SST model configs in the future that stresses the integration rather than just using an existing model. Not required for this PR but something to consider for the future.

jj16791 (changes requested):
Several comments

sst/SimEngMemInterface.cc
rahahahat: This change was done yesterday while I was exploring inorderpipeline archtypes, wherein multiple requests didn't have unique ids hence I was forced was resort to this. I will revert this change now.

sst/SimEngMemInterface.cc
rahahahat: Answered [here](https://github.com/UoB-HPC/SimEng/pull/250#discussion_r1007926325). Will be changing it back.

sst/SimEngCoreWrapper.cc
rahahahat: Okay sounds good.

sst/SimEngCoreWrapper.cc
```diff
@@ -277,8 +297,8 @@ void SimEngCoreWrapper::fabricateSimEngCore() {
   if (maxAddrMemory_ < coreInstance_->getProcessImageSize()) {
     output_.verbose(
         CALL_INFO, 1, 0,
-        "Error: SST backend memory is less than processImage size. Please "
-        "increase the memory allocated to memHierarchy.memBackend and "
+        "Error: SST backend memory is less than processImage size. "
+        "Please increase the memory allocated to memHierarchy.memBackend and "
```
FinnWilkinson, line 301: Again, does this print to stdout / console and need a prefix?

sst/test/src/tg1_load_store.cc
```diff
@@ -1,69 +1,73 @@
...
+TEST_CASE(TG1, "load_of_different_size_from_memory_64bits", "withSrc=True",
+          R"(source=
     # Get heap address
     mov x0, 0
     mov x8, 214
     svc #0
     ldr x1, [x0]
     )",
-          "348709988") {
+          "heap=348709988") {
```
FinnWilkinson, line 15: Thanks for the added clarification! 😅

FinnWilkinson (changes requested):
This is looking very polished. Some minor final comments / questions

sst/SimEngCoreWrapper.cc
```diff
@@ -212,9 +223,10 @@ std::vector<std::string> SimEngCoreWrapper::splitArgs(std::string strArgs) {
     }
   }
   if (escapeSingle || escapeDouble) {
-    std::cerr << "Parsing failed: Invalid format - Please make sure all "
-                 "characters/strings are escaped properly."
-              << std::endl;
+    std::cerr
```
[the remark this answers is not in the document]
rahahahat: Please check

sst/SimEngMemInterface.cc
rahahahat: Please check

sst/SimEngCoreWrapper.cc
rahahahat: Done. Please check

sst/SimEngCoreWrapper.cc
```diff
@@ -15,7 +15,7 @@ using namespace SST::Interfaces;
 
 SimEngCoreWrapper::SimEngCoreWrapper(SST::ComponentId_t id, SST::Params& params)
     : SST::Component(id) {
-  output_.init("SimEngCoreWrapper[" + getName() + ":@p:@t]:", 999, 0,
+  output_.init("[SimEng] SimEngCoreWrapper[" + getName() + ":@p:@t]:", 999, 0,
```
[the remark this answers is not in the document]
rahahahat: Would `SSTSimEng:SimEngCoreWrapper` suffice?

CMakeLists.txt
```diff
@@ -238,3 +238,7 @@ if (SIMENG_ENABLE_SST)
     Please specify -DSST_INSTALL_DIR=<path> for the SST build to proceed.")  
   endif()
 endif()
+
+# Install SimEng model configs in the build directory
+set(SIMENG_CONFIG_INSTALL_DIR "${CMAKE_BINARY_DIR}/simeng-configs")
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/configs/ DESTINATION ${SIMENG_CONFIG_INSTALL_DIR})
```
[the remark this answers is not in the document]
rahahahat: these lines of code install all simeng model configs to the build directory so SSTSimEng can link the path to default a64fx-sst.yaml config using `SIMENG_BUILD_DIR` defined in `version.hh` programmatically, without the user having to do it manually.

sst/SimEngCoreWrapper.cc
```diff
@@ -248,20 +248,40 @@ void SimEngCoreWrapper::fabricateSimEngCore() {
...
   } else {
+#ifdef SIMENG_ENABLE_SST_TESTS
+    std::string a64fxConfigPath = std::string(SIMENG_BUILD_DIR) +
+                                  "/simeng-configs/sst-cores/a64fx-sst.yaml";
+    output_.verbose(
+        CALL_INFO, 1, 0,
+        "No config path provided so defaulting to a64fx-sst.yaml\n");
+    if (assembleWithSource_) {
+      output_.verbose(CALL_INFO, 1, 0,
+                      "Assembling source instructions using LLVM\n");
```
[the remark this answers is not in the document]
rahahahat: Anything outputted using `output_` will already have a prefix as defined [here.](https://github.com/UoB-HPC/SimEng/pull/250/files#diff-f0842ed79fb00ebcea85d3da0167eab8a989527ef758bb6b4f508149b956a3d5L16). This is an SST class, which you initialise with prefix. Anything outputted using its public functions will already have the prefix.

sst/SimEngCoreWrapper.cc
```diff
@@ -248,20 +248,40 @@ void SimEngCoreWrapper::fabricateSimEngCore() {
...
     coreInstance_ = std::make_unique<simeng::CoreInstance>(executablePath_,
                                                            executableArgs_);
-  }
 #endif
-
+  }
+  if (coreInstance_->getSimulationMode() !=
+      simeng::SimulationMode::OutOfOrder) {
+    std::cerr << "[SimEng] SimEng only supports out-of-order cores with SST "
+                 "currently."
```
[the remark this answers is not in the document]
rahahahat: I have changed this to "archetypes" per Jack's earlier comments

sst/SimEngCoreWrapper.cc
rahahahat: This will already have the prefix as it its outputted using the `output_` class.

sst/SimEngCoreWrapper.cc
```diff
@@ -212,9 +223,11 @@ std::vector<std::string> SimEngCoreWrapper::splitArgs(std::string strArgs) {
     }
   }
   if (escapeSingle || escapeDouble) {
-    std::cerr << "Parsing failed: Invalid format - Please make sure all "
-                 "characters/strings are escaped properly."
-              << std::endl;
+    std::cerr << R"(
```
jj16791: Whilst this is more verbose, it's still not telling the user which part of their input is failing i.e. what exactly has failed in its parsing

sst/SimEngCoreWrapper.cc
```diff
@@ -223,16 +236,54 @@ std::vector<std::string> SimEngCoreWrapper::splitArgs(std::string strArgs) {
...
+    }
+#else
     coreInstance_ = std::make_unique<simeng::CoreInstance>(executablePath_,
                                                            executableArgs_);
+#endif
+  }
+  if (coreInstance_->getSimulationMode() !=
+      simeng::SimulationMode::OutOfOrder) {
+    std::cerr
+        << "[SimEng] SimEng only supports Out-of-order archetypes with SST "
```
jj16791: "Out-of-Order"

jj16791 (changes requested):
Looking great, just some minor comments

sst/SimEngCoreWrapper.cc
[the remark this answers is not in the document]
rahahahat: Please check.

sst/SimEngCoreWrapper.cc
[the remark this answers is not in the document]
FinnWilkinson: Understood, thanks for explaining

FinnWilkinson (approved):
Great work Rahat, all looks good to me

sst/SimEngCoreWrapper.cc
rahahahat: Please check if the new format is sufficient.

sst/SimEngCoreWrapper.cc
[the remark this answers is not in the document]
rahahahat: Shouldn't this be resolved?

sst/test/sstconfigs/fastL1WithParams_config.py
rahahahat: Shouldn't this be resolved?

sst/SimEngCoreWrapper.cc
jj16791: Much better!

sst/test/sstconfigs/fastL1WithParams_config.py
jj16791: I'm running the testing framework on my local system to ensure this works. I will resolve this once I can confirm it does.

jj16791 (approved):
The implementation looks great, well done Rahat!

sst/test/src/tg0_llvm_assemble.cc
```diff
@@ -0,0 +1,68 @@
...
+  for (size_t y = pos; y < capturedStdout.length(); y++) {
+    if (capturedStdout[y] != '\n') {
+      retired += capturedStdout[y];
+    } else {
+      break;
+    }
+  }
+  // Extract retired instruction count from "retired: <count>" string and cast
+  // to uint64_t.
+  size_t len = retired.length() - 18;
```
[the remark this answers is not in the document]
rahahahat: It is the length of the prefix i.e. `[SimEng] retired:`, which is subtracted from the total length of the string (`[SimEng] retired: <count>`) so that substring representing the number of instructions retired (`<count>`)  can be extracted and converted from string to uint64_t.

sst/test/src/tg0_llvm_assemble.cc
```diff
@@ -0,0 +1,68 @@
...
+  uint64_t retiredCount = std::stoull(retired.substr(18, len));
+  std::cout << "Total instructions retired: " << retiredCount << std::endl;
+  // This should be equal to the total number of instructions in the test case.
+  EXPECT_GTE(retiredCount, (uint64_t)37);
```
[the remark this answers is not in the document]
rahahahat: Good spot, Changed to EQ

sst/include/SimEngCoreWrapper.hh
```diff
@@ -106,8 +106,23 @@ class SimEngCoreWrapper : public SST::Component {
...
+       "false"},
+      {"heap",
+       "comma separated uint64_t values used to populate the heap. This "
+       "parameter will only be used if assemble_with_source=true",
+       ""},
+      {"debug",
+       "Enabling this parameter outputs statistics that can be parsed by the "
+       "testing "
+       "framework",
+       "false"})
```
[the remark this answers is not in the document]
rahahahat: Changed. Please check.

CMakeLists.txt
```diff
@@ -123,6 +123,7 @@ option(SIMENG_USE_EXTERNAL_LLVM "Use an external LLVM rather than building it as
 option(SIMENG_SANITIZE "Enable compiler sanitizers" OFF)
 option(SIMENG_OPTIMIZE "Enable Extra Compiler Optimizatoins" OFF)
 option(SIMENG_ENABLE_SST "Compile SimEng SST Wrapper" OFF)
+option(SIMENG_ENABLE_SST_TESTS "Whether to enable testing for sst" OFF)
```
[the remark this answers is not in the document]
rahahahat: Changed. Please check.

sst/test/src/tg0_llvm_assemble.cc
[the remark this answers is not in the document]
dANW34V3R: A comment explaining this would be useful

sst/SimEngMemInterface.cc
```diff
@@ -168,6 +169,15 @@ void SimEngMemInterface::requestRead(const MemoryAccessTarget& target,
   AggregateReadRequest* aggrReq = new AggregateReadRequest(target, requestId);
   std::vector<StandardMem::Request*> requests =
       makeSSTRequests<AggregateReadRequest>(aggrReq, addrStart, addrEnd, size);
+  // SST output data parsed by the testing framework.
+  // Format:
+  // [SSTSimEng:SSTDebug] MemRead-read-<type=request|response>-<request ID>
+  // -cycle-<cycle count>-split-<number of requests>
+  if (debug_) {
```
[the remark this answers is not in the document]
dANW34V3R: Based on response to [this comment](https://github.com/UoB-HPC/SimEng/pull/250#discussion_r1004321542)

sst/test/src/tg0_llvm_assemble.cc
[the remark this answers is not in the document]
rahahahat: Done. Please check.

dANW34V3R (approved):

PR #252, opened 2022-10-10
Dump stats SE-147

jj16791:
This PR introduces a new class named Statistics which supports the collection and dumping of stats at defined points in the simulation. Two new SimEng-specific instructions have been introduced, namely `0xABBACAFE` for dumping the current stats and `0xDEADBEEF` for resetting the stats.

The class holds two maps whose keys represent the stat and values the occurrences of that stat. The first map records the entire simulation and the latter a region controlled by the two new SimEng-specific instructions.


FinnWilkinson:
Whilst this idea is something that we should have in SimEng, different assembly values will need to be chosen. Running `0xABBACAFE`, `0xFECABAAB`, `0xDEADBEEF`, `EFBEADDE` (The literal and alternate way of reading the two new "instructions") through the Capstone cstool for every supported ISA, these values often disassemble as one or more instructions for a variety of ISAs. For AArch64 this isn't the case, but both RISCV32 and RISCV64 had valid disassembly for `0xFECABAAB`.

.gitignore
```diff
@@ -9,8 +9,9 @@ binaries
 install
 install-debug
 capstone.pc
-version.hh
```
FinnWilkinson, line 12: `version.hh`  should be in the `.gitignore` file as it contains user specific paths generated at build time

jj16791:
add commas to stat outputs

jj16791:
Update stat variables, in main and corewrapper, to a wider data type and add in comma separation

PR #253, opened 2022-10-10
Enable support for CoreInstance to assemble and execute string instructions via LLVM - SE-145

rahahahat:
This PR adds a new constructor to CoreInstance (`CoreInstance::CoreInstance(std::string instructions, std::string configPath)`). This constructor is only available after specifying the flag `-DSIMENG_ENABLE_TESTS=ON`. Calling this constructor with a string of instructions and a config path will assemble the string of instructions via LLVM and construct the process image with the assembled source.  SimEng is then able to execute these assembled instructions. This constructor is not exposed to `main.cc` so this functionality is not accessible through the compiled SimEng executable. 
This feature will mainly be used for testing SST. Instructions supplied via `TEST_GROUP(s)` will be assembled via this constructor and executed.

### DIFF
--- a/src/include/simeng/CoreInstance.hh
+++ b/src/include/simeng/CoreInstance.hh
@@ -22,6 +22,10 @@
 #include "simeng/pipeline/BalancedPortAllocator.hh"
 #include "yaml-cpp/yaml.h"
 
+#ifdef SIMENG_ENABLE_TESTS
+
+#endif
+
 // Program used when no executable is provided; counts down from
 // 1024*1024, with an independent `orr` at the start of each branch.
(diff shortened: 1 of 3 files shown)


FinnWilkinson:
Given that the SimEng test suite also has a very similar function, would it be worth creating this function once to suit both use cases?

FinnWilkinson:
Additionally PR #248 Updates the LLVM version used in SimEng to 14.0.5, so it may be worth implementing these changes after this PR (given #248 updates the test suite function and ensures backwards compatability)? Open to thoughts on this from all

PR #264, opened 2022-11-03
Don't use type punning for `RegressionTest::getMemoryValue`

### LINKED ISSUE #260: `RegressionTest::getMemoryValue` does not assert for alignment/aliasing, triggers SIGSEGV/SIGBUS in ASan
The following function casts a `char*` to `T*` without any check for alignment or aliasing, and doing this will almost certainly be UB unless `T` is `char` or something:

https://github.com/UoB-HPC/SimEng/blob/62c2890960c5b88adf0cbaae84bd723eded77cb5/test/regression/RegressionTest.hh#L64-L68

Running this through ASan, we get more or less SIGBUS:

```
/home/tom/SimEng/test/regression/RegressionTest.hh:72:42: runtime error: load of misaligned address 0x7fe2194a3541 for type 'long unsigned int', which requires 8 byte alignment
0x7fe2194a3541: note: pointer points here
 00 00 00  00 ef be ad de 00 00 00  00 78 56 34 12 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00
              ^ 
pc_0x55c93f736756###func_unsigned long RegressionTest::getMemoryValue<unsigned long>(unsigned long) const###file_/home/tom/SimEng/test/regression/RegressionTest.hh###line_72###obj_(regression-aarch64+0x30bb756)
pc_0x55c93fbb1031###func_TestBody###file_/home/tom/SimEng/test/regression/aarch64/instructions/sve.cc###line_6125###obj_(regression-aarch64+0x3536031)
...
```
Note that ASan isn't reporting false positives here as most *but not all* uses of `getMemoryValue` in `test/regression/aarch64/instructions/sve.cc` triggered a runtime error.

A quick (possibly the only UB-free thing to do) solution is to just `memcpy` the chars:

```cpp
template <typename T>
  T getMemoryValue(uint64_t address) const {
    EXPECT_LE(address + sizeof(T), processMemorySize_);
    T dest{};
    std::memcpy(&dest, processMemory_ + address, sizeof(T));
    return dest;
  }
```
With this change, all tests pass without further ASan asserts.
Let me know if PRs are welcome.

tom91136:
This resolves #260.

### DIFF
--- a/test/regression/RegressionTest.hh
+++ b/test/regression/RegressionTest.hh
@@ -86,7 +86,9 @@ class RegressionTest
   template <typename T>
   T getMemoryValue(uint64_t address) const {
     EXPECT_LE(address + sizeof(T), processMemorySize_);
-    return *(T*)(processMemory_ + address);
+    T dest{};
+    std::memcpy(&dest, processMemory_ + address, sizeof(T));
+    return dest;
   }
 
   /** The initial data to populate the heap with. */


FinnWilkinson (approved):

jj16791:
#rerun tests

rahahahat (approved):

jj16791 (approved):

dANW34V3R (approved):

tom91136:

FinnWilkinson (approved):

PR #265, opened 2022-11-03
Don't const qualify things we'll be moving later.

### LINKED ISSUE #262: Misleading `std::move` in `DispatchIssueUnit::forwardOperands`
In `DispatchIssueUnit::forwardOperands`:
https://github.com/UoB-HPC/SimEng/blob/62c2890960c5b88adf0cbaae84bd723eded77cb5/src/lib/pipeline/DispatchIssueUnit.cc#L173-L198

On line 191, there's a `std::move` on the `const` qualified `entry`, the entry's qualification is inherited from `dependents`.
A move of a const  is a no-op, it just so happens to work out correctly behaviour wise  because  `entry.uop` is a `std::shared_ptr<Instruction>` so we're just incrementing the referencing counter.

I assume the desired effect is that we actually move the uop to one of the ports.
In that case, removing the `const` on line 184 solves this with no regression tests fails.

Side note: I'm not too well versed in SimEng's performance bottlenecks but removing the const decreases the runtime of `regression-aarch64` by about 1 second (19s to 18s on Ryzen 5900X), although this is at -O0 so it probably doesn't mean much.

tom91136:
This resolves #262.

### DIFF
--- a/src/lib/pipeline/DispatchIssueUnit.cc
+++ b/src/lib/pipeline/DispatchIssueUnit.cc
@@ -181,7 +181,7 @@ void DispatchIssueUnit::forwardOperands(const span<Register>& registers,
     scoreboard_[reg.type][reg.tag] = true;
 
     // Supply the value to all dependent uops
-    const auto& dependents = dependencyMatrix_[reg.type][reg.tag];
+    auto& dependents = dependencyMatrix_[reg.type][reg.tag];
     for (auto& entry : dependents) {
       entry.uop->supplyOperand(entry.operandIndex, values[i]);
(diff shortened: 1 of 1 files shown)


FinnWilkinson (approved):

dANW34V3R (approved):

jj16791:
#rerun tests

rahahahat (approved):

rahahahat (approved):

jj16791 (approved):

tom91136:

jj16791 (approved):

PR #267, opened 2022-11-04
Fix memory leak in `Linux::getdents64`

### LINKED ISSUE #263: Memory leak in `Linux::getdents64`
@FinnWilkinson  The function's `malloc` is missing the corresponding `free`:

https://github.com/UoB-HPC/SimEng/blob/62c2890960c5b88adf0cbaae84bd723eded77cb5/src/lib/kernel/Linux.cc#L542

Reading through the `getdents64` implementation, the whole logic around populating this  `result.d_name` field is a bit weird.

 * Some calls to `memcpy` is not qualified.
 * Computation of `result.d_namlen` creates a copy of `next_direct->d_name` as `std::string` only to call `size()` and discard the string.
 * Memory for that field is allocated but not freed, and this happens in a while loop where `linux_dirent64 result` does not escape.

Because `dirent->d_name` is guaranteed to be `\0` terminated, we can just do this:

```cpp
...
    linux_dirent64 result;
    result.d_ino = next_direct->d_ino;
#ifdef __MACH__
    result.d_off = next_direct->d_seekoff;
#else
    result.d_off = next_direct->d_off;
#endif
    std::string d_name = next_direct->d_name;
    result.d_type = next_direct->d_type;
    result.d_namlen = d_name.size();
    result.d_name = d_name.data();  // C++17's non-const overload, mutation of the returned char* is permitted
    // Get size of struct before alignment
    // 20 = combined size of d_ino, d_off, d_reclen, d_type, and d_name's
    // null-terminator
...
```
`d_name`'s memory is now managed by RAII so all is sound.
Let me know if we want this version or just call `free` and I'll get a PR going.

PS: When we eventually switch to C++20, we probably want to rewrite the initialisation of `linux_dirent64` to use designations.

tom91136:
Resolves #263

### DIFF
--- a/src/lib/kernel/Linux.cc
+++ b/src/lib/kernel/Linux.cc
@@ -561,13 +561,10 @@ int64_t Linux::getdents64(int64_t fd, void* buf, uint64_t count) {
 #else
     result.d_off = next_direct->d_off;
 #endif
+    std::string d_name = next_direct->d_name;
     result.d_type = next_direct->d_type;
-    result.d_namlen = std::string(next_direct->d_name).size();
-    // `+1` for null-terminator in d_name
-    result.d_name = (char*)malloc(result.d_namlen + 1);
-    memcpy(result.d_name, next_direct->d_name, result.d_namlen);
-    // Ensure final char is a null-terminator
-    result.d_name[result.d_namlen] = '\0';
+    result.d_namlen = d_name.size();
+    result.d_name = d_name.data();
     // Get size of struct before alignment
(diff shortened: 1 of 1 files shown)


FinnWilkinson (approved):

jj16791:
#rerun tests

rahahahat (approved):

jj16791 (approved):

dANW34V3R (approved):

tom91136:

PR #272, opened 2022-11-08
Reassign iterator after erasing element in std::vector

mutalibmohammed:
Iterator is invalidated when erasing element in Vector, hence reassign iterator variable to the iterator returned by erase.

PR #273, opened 2022-11-16
AArch64 Instruction support & minor bug fixes

FinnWilkinson:
As well as some AArch64 instruction support, the following fixes are implemented : 

- Added example SME core (A64FX with in-core SME support)
- Fixes MIPS calculation in CoreWrapper.cc
- Fixed store instruction group allocation logic for non-SVE instructions
- Corrected SME instruction group allocation in
- Reassign LdAddr iterator after erasing element in LSQ load conflict logic (originally in PR #272)
- Moved optinisations of PR #274 into this PR :
   - Refactored the `ReorderBuffer::commitMicroOps` method by using binary search to find the list of uops with the same instruction id. This improves the complexity from O(n) to O(logn)
   - Changed the `DispatchIssueUnit::tick` method to use a dynamic array instead of initializing a vector for each method call. This eliminates ~ 14m memory allocations for stream-triad.
   - Replaced `std::vector` to `std::list` in `LoadStoreQueue::startLoad` as it provides O(1) deletion compared to O(n) (worst-case).
- Removed the explicit copy constructor from the AArch64 Instruction class
- Changed `MatrixRow-Count` config option to `Matrix-Count` 
   - Makes it easier for the user to understand how many physcial ZA registers they are defining
   - Hides the internal ZA implementation, again making it easier for users to understand and use

### DIFF
--- a/src/lib/arch/aarch64/Instruction_decode.cc
+++ b/src/lib/arch/aarch64/Instruction_decode.cc
@@ -467,7 +467,7 @@ void Instruction::decode() {
     }
 
     if (isStoreData_) {
-      // Idnetify store instruction group
+      // Identify store instruction group
       if (ARM64_REG_Z0 <= metadata.operands[0].reg &&
           metadata.operands[0].reg <= ARM64_REG_Z31) {
         isSVEData_ = true;
--- a/src/lib/arch/aarch64/Instruction_decode.cc
+++ b/src/lib/arch/aarch64/Instruction_decode.cc
@@ -206,27 +206,29 @@ void Instruction::decode() {
   for (size_t i = 0; i < metadata.operandCount; i++) {
     const auto& op = metadata.operands[i];
 
+    // Determine the data type the instruction operates on based on the
+    // register operand used
+    if (i == 0) {
+      // Belongs to the predicate group if the detsination register is a
+      // predicate
+      if (op.reg >= ARM64_REG_V0) {
+        isVectorData_ = true;
+      } else if (op.reg >= ARM64_REG_ZAB0 || op.reg == ARM64_REG_ZA) {
+        isSMEData_ = true;
+      } else if (op.reg >= ARM64_REG_Z0) {
+        isSVEData_ = true;
+      } else if (op.reg <= ARM64_REG_S31 && op.reg >= ARM64_REG_Q0) {
+        isScalarData_ = true;
+      } else if (op.reg <= ARM64_REG_P15 && op.reg >= ARM64_REG_P0) {
+        isPredicate_ = true;
+      } else if (op.reg <= ARM64_REG_H31 && op.reg >= ARM64_REG_B0) {
+        isScalarData_ = true;
+      }
+    }
+
     if (op.type == ARM64_OP_REG) {  // Register operand
       if ((op.access & cs_ac_type::CS_AC_WRITE) && op.reg != ARM64_REG_WZR &&
           op.reg != ARM64_REG_XZR) {
-        // Belongs to the predicate group if the destination register is a
-        // predicate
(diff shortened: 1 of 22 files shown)


FinnWilkinson:
#rerun tests

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -463,10 +467,21 @@ void Instruction::decode() {
     }
 
     if (isStoreData_) {
-      // Identify whether a store operation uses Z source registers
+      // Idnetify store instruction group
```
jj16791, line 470: undo misspell

jj16791 (changes requested):
One small comment about a misspelling but looks good otherwise

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -1611,6 +1611,12 @@ void Instruction::execute() {
         results[0] = static_cast<int64_t>(std::trunc(operands[0].get<float>()));
         break;
       }
+      case Opcode::AArch64_FCVTZUv1i64: {  // fcvtzu dd, dn
+        // TODO: Handle NaNs, denorms, and saturation
```
dANW34V3R, line 1615: Is this something that would be difficult "TODO" now?

dANW34V3R (commented):
One comment about the TODO. The rest looks good

src/lib/arch/aarch64/Instruction_execute.cc
FinnWilkinson: I think this has been present for some time (TODO now included in 10 instructions). Having looked into possible handling, it is probably best to do this in a separate PR in order to account for rounding modes, handling of NaNs, inf, etc.

src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -34,40 +34,25 @@ void ReorderBuffer::reserve(const std::shared_ptr<Instruction>& insn) {
...
-      }
-      if (!validForCommit) return;
-
-      // No early return thus all uops are committable
-      for (; firstOp < buffer_.size(); firstOp++) {
-        if (buffer_[firstOp]->getInstructionId() != insnId) break;
-        buffer_[firstOp]->setCommitReady();
-      }
-    }
+    std::for_each(first, last, [](std::shared_ptr<Instruction>& insn) {
```
dANW34V3R, line 52: Not really a fan of this "std::for_each", it breaks from the usual structure used everywhere else and doesn't simplify much. Can it be returned to a simple "for"

jj16791 (changes requested):
A few comments for points of discussion

src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -34,40 +34,26 @@ void ReorderBuffer::reserve(const std::shared_ptr<Instruction>& insn) {
...
-        }
-      }
-      if (!validForCommit) return;
-
-      // No early return thus all uops are committable
-      for (; firstOp < buffer_.size(); firstOp++) {
-        if (buffer_[firstOp]->getInstructionId() != insnId) break;
-        buffer_[firstOp]->setCommitReady();
-      }
+    while (first != last) {
```
dANW34V3R, line 52: Add comment explaining why. The one from before is "No early return thus all uops are committable"

jj16791 (approved):

src/lib/pipeline/ReorderBuffer.cc
dANW34V3R: Comments adjusted, now resolving

dANW34V3R (approved):

jj16791 (approved):

PR #274, opened 2022-11-25
Minor optimizations

mutalibmohammed:
This PR refactors the `ReorderBuffer::commitMicroOps` method by using binary search to find the list of uops with the same instruction id. This improves the complexity from O(n) to O(logn). 

Changes the `DispatchIssueUnit::tick` method to use a dynamic array instead of initializing a vector for each method call. This eliminates  ~ 14m memory allocations for stream-triad.

Replaced `std::vector` to `std::list` in `LoadStoreQueue::startLoad` as it provides O(1) deletion compared to O(n) (worst-case).

Turn off the creation of special files by default as it creates a significant performance drag during startup.


FinnWilkinson:
#rerun tests

FinnWilkinson:
These are great optimisations. Thanks Mutalib! Again, these changes will be moved over to PR #273 so that we can merge them into `dev` asap without the last commit (Turn off special File generation by default) included.

PR #278, opened 2022-12-12
Fix NUM_GROUPS warning

dANW34V3R:
This pull request fixes the warning produced due to the name clash of NUM_GROUPS within riscv/InstructionGroups.hh and aarch64/InstructionGroups.hh. Instead of using #define which is resolved during pre-processing, these are changed to constexpr's which conform to namespaces so don't clash and also allows for type checking.


dANW34V3R:
#rerun tests

FinnWilkinson (approved):

jj16791 (approved):

rahahahat (approved):

PR #287, opened 2023-02-22
Prepare SimOS for multi-threading

FinnWilkinson:
Some light background work which will make the adoption of multi-threading much easier

### DIFF
--- a/src/include/simeng/OS/Process.hh
+++ b/src/include/simeng/OS/Process.hh
@@ -156,16 +156,17 @@ class Process {
   cpuContext context_;
 
   /** Holds the address that the process should write its TID to.
-   * Set using the `clone` syscall given the CLONE_CHILD_SETTID flag. NULL
-   * otherise. When set, the very first thing the new thread does is to write
-   * its TID at this address. */
-  uint64_t setChildTid = NULL;
+   * Set using the `clone` syscall given the CLONE_CHILD_SETTID flag. Zero
+   * otherise.
+   * When set, the very first thing the new thread does is to write its TID at
+   * this address. */
+  uint64_t setChildTid = 0;
 
   /** Holds the address of where a thread should write 0 to on termination if it
    * shared memory with other processes.
    * Set using the `clone` syscall given the CLONE_CHILD_CLEARTID flag, or using
-   * the `set_tid_address` syscall. NULL otherwise. */
-  uint64_t clearChildTid = NULL;
+   * the `set_tid_address` syscall. Zero otherwise. */
+  uint64_t clearChildTid = 0;
 
  private:
   /** Create and populate the initial process stack and returns the stack
--- a/src/include/simeng/OS/Process.hh
+++ b/src/include/simeng/OS/Process.hh
@@ -148,17 +148,24 @@ class Process {
   /** Unique pointer to FileDescArray class.*/
   std::unique_ptr<FileDescArray> fdArray_;
 
-  // Thread state
-  // TODO: Support multiple threads per process
-  /** The clear_child_tid value. */
-  uint64_t clearChildTid = 0;
-
   /** Current status of the process. */
   procStatus status_ = procStatus::waiting;
 
   /** The CPU context associated with this process. Used to enable context
    * switching between multiple processes. */
   cpuContext context_;
+
+  /** Holds the address that the process should write its TID to.
+   * Set using the `clone` syscall given the CLONE_CHILD_SETTID flag. NULL
+   * otherise. When set, the very first thing the new thread does is to write
+   * its TID at this address. */
+  uint64_t setChildTid = NULL;
+
+  /** Holds the address of where a thread should write 0 to on termination if it
+   * shared memory with other processes.
+   * Set using the `clone` syscall given the CLONE_CHILD_CLEARTID flag, or using
+   * the `set_tid_address` syscall. NULL otherwise. */
+  uint64_t clearChildTid = NULL;
 
  private:
   /** Create and populate the initial process stack and returns the stack
--- a/src/include/simeng/OS/Process.hh
+++ b/src/include/simeng/OS/Process.hh
@@ -155,18 +155,20 @@ class Process {
    * switching between multiple processes. */
   cpuContext context_;
 
-  /** Holds the address that the process should write its TID to.
-   * Set using the `clone` syscall given the CLONE_CHILD_SETTID flag. Zero
-   * otherise.
-   * When set, the very first thing the new thread does is to write its TID at
-   * this address. */
-  uint64_t setChildTid = 0;
+  /** The memory address at which the process should write its TID to.
+   * Default value is 0.
+   * It can be set using the `clone` syscall if the CLONE_CHILD_SETTID flag is
+   * present.
+   * If updated, the very first thing the new thread does is to write
+   * its TID at this address. */
+  uint64_t setChildTid_ = 0;
 
-  /** Holds the address of where a thread should write 0 to on termination if it
-   * shared memory with other processes.
-   * Set using the `clone` syscall given the CLONE_CHILD_CLEARTID flag, or using
-   * the `set_tid_address` syscall. Zero otherwise. */
-  uint64_t clearChildTid = 0;
+  /** The memory address of where a thread should write 0 to on termination if
+   * it shares memory with other processes.
+   * Default value is 0.
+   * It can be set using the `clone` syscall if the CLONE_CHILD_CLEARTID flag is
+   * present, or by calling the `set_tid_address` syscall. */
+  uint64_t clearChildTid_ = 0;
 
  private:
   /** Create and populate the initial process stack and returns the stack
--- a/src/include/simeng/OS/SimOS.hh
+++ b/src/include/simeng/OS/SimOS.hh
@@ -63,6 +63,9 @@ class SimOS {
   /** Get a process with specified `tid`. */
   const std::shared_ptr<Process>& getProcess(uint64_t tid);
 
+  /** Get the number of active processes. */
+  const size_t getNumProcesses() { return processes_.size(); }
+
   /** Terminate the process with threadID = `tid`, and set corresponding core
    * to idle state if applicable. */
   void terminateThread(uint64_t tid);
@@ -107,6 +110,10 @@ class SimOS {
  private:
   /** Construct the special file directory. */
   void createSpecialFileDirectory() const;
+
+  /** Change the state of a process or a core.
+   * Used by the terminateThread & terminateThreadGroup functions. */
+  void terminateThreadHelper(const std::shared_ptr<Process>& proc);
 
   /** The path of user defined Executable. */
   std::string executablePath_ = DEFAULT_STR;
--- a/src/include/simeng/OS/SimOS.hh
+++ b/src/include/simeng/OS/SimOS.hh
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <optional>
 #include <string>
 #include <tuple>
 #include <unordered_map>
@@ -48,10 +49,16 @@ class SimOS {
  public:
   /** Construct a SimOS object. */
   SimOS(std::string executablePath, std::vector<std::string> executableArgs,
-        std::shared_ptr<simeng::memory::Mem> mem, bool setProcess = false);
+        std::shared_ptr<simeng::memory::Mem> mem);
 
   /** Tick SimOS. */
   void tick();
+
+  /** Create a new Process Object.
+   * A span<char> is optionally passed into the function to indicate if the
+   * Process is created via raw bytes or via a compiled binary.
+   * Returns the tid of the process that was created. */
+  uint64_t createProcess(std::optional<span<char>> instructionBytes = {});
 
   /** Get a process with specified TID. */
   const std::shared_ptr<Process>& getProcess(uint64_t TID);
@@ -94,7 +101,7 @@ class SimOS {
    * line arguments.
    * Empty command line arguments denote the usage of hardcoded instructions
    * held in the hex_ array.*/
-  void createInitialProcess();
+  // void createInitialProcess();
 
   /** Construct the special file directory. */
   void createSpecialFileDirectory() const;
@@ -130,29 +137,9 @@ class SimOS {
    * exception. */
   bool halted_ = false;
 
-  /** Update the initial process to a pre-defined one.
-   * !!NOTE: Should be used EXCLUSIVELY by the test suite !! */
-  void setInitialProcess(std::shared_ptr<Process> proc,
-                         const simeng::arch::Architecture& arch) {
-    // Set Initial state of registers
-    if (Config::get()["Core"]["ISA"].as<std::string>() == "rv64") {
-      proc->context_.regFile[arch::riscv::RegisterType::GENERAL][2] = {
-          proc->context_.sp, 8};
-    } else if (Config::get()["Core"]["ISA"].as<std::string>() == "AArch64") {
-      // Set the stack pointer register
-      proc->context_.regFile[arch::aarch64::RegisterType::GENERAL][31] = {
-          proc->context_.sp, 8};
-      // Set the system registers
-      // Temporary: state that DCZ can support clearing 64 bytes at a time,
-      // but is disabled due to bit 4 being set
-      proc->context_
-          .regFile[arch::aarch64::RegisterType::SYSTEM]
-                  [arch.getSystemRegisterTag(ARM64_SYSREG_DCZID_EL0)] = {
-          static_cast<uint64_t>(0b10100), 8};
-    }
-
-    processes_[0] = proc;
-  }
+  /** The value of the next TID value that should be assigned to a process on
+   * instantiation. */
+  uint64_t nextFreeTID_ = 0;
 
   /** Reference to the PageFrameAllocator object.  */
   PageFrameAllocator pageFrameAllocator_;
--- a/src/include/simeng/OS/SimOS.hh
+++ b/src/include/simeng/OS/SimOS.hh
@@ -60,18 +60,33 @@ class SimOS {
    * Returns the tid of the process that was created. */
   uint64_t createProcess(std::optional<span<char>> instructionBytes = {});
 
-  /** Get a process with specified TID. */
-  const std::shared_ptr<Process>& getProcess(uint64_t TID);
+  /** Get a process with specified `tid`. */
+  const std::shared_ptr<Process>& getProcess(uint64_t tid);
 
-  /** Get shared_ptr to syscallHandler instance. */
-  std::shared_ptr<SyscallHandler> getSyscallHandler() const {
-    return syscallHandler_;
-  };
+  /** Terminate the process with threadID = `tid`, and set corresponding core
+   * to idle state if applicable. */
+  void terminateThread(uint64_t tid);
+
+  /** Terminate all processes with threadGroupID = `tgid`, and set corresponding
+   * cores to idle state if applicable. */
+  void terminateThreadGroup(uint64_t tgid);
+
+  /** Method which allocates multiple page frames of size 'PAGE_SIZE' to cover
+   * an address range of 'size' and returns the starting physical address. */
+  uint64_t requestPageFrames(size_t size);
+
+  /** Method which handles process specific page table translation. */
+  uint64_t handleVAddrTranslation(uint64_t vaddr, uint64_t tid);
 
   /** This method returns a callback function that is passed to the MMU.
    * The callback function will be used by the MMU to handle TLB misses. The
    * callback invokes SimOS for virtual address translations. */
   virtual VAddrTranslator getVAddrTranslator();
+
+  /** Get shared_ptr to syscallHandler instance. */
+  std::shared_ptr<SyscallHandler> getSyscallHandler() const {
+    return syscallHandler_;
+  }
 
   /** Register a core with the OS to enable process scheduling. */
   void registerCore(std::shared_ptr<simeng::Core> core) {
@@ -80,13 +95,6 @@ class SimOS {
 
   /** Check if OS has halted. */
   bool hasHalted() const { return halted_; };
-
-  /** Method which allocates multiple page frames of size 'PAGE_SIZE' to cover
-   * an address range of 'size' and returns the starting physical address. */
-  uint64_t requestPageFrames(size_t size);
-
-  /** Method which handles process specific page table translation. */
-  uint64_t handleVAddrTranslation(uint64_t vaddr, uint64_t tid);
 
   /** Unique pointer to host backed file mmap. */
   std::unique_ptr<HostBackedFileMMaps> const hfmmap_ =
@@ -97,12 +105,6 @@ class SimOS {
   friend class ::RegressionTest;
 
  private:
-  /** Create the initial SimOS Process running above this kernel from command
-   * line arguments.
-   * Empty command line arguments denote the usage of hardcoded instructions
-   * held in the hex_ array.*/
-  // void createInitialProcess();
-
   /** Construct the special file directory. */
   void createSpecialFileDirectory() const;
 
--- a/src/include/simeng/OS/SimOS.hh
+++ b/src/include/simeng/OS/SimOS.hh
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <optional>
 #include <string>
 #include <tuple>
 #include <unordered_map>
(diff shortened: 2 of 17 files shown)


src/lib/OS/SimOS.cc
```diff
@@ -90,18 +93,21 @@ void SimOS::tick() {
...
-            // Change status from Executing to Waiting
-            currProc->status_ = procStatus::waiting;
-            waitingProcs_.push(currProc);
+            auto currProc = processes_.find(currContext.TID);
+            // If proccess can't be found then it has been terminated so no need
+            // to update context.
+            if (currProc != processes_.end()) {
+              assert((currProc->second->status_ == procStatus::executing) &&
+                     "[SimEng:SimOS] Process updated when not in executing "
+                     "state.");
```
FinnWilkinson, line 102: This assert is wrong : needs to be changed to `status_ != executing`

src/lib/OS/SimOS.cc
```diff
@@ -169,72 +164,124 @@ void SimOS::createInitialProcess() {
...
+      std::cerr << "[SimEng:SimOS] Could not create process based on "
+                   "supplied instruction span"
+                << std::endl;
+      exit(1);
+    }
+  } else {
+    // Construct Process from specified binary file
+    assert(executablePath_ == DEFAULT_STR &&
+           "[SimEng:SimOS] Tried to construct a Process without specifying a "
+           "pre-compiled binary or raw assembly byte stream.");
```
FinnWilkinson, line 188: Assert is wrong, needs to be changed to `path_ != DEFAULT_STR`

src/include/simeng/OS/Process.hh
```diff
@@ -148,18 +148,26 @@ class Process {
...
-  uint64_t clearChildTid = 0;
-
   /** Current status of the process. */
   procStatus status_ = procStatus::waiting;
 
   /** The CPU context associated with this process. Used to enable context
    * switching between multiple processes. */
   cpuContext context_;
 
+  /** Holds the address that the process should write its TID to.
```
rahahahat, line 158: Can it please be reworded to say: The memory address at which the process writes its TID. The default value is 0 and is only updated when the process is created using the clone syscall with the CLONE_CHILD_SETTID flag; upon creation, the first thing the thread should do is store/write its TID at this address.

rahahahat: Given its use case can the variable not be named `childTid_`? Default it either to UINT64_MAX or 0. If it is set to a value other than default then write it to memory. 
Upon reading the syscall spec I think its more similar to actual impl

rahahahat: Process and thread are used interchangeably either this should be clarified or the difference between them should be explicitly defined.

src/include/simeng/OS/Process.hh
```diff
@@ -148,18 +148,26 @@ class Process {
...
+   * Set using the `clone` syscall given the CLONE_CHILD_SETTID flag. Zero
+   * otherise.
+   * When set, the very first thing the new thread does is to write its TID at
+   * this address. */
+  uint64_t setChildTid = 0;
+
+  /** Holds the address of where a thread should write 0 to on termination if it
```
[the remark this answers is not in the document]
rahahahat: Please add `_` to the variable name

src/lib/OS/SimOS.cc
```diff
@@ -133,22 +143,7 @@ void SimOS::tick() {
...
-  }
-  return proc->second;
-}
-
-void SimOS::createInitialProcess() {
-  // TODO : When supporting multiple processes, need to keep track of next
-  // available TGID and TIDs, and pass these when constructing a Process
-  // object
-
+uint64_t SimOS::createProcess(std::optional<span<char>> instructionBytes) {
```
rahahahat, line 146: This can be simplified by using a span created by the default constructor as the default value in the function i.e 
`uint64_t SimOS:createProcess(span<char> instrBytes = span<char>())`
then 
`if (instructionBytes.has_value()) {}` can be replaced with `if(!instructionBytes.empty()) {}`

This will also replace calls make to `createProcess({})` with `createProcess()`. IMO it is much cleaner and doesn't require `std::optional`

src/lib/OS/SimOS.cc
```diff
@@ -169,72 +164,124 @@ void SimOS::createInitialProcess() {
...
         static_cast<uint64_t>(0b10100), 8};
   }
 
-  // In a simulation's initial state, all cores will be idle. Only
-  // 'scheduled' processes may be sent to a core for execution therefore
-  // we must update the initial processes status and push it to the
-  // scheduledProcs queue
-  processes_[0]->status_ = procStatus::scheduled;
-  scheduledProcs_.push(processes_[0]);
+  // If this is the initial process (tid = 0) then add to scheduledProcs_ as
```
rahahahat, line 226: very minor: please add scheduledProcs_ queue

src/lib/OS/SimOS.cc
```diff
@@ -169,72 +164,124 @@ void SimOS::createInitialProcess() {
...
+    // If TID doesn't exist then hard exit
+    std::cerr << "[SimEng:SimOS] ERROR : Process with TID `" << tid
+              << "` does not exist.\n";
+    exit(1);
+  }
+  return proc->second;
+}
+
+void SimOS::terminateThread(uint64_t tid) {
+  auto proc = processes_.find(tid);
```
rahahahat, line 253: Nitpick: But could do:
```
auto itr = processes_.find(tid);
...
auto process = itr->second;
 ```

File: src/lib/OS/SimOS.cc [hunk 8]
```diff
@@ -169,72 +164,124 @@ void SimOS::createInitialProcess() {
...
+    case procStatus::executing:
+      // Set core's status to idle
+      for (auto core : cores_) {
+        if (core->getCurrentTID() == tid) {
+          core->setStatus(CoreStatus::idle);
+          break;
+        }
+      }
+      break;
+    case procStatus::scheduled:
```
rahahahat, line 268: Why can't the removal from scheduledProcs_ or waitingProcs_ take place here, given that we are terminating the process, wouldn't it be easier to just do it here?

src/lib/OS/SimOS.cc
```diff
@@ -169,72 +164,124 @@ void SimOS::createInitialProcess() {
...
+      // Set status to complete so it can be removed from scheduledProcs_
+    case procStatus::waiting:
+      // Set status to complete so it can be removed fromwaitingProcs_
```
rahahahat, line 271: "from<space>waitingProcs_"

src/lib/OS/SimOS.cc
rahahahat: you can use std::deque if std::queue doesn't allow the functionality

src/tools/simeng/main.cc
```diff
@@ -81,6 +81,15 @@ int main(int argc, char** argv) {
   simeng::OS::SimOS OS =
       simeng::OS::SimOS(executablePath, executableArgs, memory);
 
+  // Create the initial Process
+  if (executablePath == DEFAULT_STR) {
+    // Use default program
+    OS.createProcess({simeng::span<char>(
+        reinterpret_cast<char*>(simeng::OS::hex_), sizeof(simeng::OS::hex_))});
+  } else {
+    OS.createProcess({});
```
rahahahat, line 90: put a small comment stating would an empty createProcess will do

test/regression/RegressionTest.cc
```diff
@@ -38,21 +38,25 @@ void RegressionTest::run(const char* source, const char* triple,
   memory_ = std::make_shared<simeng::memory::SimpleMem>(memorySize);
 
   // Initialise a SimOS object.
-  simeng::OS::SimOS OS = simeng::OS::SimOS(DEFAULT_STR, {}, memory_, true);
+  simeng::OS::SimOS OS = simeng::OS::SimOS(DEFAULT_STR, {}, memory_);
 
   // Create a Process from the assembled code block.
   // Memory allocation for process images also takes place
   // during Process creation. The Elf binary is parsed
```
rahahahat, line 45: I know this change isn't from your PR but the comment is wrong.  Please remove the sentence starting from "The Elf binary is parsed..."

src/include/simeng/OS/Process.hh
```diff
@@ -148,18 +148,26 @@ class Process {
...
   /** Current status of the process. */
   procStatus status_ = procStatus::waiting;
 
   /** The CPU context associated with this process. Used to enable context
    * switching between multiple processes. */
   cpuContext context_;
 
+  /** Holds the address that the process should write its TID to.
+   * Set using the `clone` syscall given the CLONE_CHILD_SETTID flag. Zero
+   * otherise.
```
jj16791, line 160: *otherwise

src/include/simeng/OS/Process.hh
```diff
@@ -148,18 +148,26 @@ class Process {
...
+   * When set, the very first thing the new thread does is to write its TID at
+   * this address. */
+  uint64_t setChildTid = 0;
+
+  /** Holds the address of where a thread should write 0 to on termination if it
+   * shared memory with other processes.
```
jj16791, line 166: *shares memory

jj16791: *with another

src/include/simeng/OS/SimOS.hh
```diff
@@ -48,24 +49,45 @@ class SimOS {
...
+  uint64_t createProcess(std::optional<span<char>> instructionBytes = {});
 
-  /** Get shared_ptr to syscallHandler instance. */
-  std::shared_ptr<SyscallHandler> getSyscallHandler() const {
-    return syscallHandler_;
-  };
+  /** Get a process with specified `tid`. */
+  const std::shared_ptr<Process>& getProcess(uint64_t tid);
+
+  /** Terminate the process with threadID = `tid`, and set corresponding core
```
jj16791, line 66: *set the corresponding

jj16791: *to an idle

src/include/simeng/OS/SimOS.hh
```diff
@@ -48,24 +49,45 @@ class SimOS {
...
+   * to idle state if applicable. */
+  void terminateThread(uint64_t tid);
+
+  /** Terminate all processes with threadGroupID = `tgid`, and set corresponding
+   * cores to idle state if applicable. */
```
jj16791, line 71: *set the corresponding cores to an idle state

src/include/simeng/OS/SimOS.hh
```diff
@@ -1,5 +1,6 @@
 #pragma once
```
rahahahat, line 1: While I really like what you've done by refactoring createProcess,  I think we can still remove the need to call `createProcess` after `SimOS` has been initialised.

 We can solve this problem by using delegating constructors:
**Note:** Look at the order of the print statements.
```
class A {
    public:
    A(int i): A() { // delegate rest of the initialisation to A();
        b = i;
        std::cout << "Public: " << i << std::endl;
    };
    private:
        int a;
        int b;
        A() {
                a = 10;
                std::cout << "Private: " << a << std::endl;
            };
};
int main(void) {
 A a(2);
 return 0;
}
==================== Output ====================
Private: 10
Public: 2
```
Following the same logic we can have a private constructor of SimOS executing the common logic:
```
SimOS(std::shared_ptr<simeng::memory::Mem> mem) {
  memory_ = mem;
  syscallHandler_ = std::make_shared<SyscallHandler>(this);
  pageFrameAllocator_ = PageFrameAllocator(mem->getMemorySize());
  if (Config::get()["CPU-Info"]["Generate-Special-Dir"].as<bool>() == true)
    createSpecialFileDirectory();
}
```
And then have two different public constructors:
#### 1st Ctor
```
SimOS(std::shared_ptr<simeng::memory::Mem> mem, simeng::span<char> instrBytes): SimOS(mem) {
  createProcess(instrBytes);
}
Usage:
SimOS(mem, instrBytes); Used in RegressionTest.cc or while using hex_
```
#### 2nd Ctor
```
SimOS(std::shared_ptr<simeng::memory::Mem> mem, std::string executablePath, std::vector<std::string> executableArgs = {}):  SimOS(mem) {
  executablePath_ = executablePath
  executableArgs_ = executableArgs;
  createProcess()
}
Usage:
SimOS(mem, DEFAULT_STR) // no args
SimOS(mem, executablePath, execArgs) // with args

```
This way we eliminate the need for calling `createProcess` after OS has been initialised.

src/lib/OS/SimOS.cc
```diff
@@ -90,18 +93,21 @@ void SimOS::tick() {
...
-            assert(
-                (currProc->status_ == procStatus::executing) &&
-                "[SimEng:SimOS] Process updated when not in executing state.");
-            // Only update values which have changed
-            currProc->context_.pc = currContext.pc;
-            currProc->context_.regFile = currContext.regFile;
-            // Change status from Executing to Waiting
-            currProc->status_ = procStatus::waiting;
-            waitingProcs_.push(currProc);
+            auto currProc = processes_.find(currContext.TID);
```
rahahahat, line 96: Nitpick: Could do `auto itr = processes_.find()` and then `auto process = itr->second`

jj16791 (changes requested):
Some minor comment changes. Most logic changes have been identified by @rahahahat

src/include/simeng/OS/Process.hh
FinnWilkinson: Renaming to `childTid_` could become confusing when we get on to support parent and child processes (i.e. for MPI). 

In the documentation it is named `set_child_tid` and so I think our implementation should reflect that in order to maintain consistency. And it is already defaulted to 0.

rahahahat: Which documentation is being refereed here? I looked at 
image
Apologies If i've looked at the wrong one

src/include/simeng/OS/Process.hh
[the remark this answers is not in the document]
FinnWilkinson: I don't think that description is as accurate, however I agree that it needs re-wording to be more clear. I will take another look at the specification and update the comment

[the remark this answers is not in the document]
rahahahat: I have been looking at[ these](https://man7.org/linux/man-pages/man2/clone.2.html) docs. Apologies if these sources are incorrect
image

src/lib/OS/SimOS.cc
FinnWilkinson: changed to `std::deque` so processes can be removed immidatley

FinnWilkinson: Having tried an implementation of deque, I think the current implementation is better. 

The removal logic required is :
```
      auto schedItr =
          std::find_if(scheduledProcs_.begin(), scheduledProcs_.end(),
                       [tid](const std::shared_ptr<Process>& proc) {
                         return tid == proc->getTID();
                       });
      scheduledProcs_.erase(schedItr);
```
Which isn't as nice in my opinion, and is most likely less performant.

Let me know your thoughts

src/lib/OS/SimOS.cc
FinnWilkinson: done

src/lib/OS/SimOS.cc
rahahahat: Resolving it as we discussed it offline.

src/include/simeng/OS/Process.hh
FinnWilkinson: Looks like some inconsistency with the docs. I have been looking at https://man7.org/linux/man-pages/man2/set_tid_address.2.html

src/include/simeng/OS/Process.hh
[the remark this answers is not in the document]
FinnWilkinson: see https://github.com/UoB-HPC/SimEng/pull/287#discussion_r1115751748

src/lib/OS/SimOS.cc
FinnWilkinson: Done

src/lib/OS/SimOS.cc
FinnWilkinson: Done

src/lib/OS/SimOS.cc
FinnWilkinson: Function re-factored

src/lib/OS/SimOS.cc
FinnWilkinson: Done

src/tools/simeng/main.cc
FinnWilkinson: Done

test/regression/RegressionTest.cc
FinnWilkinson: Done

src/lib/OS/SimOS.cc
FinnWilkinson: Done

src/include/simeng/OS/Process.hh
FinnWilkinson: Comment re-written

src/include/simeng/OS/Process.hh
FinnWilkinson: Done first change. Second I don't think makes sense as it could share memory with multiple other processes

src/include/simeng/OS/SimOS.hh
FinnWilkinson: Done

src/include/simeng/OS/SimOS.hh
FinnWilkinson: Done

src/include/simeng/OS/SimOS.hh
FinnWilkinson: Good recommendation, have changed to this model

src/lib/OS/SimOS.cc
```diff
@@ -270,5 +311,50 @@ VAddrTranslator SimOS::getVAddrTranslator() {
   return fn;
 }
 
+// The Private constructor
+SimOS::SimOS(std::shared_ptr<simeng::memory::Mem> mem)
```
rahahahat, line 315: Would it be better to put all constructor near each other?

src/lib/OS/SimOS.cc
```diff
@@ -169,72 +169,113 @@ void SimOS::createInitialProcess() {
...
     // create the process image
     std::vector<std::string> commandLine = {executablePath_};
     commandLine.insert(commandLine.end(), executableArgs_.begin(),
                        executableArgs_.end());
 
-    processes_.emplace(
-        0, std::make_shared<Process>(commandLine, this, regFileStructure, 0, 0,
-                                     sendToMem, memory_->getMemorySize()));
+    // Create new Process. As this is a new process, the TID = TGID.
+    processes_.emplace(tid, std::make_shared<Process>(
```
rahahahat, line 200: Any reason why emplace is used? does an insert not suffice?

src/lib/OS/SimOS.cc
```diff
@@ -169,72 +169,113 @@ void SimOS::createInitialProcess() {
...
       arch->getRegisterFileStructures();
 
-  if (executablePath_ != DEFAULT_STR) {
+  // Get the tid for new Process
+  uint64_t tid = nextFreeTID_;
+  nextFreeTID_++;
+
+  if (!instructionBytes.empty()) {
+    // Construct Process from `instructionBytes`
+    processes_.emplace(tid, std::make_shared<Process>(
```
rahahahat, line 178: Any reason why emplace is used? does an insert not suffice?

src/include/simeng/OS/SimOS.hh
rahahahat: resolving

src/include/simeng/OS/Process.hh
rahahahat: this is the clone syscall docs

src/include/simeng/OS/Process.hh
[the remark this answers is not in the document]
rahahahat: these are clone syscall docs

src/include/simeng/OS/Process.hh
FinnWilkinson: As discussed offline - I think until it is exactly clear the distinction / overlap between them it is better to be more verbose. If it turns out during the clone syscall that both are not needed then I will happily change it to your suggestion

src/include/simeng/OS/Process.hh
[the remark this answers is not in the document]
FinnWilkinson: I think until it is exactly clear the distinction / overlap between them it is better to be more verbose. If it turns out during the clone syscall that both are not needed then I will happily change it to your suggestion

src/lib/OS/SimOS.cc
FinnWilkinson: Emplace will do in-place insertion, whereas insert construct the Process object and copy - as from https://www.geeksforgeeks.org/emplace-vs-insert-c-stl/

src/lib/OS/SimOS.cc
FinnWilkinson: see https://github.com/UoB-HPC/SimEng/pull/287#discussion_r1115929041

src/lib/OS/SimOS.cc
FinnWilkinson: Done

src/include/simeng/OS/SimOS.hh
```diff
@@ -1,5 +1,6 @@
...
 
+#include <optional>
```
jj16791, line 3: Still need this include?

rahahahat (approved):
Great Job Finn! Looks clean.

src/lib/OS/SimOS.cc
```diff
@@ -169,72 +179,113 @@ void SimOS::createInitialProcess() {
...
     // Concatenate the command line arguments into a single vector and
     // create the process image
     std::vector<std::string> commandLine = {executablePath_};
     commandLine.insert(commandLine.end(), executableArgs_.begin(),
                        executableArgs_.end());
 
-    processes_.emplace(
-        0, std::make_shared<Process>(commandLine, this, regFileStructure, 0, 0,
-                                     sendToMem, memory_->getMemorySize()));
+    // Create new Process. As this is a new process, the TID = TGID.
```
jj16791, line 209: Is the TID=TGID also true for creating a process from an insn span?

src/lib/OS/SimOS.cc
```diff
@@ -270,5 +321,30 @@ VAddrTranslator SimOS::getVAddrTranslator() {
...
+  uint64_t tid = proc->getTID();
+  if (proc->status_ == procStatus::executing) {
+    // Set core's status to idle, stopping execution immediately
+    for (auto core : cores_) {
+      if (core->getCurrentTID() == tid) {
+        core->setStatus(CoreStatus::idle);
+        break;
+      }
+    }
+  } else {
```
jj16791, line 342: Couldn't we just set the status of the process to be complete in all cases? I'm aware that the process doesn't live in either queue if it's executing but just thinking about avoiding any future error with different schedulers. Not too attached to the idea so don't mind if we keep the current logic.

test/regression/RegressionTest.cc
```diff
@@ -38,38 +38,24 @@ void RegressionTest::run(const char* source, const char* triple,
...
+  simeng::OS::SimOS OS = simeng::OS::SimOS(
+      memory_, simeng::span<char>(reinterpret_cast<char*>(code_), codeSize_));
 
   // Create a Process from the assembled code block.
   // Memory allocation for process images also takes place
-  // during Process creation. The Elf binary is parsed
-  // and relevant sections are copied to the process image.
+  // during Process creation.
   // The process image is finalised by the createStack method
   // which creates and populates the initial process stack.
```
jj16791, line 48: I feel like this comment is misplaced now. Doesn't really pertain to the code beneath it.

jj16791 (changes requested):
Looking good, some minor comments

src/lib/OS/SimOS.cc
FinnWilkinson: Yes, for any new process the TID will equal the TGID. Only when using `clone` or `fork` is a thread is created where the TGID will be that of the caller and the TID will be globally unique (and so they will differ).

src/lib/OS/SimOS.cc
FinnWilkinson: Changed so a process in any state is updated to `completed`

src/lib/OS/SimOS.cc
FinnWilkinson: Added comment to case when you create a process from an insn span

PR #288, opened 2023-02-23
Minibude instruction support for openmp

FinnWilkinson:
Along with the ldaddal instruction being added, the user can now input the environment variables they want to use via the YAML config file, rather than in Process.cc

### DIFF
--- a/configs/DEMO_RISCV.yaml
+++ b/configs/DEMO_RISCV.yaml
@@ -144,4 +144,4 @@ CPU-Info:
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
 Environment-Variables:
-  - "OMP_NUM_THREADS=2"
+  - "OMP_NUM_THREADS=1"
--- a/configs/DEMO_RISCV.yaml
+++ b/configs/DEMO_RISCV.yaml
@@ -143,3 +143,5 @@ CPU-Info:
   # Package-Count is used to generate
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
+Environment-Variables:
+  - "OMP_NUM_THREADS=2"
--- a/configs/a64fx.yaml
+++ b/configs/a64fx.yaml
@@ -273,4 +273,4 @@ CPU-Info:
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
 Environment-Variables:
-  - "OMP_NUM_THREADS=2"
+  - "OMP_NUM_THREADS=1"
--- a/configs/a64fx.yaml
+++ b/configs/a64fx.yaml
@@ -272,3 +272,5 @@ CPU-Info:
   # Package-Count is used to generate
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
+Environment-Variables:
+  - "OMP_NUM_THREADS=2"
--- a/configs/a64fx_SME.yaml
+++ b/configs/a64fx_SME.yaml
@@ -345,4 +345,4 @@ CPU-Info:
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
 Environment-Variables:
-  - "OMP_NUM_THREADS=2"
+  - "OMP_NUM_THREADS=1"
--- a/configs/a64fx_SME.yaml
+++ b/configs/a64fx_SME.yaml
@@ -344,3 +344,5 @@ CPU-Info:
   # Package-Count is used to generate 
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
+Environment-Variables:
+  - "OMP_NUM_THREADS=2"
--- a/configs/m1_firestorm.yaml
+++ b/configs/m1_firestorm.yaml
@@ -322,4 +322,5 @@ CPU-Info:
   # Package-Count is used to generate 
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
-
+Environment-Variables:
+  - "OMP_NUM_THREADS=2"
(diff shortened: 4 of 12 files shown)


jj16791 (changes requested):
Just one comment concerning the new config option

src/lib/OS/Process.cc
```diff
@@ -274,10 +274,11 @@ uint64_t Process::createStack(uint64_t stackStart) {
     stringBytes.push_back(0);
   }
   // Environment strings
-  std::vector<std::string> envStrings = {"OMP_NUM_THREADS=1"};
-  for (std::string& env : envStrings) {
-    for (int i = 0; i < env.size(); i++) {
-      stringBytes.push_back(env.c_str()[i]);
+  for (size_t i = 0; i < Config::get()["Environment-Variables"].size(); i++) {
```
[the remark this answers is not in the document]
FinnWilkinson: I'm unsure how this would be done... Given an environment variable could be anything, or even have none present, how would the node checker allow us to ensure it cam be read as a string? Doesn't the 2nd argument have to be the value itself which we cannot know?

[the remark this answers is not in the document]
jj16791: You just check against an empty vector and it ensures existence and ability to be read as a string. Then the 2nd arg is just an human-readable key for the failed output.

jj16791 (approved):
all looks good to me

src/lib/OS/Process.cc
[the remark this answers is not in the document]
FinnWilkinson: Check added

configs/DEMO_RISCV.yaml
```diff
@@ -143,3 +143,5 @@ CPU-Info:
   # Package-Count is used to generate
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
+Environment-Variables:
+  - "OMP_NUM_THREADS=2"
```
rahahahat, line 147: What has this env variable been set to 2

configs/a64fx.yaml
```diff
@@ -272,3 +272,5 @@ CPU-Info:
   # Package-Count is used to generate
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
+Environment-Variables:
+  - "OMP_NUM_THREADS=2"
```
rahahahat, line 276: What has this env variable been set to 2

configs/a64fx_SME.yaml
```diff
@@ -344,3 +344,5 @@ CPU-Info:
   # Package-Count is used to generate 
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
+Environment-Variables:
+  - "OMP_NUM_THREADS=2"
```
rahahahat, line 348: What has this env variable been set to 2

configs/tx2.yaml
```diff
@@ -188,3 +188,5 @@ CPU-Info:
   # Package-Count is used to generate 
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
+Environment-Variables:
+  - "OMP_NUM_THREADS=2"
```
rahahahat, line 192: What has this env variable been set to 2

configs/m1_firestorm.yaml
```diff
@@ -322,4 +322,5 @@ CPU-Info:
   # Package-Count is used to generate 
   # /sys/devices/system/cpu/cpu{0..Core-Count}/topology/{physical_package_id, core_id}
   Package-Count: 1
-
+Environment-Variables:
+  - "OMP_NUM_THREADS=2"
```
rahahahat, line 326: What has this env variable been set to 2

configs/DEMO_RISCV.yaml
rahahahat: should SST configs be updated as well?

src/lib/ModelConfig.cc
```diff
@@ -621,6 +621,16 @@ void ModelConfig::validate() {
   }
   subFields.clear();
 
+  // Environment-Variables
+  root = "Environment-Variables";
+  size_t numEnvVars = configFile_[root].size();
+  for (size_t i = 0; i < numEnvVars; i++) {
+    char envVarNum[50];
```
rahahahat, line 628: why is the hard-coded value 50?

src/lib/ModelConfig.cc
```diff
@@ -621,6 +621,16 @@ void ModelConfig::validate() {
...
+    snprintf(envVarNum, 50, "Environment-Variable %zu ", i);
```
rahahahat, line 629: Why couldn't you do: 
`std::string envVarStr = "Environment-Variable " + std::to_string(i);`

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -548,6 +548,7 @@ void Instruction::execute() {
         break;
       }
       case Opcode::AArch64_CASALW: {  // casal ws, wt, [xn|sp]
+        // TODO: Load and Store must occur atomically
```
rahahahat, line 551: Should this TODO have been removed given we are not sure if it will be atomic in SST

configs/DEMO_RISCV.yaml
FinnWilkinson: changed to 1

configs/a64fx.yaml
FinnWilkinson: changed to 1

configs/a64fx_SME.yaml
FinnWilkinson: changed to 1

configs/tx2.yaml
FinnWilkinson: changed to 1

configs/m1_firestorm.yaml
FinnWilkinson: changed to 1

src/lib/ModelConfig.cc
FinnWilkinson: This char* is just used to give each environment variable a label so that inside nodeChecker we can validate that each env variable is readable as a string. The constant part of the label-string is 22 chars, so the additional space can guarentee that each label can be assigned a unique ID number (up to 28 digits long, much more than ever will be needed)

src/lib/ModelConfig.cc
FinnWilkinson: updated

src/lib/ModelConfig.cc
FinnWilkinson: Comment now irrelevant

jj16791 (approved):

rahahahat (approved):

PR #289, opened 2023-02-24
Exit and Exit_Group syscalls

FinnWilkinson:
Implemented the `exit` and `exit_group` syscalls, along with a mechanism to allow a syscall to update the core's status to idle, once the process state returned from the syscall has been applied.

Also includes a minor tweak to the scheduling logic, meaning that if `scheduledProcs_` is empty then a process in `waitingProcs_` can jump ahead.

(Will re-target to `multi-thread-support` branch once the `exceptionHandler-refactor` PR has been merged)

### DIFF
--- a/src/include/simeng/OS/SyscallHandler.hh
+++ b/src/include/simeng/OS/SyscallHandler.hh
@@ -150,6 +150,10 @@ struct SyscallResult {
    * core and it should therefore halt. */
   bool fatal = false;
 
+  /** Indicates whether the recieving core should go into an idle state after
+   * the syscall has concluded and all state changes have been processed. */
+  bool idleAfterSyscall = false;
+
   /** Id of the syscall to aid exception handler processing. */
   uint64_t syscallId = 0;
 
@@ -228,7 +232,8 @@ class SyscallHandler {
   /** Once the syscall is complete, conclude its execution by
    * constructing a SyscallResult and supplying it to the returnSyscall_
    * function. */
-  void concludeSyscall(const ProcessStateChange& change, bool fatal = false);
+  void concludeSyscall(const ProcessStateChange& change, bool fatal = false,
+                       bool idleAftersycall = false);
 
   /** Attempt to read a string of max length `maxLength` from address `address`
    * into the supplied buffer, starting from character `offset`. */
--- a/src/include/simeng/OS/SyscallHandler.hh
+++ b/src/include/simeng/OS/SyscallHandler.hh
@@ -150,7 +150,7 @@ struct SyscallResult {
    * core and it should therefore halt. */
   bool fatal = false;
 
-  /** Indicates whether the recieving core should go into an idle state after
+  /** Indicates whether the receiving core should go into an idle state after
    * the syscall has concluded and all state changes have been processed. */
   bool idleAfterSyscall = false;
 
--- a/src/include/simeng/arch/ExceptionHandler.hh
+++ b/src/include/simeng/arch/ExceptionHandler.hh
@@ -17,6 +17,10 @@ struct ExceptionResult {
    * core and it should therefore halt. */
   bool fatal;
 
+  /** Indicates whether the recieving core should go into an idle state after
+   * the syscall has concluded and all state changes have been processed. */
+  bool idleAfterSyscall = false;
+
   /** The address to resume execution from. */
   uint64_t instructionAddress;
 
--- a/src/include/simeng/arch/ExceptionHandler.hh
+++ b/src/include/simeng/arch/ExceptionHandler.hh
@@ -17,7 +17,7 @@ struct ExceptionResult {
    * core and it should therefore halt. */
   bool fatal;
 
-  /** Indicates whether the recieving core should go into an idle state after
+  /** Indicates whether the receiving core should go into an idle state after
    * the syscall has concluded and all state changes have been processed. */
   bool idleAfterSyscall = false;
 
--- a/src/lib/OS/SimOS.cc
+++ b/src/lib/OS/SimOS.cc
@@ -105,36 +105,44 @@ void SimOS::tick() {
                (scheduledProcs_.front()->status_ == procStatus::completed)) {
           scheduledProcs_.pop();
         }
-        if (!scheduledProcs_.empty()) {
-          // Get context of process that was executing on core before interrupt
-          // was signalled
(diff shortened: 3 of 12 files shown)


src/include/simeng/arch/ExceptionHandler.hh
```diff
@@ -17,6 +17,10 @@ struct ExceptionResult {
    * core and it should therefore halt. */
   bool fatal;
 
+  /** Indicates whether the recieving core should go into an idle state after
```
jj16791, line 20: *receiving

src/lib/OS/SimOS.cc
```diff
@@ -244,17 +254,8 @@ uint64_t SimOS::createProcess(span<char> instructionBytes) {
...
-  // Otherwise, add the new process to the waitingProcs_ queue.
-  if (tid == 0) {
-    processes_[tid]->status_ = procStatus::scheduled;
-    scheduledProcs_.push(processes_[tid]);
-  } else {
-    processes_[tid]->status_ = procStatus::waiting;
-    waitingProcs_.push(processes_[tid]);
-  }
+  processes_[tid]->status_ = procStatus::waiting;
+  waitingProcs_.push(processes_.find(tid)->second);
```
jj16791, line 258: why has this been changed from `waitingProcs_.push(processes_[tid]);`?

jj16791 (changes requested):
Apart from Jenkins error, only minor comments

src/include/simeng/arch/ExceptionHandler.hh
FinnWilkinson: Done

src/lib/OS/SimOS.cc
FinnWilkinson: This was re-written so by mistake. Will change back

jj16791 (approved):
Looks good

src/include/simeng/OS/SyscallHandler.hh
```diff
@@ -150,6 +150,10 @@ struct SyscallResult {
    * core and it should therefore halt. */
   bool fatal = false;
 
+  /** Indicates whether the recieving core should go into an idle state after
```
rahahahat, line 153: receiving

test/unit/OSTest.cc
```diff
@@ -38,7 +38,7 @@ TEST(OSTest, CreateSimOS) {
   EXPECT_GT(proc->context_.sp, 0);
```
rahahahat, line 38: Should a test case be added for the case when a core is put to idle by process syscall result?

src/lib/OS/SimOS.cc
```diff
@@ -105,36 +107,44 @@ void SimOS::tick() {
...
+            auto currProc = procItr->second;
+            assert((currProc->status_ == procStatus::executing) &&
+                   "[SimEng:SimOS] Process updated when not in executing "
+                   "state.");
+            // Only update values which have changed
+            currProc->context_.pc = currContext.pc;
+            currProc->context_.regFile = currContext.regFile;
+            // Change status from Executing to Waiting
+            currProc->status_ = procStatus::waiting;
+            waitingProcs_.push(currProc);
```
rahahahat, line 130: Logic looks good, but what if we have only one process, in that case, that process will be pushed on to `waitingProcs_`,  and because of the new optimisation it will be scheduled back again [here](https://github.com/UoB-HPC/SimEng/pull/289/files#diff-23a38459d3c1bce689bdc1777a9d1bcad069b18b4b8870f78193b6ac4d5cc83bR143). Could this somehow be eliminated?

src/lib/OS/SyscallHandler.cc
```diff
@@ -370,17 +370,30 @@ void SyscallHandler::handleSyscall() {
...
-                   "terminating with exit code "
-                << exitCode << std::endl;
-      return concludeSyscall({}, true);
+      uint64_t tid = currentInfo_.threadId;
+      // TODO: When `wait` is supported, return exitCode & 0xFF to parent
+      // TODO: Call all functions registered with `atexit` and `on_exit`
+      // TODO: Flush all open `stdio` streams when supported
+      // TODO: Remove files created by `tmpfile` when supported
+      OS_->terminateThread(tid);
+      std::cerr << "[SimEng:SyscallHandler] Received exit syscall on Thread "
```
jj16791, line 379: Think we should continue to print to STDOUT not to STDERR as this isn't an error

src/lib/OS/SyscallHandler.cc
```diff
@@ -370,17 +370,30 @@ void SyscallHandler::handleSyscall() {
...
-                   "terminating with exit code "
-                << exitCode << std::endl;
-      return concludeSyscall({}, true);
+      uint64_t tgid = OS_->getProcess(currentInfo_.threadId)->getTGID();
+      // TODO: When `wait` is supported, return exitCode & 0xFF to parent
+      // TODO: Call all functions registered with `atexit` and `on_exit`
+      // TODO: Flush all open `stdio` streams when supported
+      // TODO: Remove files created by `tmpfile` when supported
+      OS_->terminateThreadGroup(tgid);
+      std::cerr << "[SimEng:SyscallHandler] Received exit_group syscall on "
```
jj16791, line 392: Think we should continue to print to STDOUT not to STDERR as this isn't an error

jj16791 (changes requested):
I think the comments I've made may fix the performance pipeline too

src/lib/OS/SimOS.cc
FinnWilkinson: If we only have a single process then no context switch will occur due to the test performed in the `CoreStatus::executing` switch case in SimOS.cc

src/include/simeng/OS/SyscallHandler.hh
FinnWilkinson: Done

src/lib/OS/SyscallHandler.cc
FinnWilkinson: Done

src/lib/OS/SyscallHandler.cc
FinnWilkinson: Done

rahahahat (approved):

jj16791 (approved):

PR #290, opened 2023-02-25
Added first draft logic and test for the sysinfo syscall

jj16791:
In this PR, the `sysinfo` syscall infrastructure has been introduced. Currently, not all values of the returned `sysinfo` struct (generated by the `sysinfo` syscall) are supported so have been set to zero and left for either future developments or within the lifetime of this PR based on actual usage of the syscall. 

Rather than define the `sysinfo` struct in the `SyscallHandler` header and verbosely assign each `sysinfo` struct entry  to then be added to the statechange one by one, the struct is represented as an array of `uint64_t` values. This allowed for less duplicate code and a smaller code footprint within the `sysinfo` case body. A somewhat detailed comment has been written to explain the mapping/use of this array.

### DIFF
--- a/src/lib/OS/SyscallHandler.cc
+++ b/src/lib/OS/SyscallHandler.cc
@@ -560,6 +560,7 @@ void SyscallHandler::handleSyscall() {
       uint64_t infoPtr = currentInfo_.registerArguments[0].get<uint64_t>();
       stateChange = {ChangeType::REPLACEMENT, {currentInfo_.ret}, {0ull}};
       std::array<uint64_t, 14> sysInfoVals;
+      std::fill(sysInfoVals.begin(), sysInfoVals.end(), 0);
       std::array<uint8_t, 14> sysInfoValsSizes = {8, 8, 8, 8, 8, 8, 8,
                                                   8, 8, 8, 2, 8, 8, 4};
       // Populate those entries within the sysinfo struct which are currently
--- a/src/lib/OS/SyscallHandler.cc
+++ b/src/lib/OS/SyscallHandler.cc
@@ -539,7 +539,41 @@ void SyscallHandler::handleSyscall() {
       break;
     }
     case 179: {  // sysinfo
+      // Sysinfo populates a sysinfo struct containing system information.
+      // Currently, the majority of the sysinfo struct entries will be 0 for
+      // SimEng's emulation of the syscall, therefore, we begin with a zero'ed
+      // out std::array, `sysInfoVals`, to represent this struct. The mapping
+      // between struct and `sysInfoVals` entries is as followed:
+      //
+      // {uptime, loads[0], loads[1], loads[2], totalram,
+      // freeram, sharedram, bufferram, totalswap, freeswap,
+      // procs,totalhigh, freehigh, mem_unit}
+      //
+      // (https://man7.org/linux/man-pages/man2/sysinfo.2.html has been used to
+      // define the entries to be mapped and understand what they equate to.)
+      //
+      // Each entry is represented as an uint64_t value and later narrowed
+      // implicity within the RegisterValue constructor where necessary.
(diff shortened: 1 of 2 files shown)


FinnWilkinson (approved):

test/regression/aarch64/Syscall.cc
```diff
@@ -872,6 +872,50 @@ TEST_P(Syscall, ftruncate) {
...
+    mov x21, x0
+  )");
+
+  // Check returned 0
+  EXPECT_EQ(getGeneralRegister<uint64_t>(21), 0);
+
+  // Check outputted sysinfo struct
+  uint64_t paddr = process_->translate(process_->getHeapStart());
+  // The size of the sysinfo struct within SimEng is 102 bytes
+  char reference[102] = {0};
```
rahahahat, line 904: Can use `std::array<char, 102>`

test/regression/aarch64/Syscall.cc
```diff
@@ -872,6 +872,50 @@ TEST_P(Syscall, ftruncate) {
...
+  // - totalram = 500000 (0x07A120)
+  // - procs = 1 (0x01)
+  // Populate reference with those non-zero entries
+  reference[32] = 0x20;
+  reference[33] = 0xA1;
+  reference[34] = 0x07;
+  reference[80] = 0x01;
+  auto data = memory_->getUntimedData(paddr, 102);
+  for (int i = 0; i < 102; i++) {
+    EXPECT_EQ(data[i], reference[i]) << "at index i=" << i << '\n';
```
rahahahat, line 915: Why isn't `uptime` tested, is it because the division would make it `0`. Is that something that should/does happen?
I know you can't check its exact value but can it not be checked if it is non-zero?

src/lib/OS/SyscallHandler.cc
```diff
@@ -526,7 +526,41 @@ void SyscallHandler::handleSyscall() {
...
+      // define the entries to be mapped and understand what they equate to.)
+      //
+      // Each entry is represented as an uint64_t value and later narrowed
+      // implicity within the RegisterValue constructor where necessary.
+      // `sysInfoValsSizes` is used to denote the sizes of each entry so that it
+      // occupies the correct amount of memory after the syscall is concluded.
+
+      uint64_t infoPtr = currentInfo_.registerArguments[0].get<uint64_t>();
       stateChange = {ChangeType::REPLACEMENT, {currentInfo_.ret}, {0ull}};
+      std::array<uint64_t, 14> sysInfoVals;
```
rahahahat, line 549: Quick question: are you sure that `uint64_t` is initialised as 0? if not could we add:
`std::fill(sysInfoVals.begin(), sysInfoVals.end(), 0);`

test/regression/aarch64/Syscall.cc
rahahahat: you might also have to add `std::fill(reference.begin(), reference.end() 0);`

rahahahat (changes requested):

src/lib/OS/SyscallHandler.cc
jj16791: It may be impl defined so will add the fill

test/regression/aarch64/Syscall.cc
jj16791: Any advantage to using std::array over char[] ?

test/regression/aarch64/Syscall.cc
jj16791: It will be 0. Would take a while to sim 1 second

test/regression/aarch64/Syscall.cc
rahahahat: I mean aren't we actively trying to not use C-style declarations?

jj16791: yeh good point, changing

rahahahat (approved):

PR #294, opened 2023-02-28
In-order Early Halting Fix

dANW34V3R:
This pull request adds:

- Correct tagging of RISC-V multiply and divide instructions
- Renaming of  knownTarget_ to knownOffset_ and fixing the RISC-V assignment to this variable

This pull request fixes:

- With the specific sequence of instructions: branch, multi-cycle instruction, branch, the in-order core would sometimes incorrectly halt. 

  This would only happen in the case that each branch instruction set knownTarget_ to a value outside the valid range defined by the program size. This is due to incorrectly setting the variable as the instruction's own address had not been initialised and so was some garbage value different on each run. 
  
  This caused a halt because the first mispredicted branch causes a bubble when the fetch unit halts not knowing what to fetch next. Once executed, a multicycle instruction is fetched and flows normally through the pipeline. This is immediately followed by another branch which mispredicts in the same way. This puts the pipeline in the state: write-back unit and completion slots empty due to bubble caused by first branch, execute unit stalled while multicycle instruction executes, front end buffers stalled waiting on the execute unit and fetch unit halted due to second branch mispredict. 
  
  The condition for the in-order core to halt was: the fetch unit has halted and the heads of the buffers are empty - in the above state this appears to be the case as, when stalled, the buffers do not swap the head and tail. Therefore, even after being decoded, the second branch is stuck in the tail of the decode to execute buffer and so it appears empty. The completion slots are empty due to the bubble from the first branch and the fetch to decode buffer is empty due to the bubble created by the second branch. The fetch unit has halted as the second branch has predicted outside the valid range. As all conditions are true, the in-order pipeline halts and the simulation ends.
  
  A test has been created to force this state and fails sometimes before the fix is implemented.
  
  To fix this, the halting condition is updated so that the head or tail of the buffer is checked depending on whether it is stalled or not and the execute unit is checked to see if there is an instruction currently executing. The test now always passes.

### DIFF
--- a/src/include/simeng/BranchPredictor.hh
+++ b/src/include/simeng/BranchPredictor.hh
@@ -49,7 +49,7 @@ class BranchPredictor {
   /** Generate a branch prediction for the specified instruction address with a
    * branch type and possible known target. */
   virtual BranchPrediction predict(uint64_t address, BranchType type,
-                                   uint64_t knownTarget) = 0;
+                                   int64_t knownOffset) = 0;
 
   /** Provide branch results to update the prediction model for the specified
    * instruction address. */
--- a/src/include/simeng/BranchPredictor.hh
+++ b/src/include/simeng/BranchPredictor.hh
@@ -47,7 +47,7 @@ class BranchPredictor {
   virtual ~BranchPredictor(){};
 
   /** Generate a branch prediction for the specified instruction address with a
-   * branch type and possible known target. */
+   * branch type and possible known branch offset. */
   virtual BranchPrediction predict(uint64_t address, BranchType type,
                                    int64_t knownOffset) = 0;
 
--- a/src/include/simeng/GenericPredictor.hh
+++ b/src/include/simeng/GenericPredictor.hh
@@ -27,10 +27,10 @@ class GenericPredictor : public BranchPredictor {
   ~GenericPredictor();
 
   /** Generate a branch prediction for the supplied instruction address, a
-   * branch type, and a known branch offset if not 0. Returns a branch direction
-   * and branch target address. */
+   * branch type, and a known branch offset; defaults to 0 meaning offset is not
+   * known. Returns a branch direction and branch target address. */
   BranchPrediction predict(uint64_t address, BranchType type,
-                           int64_t knownOffset) override;
+                           int64_t knownOffset = 0) override;
 
   /** Updates appropriate predictor model objects based on the address and
    * outcome of the branch instruction. */
--- a/src/include/simeng/GenericPredictor.hh
+++ b/src/include/simeng/GenericPredictor.hh
@@ -30,7 +30,7 @@ class GenericPredictor : public BranchPredictor {
    * branch type, and a known target if not 0. Returns a branch direction and
    * branch target address. */
   BranchPrediction predict(uint64_t address, BranchType type,
-                           uint64_t knownTarget) override;
+                           int64_t knownOffset) override;
 
   /** Updates appropriate predictor model objects based on the address and
    * outcome of the branch instruction. */
--- a/src/include/simeng/GenericPredictor.hh
+++ b/src/include/simeng/GenericPredictor.hh
@@ -27,8 +27,8 @@ class GenericPredictor : public BranchPredictor {
   ~GenericPredictor();
 
   /** Generate a branch prediction for the supplied instruction address, a
-   * branch type, and a known target if not 0. Returns a branch direction and
-   * branch target address. */
+   * branch type, and a known branch offset if not 0. Returns a branch direction
+   * and branch target address. */
   BranchPrediction predict(uint64_t address, BranchType type,
                            int64_t knownOffset) override;
 
--- a/src/include/simeng/Instruction.hh
+++ b/src/include/simeng/Instruction.hh
@@ -100,7 +100,7 @@ class Instruction {
   virtual BranchType getBranchType() const = 0;
 
   /** Retrieve a branch target from the instruction's metadata if known. */
-  virtual uint64_t getKnownTarget() const = 0;
+  virtual int64_t getKnownOffset() const = 0;
 
   /** Is this a store address operation (a subcategory of store operations which
    * deal with the generation of store addresses to store data at)? */
@@ -209,7 +209,7 @@ class Instruction {
   BranchType branchType_ = BranchType::Unknown;
 
   /** If the branch target is known at the time of decode, store it. */
-  uint64_t knownOffset_ = 0;
+  int64_t knownOffset_ = 0;
 
   // Flushing
   /** This instruction's sequence ID; a higher ID represents a chronologically
--- a/src/include/simeng/Instruction.hh
+++ b/src/include/simeng/Instruction.hh
@@ -99,7 +99,7 @@ class Instruction {
   /** Retrieve branch type. */
   virtual BranchType getBranchType() const = 0;
 
-  /** Retrieve a branch target from the instruction's metadata if known. */
+  /** Retrieve a branch offset from the instruction's metadata if known. */
   virtual int64_t getKnownOffset() const = 0;
 
   /** Is this a store address operation (a subcategory of store operations which
@@ -208,7 +208,8 @@ class Instruction {
   /** What type of branch this instruction is. */
   BranchType branchType_ = BranchType::Unknown;
 
-  /** If the branch target is known at the time of decode, store it. */
+  /** The branch offset that may be known at the time of instruction decoding.
+   * The default value of 0 represents an unknown branch offset.*/
   int64_t knownOffset_ = 0;
 
   // Flushing
--- a/src/include/simeng/Instruction.hh
+++ b/src/include/simeng/Instruction.hh
@@ -209,7 +209,7 @@ class Instruction {
   BranchType branchType_ = BranchType::Unknown;
 
   /** If the branch target is known at the time of decode, store it. */
-  uint64_t knownTarget_ = 0;
+  uint64_t knownOffset_ = 0;
 
   // Flushing
   /** This instruction's sequence ID; a higher ID represents a chronologically
--- a/src/include/simeng/arch/aarch64/Instruction.hh
+++ b/src/include/simeng/arch/aarch64/Instruction.hh
@@ -297,7 +297,7 @@ class Instruction : public simeng::Instruction {
   /** Retrieve branch type. */
   BranchType getBranchType() const override;
 
-  /** Retrieve a branch target from the instruction's metadata if known. */
+  /** Retrieve a branch offset from the instruction's metadata if known. */
   int64_t getKnownOffset() const override;
 
   /** Is this a store address operation (a subcategory of store operations which
--- a/src/include/simeng/arch/aarch64/Instruction.hh
+++ b/src/include/simeng/arch/aarch64/Instruction.hh
@@ -298,7 +298,7 @@ class Instruction : public simeng::Instruction {
   BranchType getBranchType() const override;
 
   /** Retrieve a branch target from the instruction's metadata if known. */
-  uint64_t getKnownTarget() const override;
+  int64_t getKnownOffset() const override;
 
   /** Is this a store address operation (a subcategory of store operations which
    * deal with the generation of store addresses to store data at)? */
--- a/src/include/simeng/arch/riscv/Instruction.hh
+++ b/src/include/simeng/arch/riscv/Instruction.hh
@@ -125,7 +125,7 @@ class Instruction : public simeng::Instruction {
   /** Retrieve branch type. */
   BranchType getBranchType() const override;
 
-  /** Retrieve a branch target from the instruction's metadata if known. */
+  /** Retrieve a branch offset from the instruction's metadata if known. */
   int64_t getKnownOffset() const override;
 
   /** Is this a store address operation (a subcategory of store operations which
--- a/src/include/simeng/arch/riscv/Instruction.hh
+++ b/src/include/simeng/arch/riscv/Instruction.hh
@@ -126,7 +126,7 @@ class Instruction : public simeng::Instruction {
   BranchType getBranchType() const override;
 
   /** Retrieve a branch target from the instruction's metadata if known. */
-  uint64_t getKnownTarget() const override;
+  int64_t getKnownOffset() const override;
 
   /** Is this a store address operation (a subcategory of store operations which
    * deal with the generation of store addresses to store data at)? */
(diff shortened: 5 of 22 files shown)


src/lib/pipeline/ExecuteUnit.cc
```diff
@@ -161,6 +162,7 @@ void ExecuteUnit::execute(std::shared_ptr<Instruction>& uop) {
   forwardOperands_(uop->getDestinationRegisters(), uop->getResults());
 
   output_.getTailSlots()[0] = std::move(uop);
+  empty_ = true;
```
FinnWilkinson, line 165: An instruction completing execution does not mean the EU is empty. Given it has an internal pipeline (as seen in the `tick` function) you'll need to make sure that a) the internal pipeline is empty, and b) there are no instructions present in the `operationsStalled` structure too

test/regression/riscv/InorderPipeline.cc
```diff
@@ -0,0 +1,38 @@
+#include "RISCVRegressionTest.hh"
```
FinnWilkinson, line 1: This may be better suited in the `unit` test suite

src/include/simeng/pipeline/ExecuteUnit.hh
```diff
@@ -65,6 +65,8 @@ class ExecuteUnit {
   /** Retrieve the number of active execution cycles. */
   uint64_t getCycles() const;
 
+  bool isEmpty() const;
```
jj16791, line 68: Comment needed here

src/include/simeng/pipeline/ExecuteUnit.hh
```diff
@@ -133,6 +135,8 @@ class ExecuteUnit {
 
   /** The number of active execution cycles that were observed. */
   uint64_t cycles_ = 0;
+
+  bool empty_ = true;
```
jj16791, line 139: Comment needed here

jj16791 (changes requested):

src/lib/arch/riscv/Instruction_decode.cc
```diff
@@ -248,12 +262,12 @@ void Instruction::decode() {
...
     case Opcode::RISCV_BGEU:
       branchType_ = BranchType::Conditional;
-      knownTarget_ = instructionAddress_ + metadata.operands[2].imm;
+      knownOffset_ = metadata.operands[2].imm;
       break;
     case Opcode::RISCV_JAL:
     case Opcode::RISCV_JALR:
       branchType_ = BranchType::Unconditional;
-      knownTarget_ = instructionAddress_ + metadata.operands[1].imm;
+      knownOffset_ = metadata.operands[1].imm;
```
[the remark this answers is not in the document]
dANW34V3R: The original RISC-V implementation is wrong. ```knownTarget_``` should only be storing the offset from the current address as this is what the branch predictor expects; the AArch64 implementation reflects this. The updated name reflects this and removes the confusion about its functionality. A known target can not be produced here either as ```instructionAddress_``` is set after this function completes although I would be inclined to agree that storing a full target initially makes more sense

src/include/simeng/Instruction.hh
```diff
@@ -209,7 +209,7 @@ class Instruction {
   BranchType branchType_ = BranchType::Unknown;
 
   /** If the branch target is known at the time of decode, store it. */
-  uint64_t knownTarget_ = 0;
+  uint64_t knownOffset_ = 0;
```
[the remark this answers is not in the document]
dANW34V3R: See below. We should only ever have been storing the offset. @jj16791 may be able to better explain the effects of a negative offset as he knows the internals of the branch predictor more intimately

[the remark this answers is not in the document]
dANW34V3R: See below. We should only ever have been storing the offset. @jj16791 may be able to better explain the effects of a negative offset as he knows the internals of the branch predictor more intimately

src/lib/models/inorder/Core.cc
```diff
@@ -116,14 +115,31 @@ bool Core::hasHalted() const {
...
-  // handled.
-  bool decodePending = fetchToDecodeBuffer_.getHeadSlots()[0].size() > 0;
-  bool executePending = decodeToExecuteBuffer_.getHeadSlots()[0] != nullptr;
-  bool writebackPending = completionSlots_[0].getHeadSlots()[0] != nullptr;
+  // are no uops pending in any buffer, the execute unit is not currently
+  // processing an instructions and no exception is currently being handled.
+
+  // Units place uops in buffer tail, but if buffer is stalled, uops do not
+  // move to head slots and so remain in tail slots. Buffer head appears
+  // empty. Check emptiness accordingly
```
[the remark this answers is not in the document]
dANW34V3R: This is a neater solution so will implement this

src/lib/models/inorder/Core.cc
```diff
@@ -116,14 +115,31 @@ bool Core::hasHalted() const {
...
+  bool decodePending = decSlots.size() > 0;
+
+  auto exeSlots = decodeToExecuteBuffer_.isStalled()
+                      ? decodeToExecuteBuffer_.getTailSlots()[0]
+                      : decodeToExecuteBuffer_.getHeadSlots()[0];
+  bool executePending = exeSlots != nullptr;
+
+  auto writeSlots = completionSlots_[0].isStalled()
+                        ? completionSlots_[0].getTailSlots()[0]
+                        : completionSlots_[0].getHeadSlots()[0];
```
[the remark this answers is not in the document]
dANW34V3R: Will wait on answer to comment above

test/regression/riscv/InorderPipeline.cc
dANW34V3R: I did think about this but not sure of the best way to do it as it needs all of the ```RISCVRegressionTest.cc``` etc. infrastructure to run

src/include/simeng/Instruction.hh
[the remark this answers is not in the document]
dANW34V3R: Talking in person. The negative offset issue is in regards to the type e.g. "unsigned" int. What should the correct behaviour be here @jj16791 ?

src/include/simeng/pipeline/ExecuteUnit.hh
dANW34V3R: I have taken the solution from #298 as this check is correct

src/lib/pipeline/ExecuteUnit.cc
dANW34V3R: Taken functionality from #298

src/lib/arch/riscv/Instruction_decode.cc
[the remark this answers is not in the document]
FinnWilkinson: Noted, thanks for the explanation

test/regression/riscv/InorderPipeline.cc
FinnWilkinson: Noted, leave as is

test/regression/riscv/InorderPipeline.cc
```diff
@@ -0,0 +1,38 @@
...
+
+namespace {
+
+using inorderPipeline = RISCVRegressionTest;
+
+TEST_P(inorderPipeline, halt) {
```
FinnWilkinson, line 7: I think a more descript test name than "halt" is needed

src/include/simeng/pipeline/PipelineBuffer.hh
```diff
@@ -29,6 +29,14 @@ class PipelineBuffer {
     headIsStart = !headIsStart;
   }
 
+  const T* getPendingSlots() const {
+    if (isStalled_) {
+      return getTailSlots();
+    } else {
+      return getHeadSlots();
+    }
+  }
```
FinnWilkinson, line 38: In relation to https://github.com/UoB-HPC/SimEng/pull/294/files#r1144819612, could an `isEmpty` function which loops through all buffer slots and checks for the presence of any instruction not be implemented? Something like this would also work if the buffer was expanded to more than a head and tail (i.e. 3 or 4+ buffer slots)

src/include/simeng/Instruction.hh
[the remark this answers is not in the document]
FinnWilkinson: After offline discussion, this should be an `int64_t`

src/include/simeng/pipeline/PipelineBuffer.hh
dANW34V3R: I get the idea with an isEmpty function but am not sure it is so simple. As the function is templated it would be hard to define what constitutes a slot as empty e.g. in the in-order core there are ```pipelineBuffers``` containing ```Instructions``` as well as ```MacroOp```s. For the first, a slot is empty if it is a null pointer the other if the vector in a slot is of size 0. 

As for future proofing for a longer buffer; looking at the rest of the class it is very much tied to the idea of only having a length of 2 e.g. ```getTailSlots``` calculates a pointer based on the boolean value ```headIsStart```. I think an entire rework would be needed which is outside the scope of this PR.

FinnWilkinson: 1. Given an initial value is passed to the constructor, this could be stored as a member variable and refered to by `isEmpty`. These values are `nullptr`, empty vector etc.
2. I still think that making this change now is worth while. Even if the rest of the class needed updating, this would be one area that wouldn't.

src/lib/GenericPredictor.cc
```diff
@@ -36,7 +36,7 @@ BranchPrediction GenericPredictor::predict(uint64_t address, BranchType type,
   bool direction =
       btb_[hashedIndex].first < (1 << (satCntBits_ - 1)) ? false : true;
   uint64_t target =
-      (knownTarget != 0) ? address + knownTarget : btb_[hashedIndex].second;
+      (knownOffset != 0) ? address + knownOffset : btb_[hashedIndex].second;
```
FinnWilkinson, line 39: If `knownOffset` is a negative value, will the target be calculated correctly? i.e. is `address + knownOffset` calculated as signed or unsigned before casting implicitly to uint64_t

src/include/simeng/pipeline/PipelineBuffer.hh
dANW34V3R: This may be resolved by work done in the 32 bit support branch, which allows for changing the length of the buffer. I think we should leave as is in this branch. But implement what you are suggesting in the updated buffer

src/lib/GenericPredictor.cc
dANW34V3R: If result of the addition is +ve the result is as expected even if offset is a negative value. e.g. address = 4000, offset = -100, target will be 3900. If the result is negative then this will underflow leading to target being a huge value, but target should never be negative anyway

src/include/simeng/BranchPredictor.hh
```diff
@@ -49,7 +49,7 @@ class BranchPredictor {
   /** Generate a branch prediction for the specified instruction address with a
    * branch type and possible known target. */
```
jj16791, line 50: We should change "possible known target" to "possible known branch offset"

src/include/simeng/GenericPredictor.hh
```diff
@@ -30,7 +30,7 @@ class GenericPredictor : public BranchPredictor {
    * branch type, and a known target if not 0. Returns a branch direction and
    * branch target address. */
```
jj16791, line 31: *known branch offset

Not sure what the "if not 0" bit is in the comment. It would make some sense, but it is badly written still, if it had a default value of 0 in the func.

src/include/simeng/Instruction.hh
```diff
@@ -100,7 +100,7 @@ class Instruction {
   virtual BranchType getBranchType() const = 0;
 
   /** Retrieve a branch target from the instruction's metadata if known. */
```
jj16791, line 102: *branch offset

src/include/simeng/Instruction.hh
```diff
@@ -209,7 +209,7 @@ class Instruction {
   BranchType branchType_ = BranchType::Unknown;
 
   /** If the branch target is known at the time of decode, store it. */
```
jj16791, line 211: The comment above this sounds better suited to a function call. It may be better as "The branch offset that may be known at the time of instruction decoding. The default value of 0 represents an unknown branch offset."?

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -298,7 +298,7 @@ class Instruction : public simeng::Instruction {
   BranchType getBranchType() const override;
 
   /** Retrieve a branch target from the instruction's metadata if known. */
```
jj16791, line 300: *branch offset

src/include/simeng/arch/riscv/Instruction.hh
```diff
@@ -126,7 +126,7 @@ class Instruction : public simeng::Instruction {
   BranchType getBranchType() const override;
 
   /** Retrieve a branch target from the instruction's metadata if known. */
```
jj16791, line 128: *branch offset

src/lib/models/inorder/Core.cc
```diff
@@ -116,14 +115,24 @@ bool Core::hasHalted() const {
...
-  bool decodePending = fetchToDecodeBuffer_.getHeadSlots()[0].size() > 0;
-  bool executePending = decodeToExecuteBuffer_.getHeadSlots()[0] != nullptr;
-  bool writebackPending = completionSlots_[0].getHeadSlots()[0] != nullptr;
+  // are no uops pending in any buffer, the execute unit is not currently
+  // processing an instructions and no exception is currently being handled.
+
+  // Units place uops in buffer tail, but if buffer is stalled, uops do not
+  // move to head slots and so remain in tail slots. Buffer head appears
+  // empty. Check emptiness accordingly
+  auto decSlots = fetchToDecodeBuffer_.getPendingSlots()[0];
```
jj16791, line 124: There could be issues with this in the future regarding pipeline buffers of size 2 or more and the use of pipeline units which stall their buffers with uops still in the head slot. However, the inorder core is about to undergo some major changes, re complexity, and the logic in place to carry out the below check handled those above issues. Thus, we don't need to consider such issues for this PR.

jj16791 (changes requested):

src/include/simeng/GenericPredictor.hh
dANW34V3R: Updated typo. Should I add a default value or just remove that part of the comment?

src/include/simeng/Instruction.hh
dANW34V3R: Sounds good

FinnWilkinson (approved):

src/include/simeng/GenericPredictor.hh
jj16791: Given that we check against the offset being 0 as a way to detect no known offset, should probably have the default value. Should be on the BP unit to mandate that, no the function caller

rahahahat (approved):

jj16791 (approved):

PR #296, opened 2023-03-01
TLS fix for OMP clone syscall.

rahahahat:


src/include/simeng/arch/aarch64/Architecture.hh
```diff
@@ -15,6 +15,18 @@ namespace simeng {
 namespace arch {
 namespace aarch64 {
 
+/** Enum which holds AArch64 System register tags used by SimEng. */
+enum ARM64_SYSREG_TAGS : uint16_t {
+  DCZID_EL0 = 0,
```
jj16791, line 20: Could we change this to have the numbering be implicit

rahahahat: Not possible for enums. What would be the benefit of making the numbering implicit?

FinnWilkinson: Implicity value allocation is possible, and also looks cleaner

src/lib/OS/SimOS.cc
```diff
@@ -321,6 +321,9 @@ int64_t SimOS::cloneProcess(uint64_t flags, uint64_t stackPtr,
   } else if (Config::get()["Core"]["ISA"].as<std::string>() == "AArch64") {
     newProc->context_.regFile[arch::aarch64::RegisterType::GENERAL][31] = {
         stackPtr, 8};
+    newProc->context_.regFile[arch::aarch64::RegisterType::SYSTEM]
+                             [arch::aarch64::ARM64_SYSREG_TAGS::TPIDR_EL0] =
+        tls;
```
FinnWilkinson, line 326: Updated `TPIDR_EL0` value needs to be a reg value, i.e. `{tls, 8}`

src/lib/arch/aarch64/Architecture.cc
```diff
@@ -28,14 +28,14 @@ Architecture::Architecture() : microDecoder_(std::make_unique<MicroDecoder>()) {
                (config["Core"]["Timer-Frequency"].as<uint32_t>() * 1e6);
 
   // Generate zero-indexed system register map
```
FinnWilkinson, line 30: Update comment

FinnWilkinson (changes requested):

src/include/simeng/arch/aarch64/Architecture.hh
rahahahat: Sorry, I misunderstood what you said, I thought you meant do something similar to `systemRegisterMap_.size()`. Changing it

rahahahat: done

src/lib/arch/aarch64/Architecture.cc
rahahahat: done

jj16791 (approved):

PR #297, opened 2023-03-13
Changed use of RLIM_INFINITY to raw value to avoid difference of unde…

jj16791:
…rlying value on different host platforms

### DIFF
--- a/src/include/simeng/OS/Process.hh
+++ b/src/include/simeng/OS/Process.hh
@@ -14,8 +14,6 @@
 
 namespace simeng {
 
-static constexpr uint64_t RLIM_INF = 0xffffffffffffffff;
-
 // Forward declaration of class simeng::memory::Mem;
 namespace memory {
 class Mem;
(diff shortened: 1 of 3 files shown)


src/include/simeng/OS/Process.hh
```diff
@@ -184,8 +184,9 @@ class Process {
    * present, or by calling the `set_tid_address` syscall. */
   uint64_t clearChildTid_ = 0;
 
-  /** The rlimit struct for RLIMIT_STACK. */
-  rlimit stackRlim = {RLIM_INFINITY, RLIM_INFINITY};
+  /** The rlimit struct for RLIMIT_STACK. 0xffffffffffffffff used to represent
+   * RLIM_INFINITY in Linux. */
+  rlimit stackRlim = {0xffffffffffffffff, 0xffffffffffffffff};
```
FinnWilkinson, line 189: I think it is preferable to define RLIM_INF as a `static constexpr` at the top of this header in case this value needs to be used else where i.e. syscall handler

rahahahat: could be moved to the Constants.hh file

jj16791: Added

FinnWilkinson (approved):

src/include/simeng/OS/Process.hh
```diff
@@ -14,6 +14,8 @@
 
 namespace simeng {
 
+static constexpr uint64_t RLIM_INF = 0xffffffffffffffff;
```
rahahahat, line 17: if they are static consexpr i think they should be in `OS/Constants.hh`

jj16791: moved

FinnWilkinson (approved):

rahahahat (approved):

PR #298, opened 2023-03-13
Added check in inorder core to check whether the execution unit is em…

jj16791:
…pty before context switching


jj16791:
#re-run tests

jj16791:
#rerun tests

PR #300, opened 2023-03-22
Correctly Fill Auxiliary Vector

dANW34V3R:
This pull request fills the auxiliary vector with values for AT_PHDR, AT_PHENT, AT_PHNUM and AT_ENTRY as well as minor cleanup of Elf.hh and Elf.cc.

 Without this, programs compiled targeting RISC-V with gcc would not execute correctly as they did not know where the program headers were located.

### DIFF
--- a/src/lib/Elf.cc
+++ b/src/lib/Elf.cc
@@ -37,7 +37,7 @@ Elf::Elf(std::string path, char** imagePointer) {
   char fileMagic[4];
   file.read(fileMagic, 4);
   if (std::memcmp(elfMagic, fileMagic, sizeof(elfMagic))) {
-    std::cout << "[SimEng:Elf] Elf magic does not match" << std::endl;
+    std::cerr << "[SimEng:Elf] Elf magic does not match" << std::endl;
     return;
   }
 
@@ -50,7 +50,7 @@ Elf::Elf(std::string path, char** imagePointer) {
   char bitFormat;
   file.read(&bitFormat, sizeof(bitFormat));
   if (bitFormat != ElfBitFormat::Format64) {
-    std::cout << "[SimEng:Elf] Unsupported architecture detected in Elf"
+    std::cerr << "[SimEng:Elf] Unsupported architecture detected in Elf"
               << std::endl;
     return;
   }


src/include/simeng/Elf.hh
```diff
@@ -12,13 +12,13 @@ const char Format32 = 1;
...
 }  // namespace ElfBitFormat
 
-struct ElfHeader {
-  uint32_t type;
-  uint64_t offset;
-  uint64_t virtualAddress;
-  uint64_t physicalAddress;
-  uint64_t fileSize;
-  uint64_t memorySize;
+struct Elf64_Phdr {
```
rahahahat, line 15: Given the name of the member, variables have been changed to be Linux-specific, it isn't obvious what each one represents. It would be good to add comments explaining what each one represents as well as a small description about `Elf64_Phdr`

src/include/simeng/Elf.hh
```diff
@@ -29,10 +29,16 @@ class Elf {
   uint64_t getProcessImageSize() const;
   bool isValid() const;
   uint64_t getEntryPoint() const;
+  uint64_t getPhdrTableAddress() const;
+  uint64_t getPHENT() const;
+  uint64_t getPHNUM() const;
```
rahahahat, line 34: can we rename the functions to `getNumPhdr` and `getPhdrEntrySize`

src/include/simeng/kernel/LinuxProcess.hh
```diff
@@ -90,6 +114,15 @@ class LinuxProcess {
...
   uint64_t entryPoint_ = 0;
 
+  /** Program header table address */
+  uint64_t phdrTableAddress_;
+
+  /** Number of program headers */
+  uint64_t phnum_;
+
+  /** Size of program header entry */
+  uint64_t phent_;
```
rahahahat, line 124: Would be better to be consistent i.e rename to `phentsize_`

src/lib/Elf.cc
```diff
@@ -36,6 +37,7 @@ Elf::Elf(std::string path, char** imagePointer) {
   char fileMagic[4];
   file.read(fileMagic, 4);
   if (std::memcmp(elfMagic, fileMagic, sizeof(elfMagic))) {
+    std::cout << "[SimEng:Elf]   Elf magic does not match" << std::endl;
```
rahahahat, line 40: Extra space after "[SimEng:Elf]"

src/lib/Elf.cc
```diff
@@ -48,6 +50,8 @@ Elf::Elf(std::string path, char** imagePointer) {
   char bitFormat;
   file.read(&bitFormat, sizeof(bitFormat));
   if (bitFormat != ElfBitFormat::Format64) {
+    std::cout << "[SimEng:Elf]   Unsupported architecture detected in Elf"
```
rahahahat, line 53: Two spaces after "[SimEng:Elf]"

src/include/simeng/Elf.hh
```diff
@@ -29,10 +29,16 @@ class Elf {
...
 
  private:
   uint64_t entryPoint_;
-  std::vector<ElfHeader> headers_;
+  std::vector<Elf64_Phdr> pheaders_;
+  uint16_t e_phentsize;
+  uint16_t e_phnum;
+  uint64_t phdrTableAddress_ = 0;
```
FinnWilkinson, line 41: Although not really concerning this PR, could comments be added to member variables and functions inline with other header files in the project. It is good to update/fix these things as we find them

src/lib/Elf.cc
```diff
@@ -73,37 +77,36 @@ Elf::Elf(std::string path, char** imagePointer) {
    */
 
   // Seek to the byte representing the start of the header offset table.
-  uint64_t headerOffset;
-  file.read(reinterpret_cast<char*>(&headerOffset), sizeof(headerOffset));
+  uint64_t e_phoff;
```
FinnWilkinson, line 80: Although this is now inline with the linux implementation, it is much less readable. I think the name should be reverted back to `headerOffset`

src/include/simeng/kernel/LinuxProcess.hh
FinnWilkinson, line 124: Given these are not used outside setting the Aux-Vector, why do they need to be member variables?

src/lib/Elf.cc
```diff
@@ -73,37 +77,36 @@ Elf::Elf(std::string path, char** imagePointer) {
...
    */
 
   // Seek to the byte representing header entry size.
   file.seekg(0x36);
-  uint16_t headerEntrySize;
-  file.read(reinterpret_cast<char*>(&headerEntrySize), sizeof(headerEntrySize));
-  uint16_t headerEntries;
-  file.read(reinterpret_cast<char*>(&headerEntries), sizeof(headerEntries));
+  file.read(reinterpret_cast<char*>(&e_phentsize), sizeof(e_phentsize));
+  file.read(reinterpret_cast<char*>(&e_phnum), sizeof(e_phnum));
```
FinnWilkinson, line 97: Again, these names may be accurate to linux but they make the code less readable. Given there are public getter functions for these anyway with the accurate-to-linux naming, I think reverting these variable names to be inline with what was here previoiusly is preferable

src/include/simeng/Elf.hh
```diff
@@ -29,10 +29,16 @@ class Elf {
...
   bool isValid() const;
   uint64_t getEntryPoint() const;
+  uint64_t getPhdrTableAddress() const;
+  uint64_t getPHENT() const;
+  uint64_t getPHNUM() const;
 
  private:
   uint64_t entryPoint_;
-  std::vector<ElfHeader> headers_;
+  std::vector<Elf64_Phdr> pheaders_;
```
FinnWilkinson, line 38: Why `pheaders_`?

FinnWilkinson (changes requested):

src/lib/Elf.cc
```diff
@@ -133,22 +136,29 @@ Elf::Elf(std::string path, char** imagePointer) {
...
-    // of the entries are unused. Also SimEng internally treats these
+    // of the entries are unused. Also, SimEng internally treats these
     // virtual address as physical addresses to index into this large array.
-    if (header.virtualAddress + header.memorySize > processImageSize_) {
-      processImageSize_ = header.virtualAddress + header.memorySize;
+    if (header.p_vaddr + header.p_memsz > processImageSize_) {
+      processImageSize_ = header.p_vaddr + header.p_memsz;
+    }
+
+    // Find the table address used to populate the auxvec
```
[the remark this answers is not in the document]
dANW34V3R: Could you provide an example here. I was using the code you had in your branch so you are probably able to give a better description

src/lib/Elf.cc
dANW34V3R: I disagree, when working alongside the linux documentation and man pages this makes it much easier to follow what variables in SimEng correspond to. Once used to the naming convention it is also more descriptive of what that variable holds e.g. e_phoff holds the "P"rogram "H"eader "OFF"set that was held in the "E"lf header. Comments in the header should now mitigate initial decrease in readability

src/include/simeng/kernel/LinuxProcess.hh
dANW34V3R: I had the same initial thought. The elf object is not available outside of the LinuxProcess constructor and so is not available within createStack. Storing as a private member was the simplest solution here though they could be passed as arguments to the function. This does bring up the question of what should happen when using the other LinuxProcess constructor which does not use the elf object

src/lib/Elf.cc
dANW34V3R: See comment above

src/include/simeng/Elf.hh
dANW34V3R: According to the [man page](https://man7.org/linux/man-pages/man5/elf.5.html) the elf header contains information about program headers as well as section headers. Although sections headers are ignored, for clarity when referencing the documentation I changed the name.

               uint16_t      e_phentsize;
               uint16_t      e_phnum;
               uint16_t      e_shentsize;
               uint16_t      e_shnum;
               uint16_t      e_shstrndx;

src/include/simeng/Elf.hh
```diff
@@ -12,28 +12,83 @@ const char Format32 = 1;
...
+
+  /** A vector holding each of the program headers extracted from the ELF */
+  std::vector<Elf64_Phdr> pheaders_;
+
+  /** The program header entry size stored in the ELF header */
+  uint16_t e_phentsize;
+
+  /** The number of entries in the program header table stored in the ELF header
+   */
+  uint16_t e_phnum;
```
FinnWilkinson, line 83: member variables need `_` suffix

src/lib/Elf.cc
```diff
@@ -73,37 +77,38 @@ Elf::Elf(std::string path, char** imagePointer) {
    */
 
   // Seek to the byte representing the start of the header offset table.
-  uint64_t headerOffset;
-  file.read(reinterpret_cast<char*>(&headerOffset), sizeof(headerOffset));
+  // Holds the program header table's file offset in bytes.  If the file has no
+  // program header table, this member holds zero
```
FinnWilkinson, line 81: The comment says "this value holds zero". Is this enforced when `file.read` is performed or should we be initialising `e_phoff` to 0 ourselves?

FinnWilkinson (changes requested):

src/lib/Elf.cc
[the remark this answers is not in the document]
rahahahat: Could probably just say determine the virtual address of the header table in memory from individual program headers

src/include/simeng/Elf.hh
```diff
@@ -12,28 +12,83 @@ const char Format32 = 1;
...
-  uint64_t fileSize;
-  uint64_t memorySize;
+// Elf64_Phdr as described in the elf man page. Only contains SimEng relevant
+// information
+
+// An executable or shared object file's program header table is an array of
+// structures, each describing a segment or other information the system needs
+// to prepare the program for execution.  An object file segment contains one or
+// more sections. Program headers are meaningful only for executable and shared
+// object files.  A file specifies its own program header size with the ELF
```
jj16791, line 22: "A file specifies its own program header size with the ELF header's e_phentsize and e_phnum members". Isn't it just the e_phentsize which defines the size whilst e_phnum details the number of headers?

src/lib/Elf.cc
```diff
@@ -36,6 +37,7 @@ Elf::Elf(std::string path, char** imagePointer) {
   char fileMagic[4];
   file.read(fileMagic, 4);
   if (std::memcmp(elfMagic, fileMagic, sizeof(elfMagic))) {
+    std::cout << "[SimEng:Elf]  Elf magic does not match" << std::endl;
```
jj16791, line 40: One too many spaces after `]`

src/lib/Elf.cc
```diff
@@ -48,6 +50,8 @@ Elf::Elf(std::string path, char** imagePointer) {
   char bitFormat;
   file.read(&bitFormat, sizeof(bitFormat));
   if (bitFormat != ElfBitFormat::Format64) {
+    std::cout << "[SimEng:Elf]  Unsupported architecture detected in Elf"
```
jj16791, line 53: One too many spaces after `]`

src/lib/Elf.cc
```diff
@@ -73,37 +77,38 @@ Elf::Elf(std::string path, char** imagePointer) {
...
+  uint64_t e_phoff;
+  file.read(reinterpret_cast<char*>(&e_phoff), sizeof(e_phoff));
 
   /**
-   * Starting 54th byte of the ELF Header a 16-bit value indicates
-   * the size of each entry in the ELF Program header. In the `elf64_hdr`
-   * struct this value maps to the member `Elf64_Half e_phentsize`. All
-   * header entries have the same size.
+   * Starting from the 54th byte of the ELF Header a 16-bit value indicates
+   * the the size in bytes of one entry in the
```
jj16791, line 87: "the the"

jj16791: Is this the comment you were referring to re weird formatting? It doesn't look right.

src/lib/Elf.cc
```diff
@@ -73,37 +77,38 @@ Elf::Elf(std::string path, char** imagePointer) {
...
+   * file's program header table; all entries are the same
+   * size. In the `elf64_hdr`
+   * struct this value maps to the member `Elf64_Half e_phentsize`.
```
jj16791, line 90: There are a lot of references to elf64_hdr and now an Elf64_Half data type. Please link in the reference material for all this (similar to how we've done in the `createStack` function).

src/lib/Elf.cc
```diff
@@ -73,37 +77,38 @@ Elf::Elf(std::string path, char** imagePointer) {
...
    * Starting from the 56th byte a 16-bit value represents the number
```
jj16791, line 91: I think it would be better to split this comment up into two, one before the read in at the 54th byte and then one before the 56th byte

src/lib/kernel/LinuxProcess.cc
```diff
@@ -160,9 +164,23 @@ void LinuxProcess::createStack(char** processImage) {
...
+  initialStackFrame.push_back(numProgHeaders_);
+
+  initialStackFrame.push_back(auxVec::AT_PAGESZ);  // AT_PAGESZ
   initialStackFrame.push_back(pageSize_);
-  initialStackFrame.push_back(0);  // null terminator
+
+  initialStackFrame.push_back(auxVec::AT_ENTRY);  // AT_ENTRY
+  initialStackFrame.push_back(entryPoint_);
+
+  initialStackFrame.push_back(auxVec::AT_NULL);  // null terminator
```
jj16791, line 182: Do we need the AT_NULL value and the 0?

jj16791 (changes requested):

src/lib/Elf.cc
dANW34V3R: Initialised to 0 to be sure

src/include/simeng/Elf.hh
dANW34V3R: This is taken directly from the [man page](https://man7.org/linux/man-pages/man5/elf.5.html). You are correct in what you say. Reading again, I am unsure whether the comment is trying to say the total size of all the headers (e_phentsize * e_phnum) or just the size of one header (e_phentsize). Will update for the latter unless you disagree

src/lib/Elf.cc
dANW34V3R: Is it one space or two spaces as @rahahahat says?

src/lib/Elf.cc
dANW34V3R: Have managed to fix the formatting

src/lib/Elf.cc
dANW34V3R: I think this is already done on line 14 but could add again if you think it would be more clear

src/lib/kernel/LinuxProcess.cc
dANW34V3R: As in figure 22 from [this page](https://refspecs.linuxfoundation.org/ELF/zSeries/lzsabi0_zSeries/x895.html) each auxvec entry is a struct of a_type and a_un so for consistency I added a value here for a_un. In Table 14 however it says that this value is ignored. I think it does work without the 0 but it does keep inline with the struct. Tell me what you think is best

rahahahat:
Looks good to me

rahahahat (approved):

FinnWilkinson (approved):

src/lib/Elf.cc
```diff
@@ -36,6 +37,7 @@ Elf::Elf(std::string path, char** imagePointer) {
   char fileMagic[4];
   file.read(fileMagic, 4);
   if (std::memcmp(elfMagic, fileMagic, sizeof(elfMagic))) {
+    std::cout << "[SimEng:Elf] Elf magic does not match" << std::endl;
```
jj16791, line 40: Given this is an error, should output to cerr not cout

src/lib/Elf.cc
```diff
@@ -48,6 +50,8 @@ Elf::Elf(std::string path, char** imagePointer) {
   char bitFormat;
   file.read(&bitFormat, sizeof(bitFormat));
   if (bitFormat != ElfBitFormat::Format64) {
+    std::cout << "[SimEng:Elf] Unsupported architecture detected in Elf"
```
jj16791, line 53: Same issue with cout vs cerr

jj16791 (changes requested):
Minor comments on output

src/lib/kernel/LinuxProcess.cc
dANW34V3R: Here is a reduced printout of memory at the SP

0x3ffffffb00: 0x0000000000000001 = argc 0x0000003ffffffd21 = argv[0] (points to string containing full path to program) 
0x3ffffffb10: 0x0000000000000000  = NULL 0x0000003ffffffd51 = envp[0]
...
0x3ffffffbd0: 0x0000003fffffff64 0x0000003fffffff9a
0x3ffffffbe0: 0x0000003fffffffb5 0x0000000000000000 = envp NULL
0x3ffffffbf0: 0x0000000000000021 0x0000003ff7ffe000 = auxvec[0]
...
0x3ffffffc10: 0x0000000000000006 0x0000000000001000 = AT_PAGESZ, 4096
...
0x3ffffffcf0: 0x000000000000001f 0x0000003fffffffc8
0x3ffffffd00: 0x0000000000000000 0x0000000000000000 = AT_NULL
0x3ffffffd10: 0x34230600805cf300 0xab5c122963e0332f = Garbage

So the current implementation does follow gdb output

jj16791 (approved):

PR #303, opened 2023-03-28
RISC-V Single and Double Precision Floating Point extension

dANW34V3R:
This PR adds support for the F and D extensions as well as a dummy implementation of the Zicsr extension which is required by the former. The CSR extension cannot be completed as Capstone does not fully support disassembly of these instructions. Once a Capstone update is performed SimEng can be updated. Many spelling updates have also been performed.

Note: 8 tests are expected to fail as these rely on the correct rounding mode being set


FinnWilkinson:
Please rebase this branch on `dev` as it currently contains AuxVector PR fixes and has merge conflicts

src/include/simeng/arch/Architecture.hh
```diff
@@ -108,7 +108,8 @@ class Architecture {
 
   /** Updates System registers of any system-based timers. */
   virtual void updateSystemTimerRegisters(RegisterFileSet* regFile,
-                                          const uint64_t iterations) const = 0;
+                                          const uint64_t iterations,
+                                          const uint64_t retired) const = 0;
 };
```
FinnWilkinson, line 113: I have two thoughts on this: 1) the function now doesn't only update system timers so it should be renamed, 2) I can see this function growing and becoming messy in the future with args being required for all architectures but only used in one. i.e. `retired` is only needed by RISC-V but you must pass it in for AArch64 also.

If it could be made more agnostic somehow (maybe the function takes no arguments but then probes the core for the information it needs?) then that would be preferable

src/lib/arch/riscv/Instruction_decode.cc
```diff
@@ -19,6 +19,24 @@ namespace riscv {
...
   // Check from top of the range downwards
 
+  // Metadata could produce either 64-bit floating point register or 32-bit
+  // floating point register. Map both encodings to the same SimEng register.
+  // Only 64-bit registers are supported
+
+  // Modulus ensures only 64 bit registers are recognised
+  if (RISCV_REG_F31_64 >= reg && reg >= RISCV_REG_F0_64 && reg % 2 == 0) {
+    // Register ft0.64 has encoding 34 with subsequent encodings interleaved
+    // with 32 bit floating point registers. See riscv.h
```
jj16791, line 29: is that `riscv.h` from Capstone? If so, expand on the file name/directory to make it more clear

src/lib/models/emulation/Core.cc
```diff
@@ -159,7 +159,7 @@ void Core::tick() {
   }
 
   execute(uop);
-  isa_.updateSystemTimerRegisters(&registerFileSet_, ticks_);
+  isa_.updateSystemTimerRegisters(&registerFileSet_, ticks_, ticks_);
```
jj16791, line 162: Although this is subject to change based on Finn's prior comments, the 3rd param for the number of retired instructions should be `instructionsExecuted_` not `ticks_` I believe. It's not always true that `ticks_` == `instructionsExecuted_`

src/lib/models/outoforder/Core.cc
```diff
@@ -164,7 +164,8 @@ void Core::tick() {
 
   flushIfNeeded();
   fetchUnit_.requestFromPC();
-  isa_.updateSystemTimerRegisters(&registerFileSet_, ticks_);
+  isa_.updateSystemTimerRegisters(&registerFileSet_, ticks_,
+                                  writebackUnit_.getInstructionsWrittenCount());
```
jj16791, line 168: Similar to the comment in the emulation core, for the number of retired instructions, it should be `reorderBuffer_.getInstructionsCommittedCount()`

src/lib/arch/riscv/Instruction.cc
```diff
@@ -134,12 +135,16 @@ int64_t Instruction::getKnownOffset() const { return knownOffset_; }
...
+    base = InstructionGroups::FLOAT;
+  }
+
   if (isBranch()) return InstructionGroups::BRANCH;
   if (isLoad()) return base + 8;
   if (isStoreAddress()) return base + 9;
   if (isDivide_) return base + 7;
   if (isMultiply_) return base + 6;
-  if (isShift_) return base + 5;
+  if (isShift_ || isConvert_) return base + 5;
```
FinnWilkinson, line 147: Is there a reason that `isShift` and `isConvert` share a base?

src/lib/arch/riscv/InstructionMetadata.cc
```diff
@@ -252,6 +254,166 @@ void InstructionMetadata::alterPseudoInstructions(const cs_insn& insn) {
       }
       break;
     }
+
+    case Opcode::RISCV_CSRRS: {
+      if (operandCount == 1 && strcmp(mnemonic, "frflags") == 0) {
+        // frflags Rs is pseudo of CSRRS Rs, fflags, zero (Read FP exception
+        // flags) CSRRS Rs, _, _ -> CSRRS Rs, fflags, zero
+        operands[1].type =
+            RISCV_OP_IMM;  // TODO needs to become reg when CS updated
```
FinnWilkinson, line 263: Change `CS` to `Capstone` to make comment clearer. Apply throughout

test/regression/riscv/instructions/float.cc
```diff
@@ -0,0 +1,2086 @@
...
+  EXPECT_EQ(getFPRegister<float>(3), (float)0);
+  EXPECT_EQ(getFPRegister<uint64_t>(3), 0xffffffff00000000);
+  EXPECT_EQ(getFPRegister<uint64_t>(4), 0xffffffff00000000);
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    RISCV, InstFloat,
+    ::testing::Values(
+        std::make_tuple(EMULATION, YAML::Load("{}"))
+        //                      std::make_tuple(INORDER, YAML::Load("{}")),
```
jj16791, line 2081: If we don't need these, remove rather than comment out

src/lib/arch/riscv/InstructionMetadata.cc
```diff
@@ -252,6 +254,166 @@ void InstructionMetadata::alterPseudoInstructions(const cs_insn& insn) {
...
+        operands[1].reg = RISCV_SYSREG_FFLAGS;  // fflags address
+
+        operands[2].type = RISCV_OP_REG;
+        operands[2].reg = 1;
+
+        operandCount = 3;
+      } else if (strcmp(mnemonic, "rdinstret") == 0) {
+        assert(false && "Unimplemented psuedoinstruction rdinstret");
```
FinnWilkinson, line 271: All statements which print to the stdout or stderr need to be prefixed. Please add `[SimEng::InstructionMetadata]` to the beginning of all assert strings in this file

src/lib/arch/riscv/Instruction_decode.cc
```diff
@@ -19,6 +19,24 @@ namespace riscv {
...
+    return {RegisterType::FLOAT, static_cast<uint16_t>((reg - 34) / 2)};
+  }
+
+  // Modulus ensures only 32 bit registers are recognised
+  if (RISCV_REG_F31_32 >= reg && reg >= RISCV_REG_F0_32 && reg % 2 == 1) {
+    // Register ft0.32 has encoding 33 with subsequent encodings interleaved
+    // with 64 bit floating point registers. See riscv.h
+    return {RegisterType::FLOAT, static_cast<uint16_t>((reg - 33) / 2)};
+  }
```
FinnWilkinson, line 38: when doing `reg-34` etc, does the `34` represent the enum value of `ft0.64`? If so then it is best to use the actual enum name (`RISCV_REG_F31_64`). 

Also, how can you ensure that dividing by 2 will always be a whole number before you static cast it to a uint?

src/lib/arch/riscv/Instruction_decode.cc
```diff
@@ -40,25 +58,7 @@ Register csRegToRegister(unsigned int reg) {
...
-      metadata.opcode <= Opcode::RISCV_SRAW)
-    return;
-  if (metadata.opcode >= Opcode::RISCV_SRL &&
-      metadata.opcode <= Opcode::RISCV_SW)
-    return;
-  if (metadata.opcode >= Opcode::RISCV_XOR &&
-      metadata.opcode <= Opcode::RISCV_XORI)
-    return;
-  if (metadata.opcode == Opcode::RISCV_FENCE) return;
+  return;
```
FinnWilkinson, line 61: Given this function now just does `return;`, can it be removed as the TODO comment states?

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
       break;
     }
 
+      // Control and Status Register extension (Zicsr)
+      // Currently do not read-modify-write ATOMICALLY
+      // Left unimplemented due to Capstone being unable to disassemble CSR
+      // addresses
```
FinnWilkinson, line 847: Comment syas un-implemented but it looks like there is an implementation?

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
...
+          default:
+            // Invalid Case
+            // TODO any subsequent attempt to execute a floating-point operation
+            // with a dynamic rounding mode will raise an illegal instruction
+            // exception.
+            std::cerr << "[SimEng:RISC-V:Execute] Invalid rounding mode"
+                      << std::endl;
+            break;
+        }
+      } else {
```
jj16791, line 883: Remove empty else case

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
...
+            break;
+          case 4:
+            fesetround(FE_TONEAREST);
+            break;
+          default:
+            // Invalid Case
+            // TODO any subsequent attempt to execute a floating-point operation
+            // with a dynamic rounding mode will raise an illegal instruction
+            // exception.
+            std::cerr << "[SimEng:RISC-V:Execute] Invalid rounding mode"
```
FinnWilkinson, line 879: Update prefix to be inline with the rest in the project i.e. `[SimEng:Instruction_execute]`. PLease apply throughout

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
...
+                      << std::endl;
+            break;
+        }
+      } else {
+        //        std::cerr << "[SimEng:RISC-V:Execute] Unknown system register:
+        //        "
+        //                  << metadata.operands[1].reg << std::endl;
```
FinnWilkinson, line 886: Remove commented out code

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
...
+      // operands are correctly NaN-boxed, i.e., all upper FLEN−n bits are 1. If
+      // so, the n least-significant bits of the input are used as the input
+      // value, otherwise the input value is treated as an n-bit canonical NaN."
+
+      // TODO need to take rounding mode into account as these can be set per
+      // instruction https://en.cppreference.com/w/cpp/numeric/fenv/FE_round
+      //      std::fesetround(FE_TONEAREST);
+
+      // Single-Precision Floating-Point (F)
+      // Double-Precision Floating-Point (D)
```
FinnWilkinson, line 926: Can we take rounding mode into account now? Or is there something blocking this making it a TODO?

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
...
+    case Opcode::RISCV_FSD: {  // FSD rs1,rs2,imm
+      memoryData[0] = operands[0];
+      break;
+    }
+    case Opcode::RISCV_FSW: {  // FSW rs1,rs2,imm
+      memoryData[0] = operands[0];
+      break;
+    }
+    case Opcode::RISCV_FLD: {  // FLD rd,rs1,imm
+      // Note: elements of memory data are RegisterValue's
```
FinnWilkinson, line 936: Comment not needed as `memoryData` is not templated to be anything other than a vector of RegisterValues

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
...
+      results[0] = memoryData[0].get<double>();
+      break;
+    }
+    case Opcode::RISCV_FLW: {  // FLW rd,rs1,imm
+      // Note: elements of memory data are RegisterValue's
```
FinnWilkinson, line 941: Comment not needed as `memoryData` is not templated to be anything other than a vector of RegisterValues

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
...
+        results[0] = RegisterValue(fmin(rs1, rs2), 8);
+      }
+
+      break;
+    }
+    case Opcode::RISCV_FMIN_S: {  // FMIN.S rd,rs1,rs2
+      const float rs1 = operands[0].get<float>();
+      const float rs2 = operands[1].get<float>();
+
+      // As fmin.d
```
FinnWilkinson, line 1049: Is this comment needed? Currently it doesn't add much context to the code

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
...
+      }
+
+      results[0] = RegisterValue(res, 8);
+      break;
+    }
+    case Opcode::RISCV_FMAX_S: {  // FMAX.S rd,rs1,rs2
+      const float rs1 = operands[0].get<float>();
+      const float rs2 = operands[1].get<float>();
+
+      // As fmax_d
```
FinnWilkinson, line 1080: Is this comment needed? Currently it doesn't add much context to the code

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
...
+      results[0] = RegisterValue(signExtendW(rs1), 8);
+      break;
+    }
+
+      // TODO FLT.S and FLE.S perform what the IEEE 754-2008 standard refers to
+      // as signaling comparisons: that is, they set the invalid operation
+      // exception flag if either input is NaN. FEQ.S performs a quiet
+      // comparison: it only sets the invalid operation exception flag if either
+      // input is a signaling NaN. For all three instructions, the result is 0
+      // if either operand is NaN.
```
FinnWilkinson, line 1396: IS there anything blocking this TODO?

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
...
+    case Opcode::RISCV_FEQ_D: {  // FEQ.D rd,rs1,rs2
+      // TODO FEQ.S performs a quiet
+      // comparison: it only sets the invalid operation exception flag if
+      // either input is a signaling NaN. Qemu doesn't seem to set CSR flags
+      // with sNANs
```
FinnWilkinson, line 1401: Is there something blocking this TODO?

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +841,642 @@ void Instruction::execute() {
...
+        results[0] = RegisterValue(static_cast<uint64_t>(0), 8);
+      }
+      break;
+    }
+    case Opcode::RISCV_FLT_D: {  // FLT.D rd,rs1,rs2
+      const double rs1 = operands[0].get<double>();
+      const double rs2 = operands[1].get<double>();
+
+      if (std::isnan(rs1) || std::isnan(rs2)) {
+        // TODO set csr flag
```
FinnWilkinson, line 1428: Is there something blocking this TODO and all similar csr TODOs?

FinnWilkinson (changes requested):

jj16791 (changes requested):
Not too many comments. However, I do agree with Finn's upcoming comment about the update system timer register function change

src/lib/arch/riscv/Instruction_decode.cc
dANW34V3R: Enum encoding used.

The last clause of each ```if``` requires ````reg``` to be either odd OR even. In the even case we subtract 34; even - even = even and so even / 2 is whole. In the odd case we subtract 33; odd - odd = even and so even / 2 is whole. So in both cases the output of the computation is whole before being cast to uint

src/lib/arch/riscv/Instruction.cc
dANW34V3R: I could split these and move the isConvert_ check into the ```if(isFloat())``` clause above if that makes if more readable

src/lib/arch/riscv/Instruction_execute.cc
dANW34V3R: There is an argument to be made to keep ```:RISC-V:``` in some form at least to distinguish which instruction_execute is being referred to. Thoughts?

src/lib/arch/riscv/Instruction_execute.cc
dANW34V3R: Rounding modes per instruction would require a potentially slow and slightly messy implementation as this would need to be checked and adjusted for every instruction (and then reset after its execution). Rounding mode info would need to be extracted from the raw bytes, far from impossible but less than ideal. The codes we run at the moment don't seem to be reliant on precise rounding  and so this was left out to reduce complexity in this initial implementation. The TODO remains to keep track of this however

src/lib/arch/riscv/Instruction_execute.cc
dANW34V3R: The control and status register implementation is not complete/non-existent and so does not support updating CSR values. The comment also adds that Qemu doesn't perform this behaviour and so I am not entirely sure what is the correct behaviour

src/lib/arch/riscv/Instruction_execute.cc
dANW34V3R: The control and status register implementation is not complete/non-existent and so does not support updating CSR values

dANW34V3R: I suggest I look at the new Huawei CSR implementation as this seems more complete. I can convert this into a PR later but leave CSR instructions with dummy implementations for now as they are not integral to the codes that we are currently running but some implementation is required for progression

src/lib/arch/riscv/Instruction.cc
FinnWilkinson: I'm a bit confused at your explanation, but I think splitting it so that `isShift` and `isConvert` have differet offsets (i.e. how in AArch64 all types have their own offset). This is mainly for the fact that in future instructions _could_ be added which require `isShift` and `isConvert` to be seperate...

src/lib/arch/riscv/Instruction_decode.cc
FinnWilkinson: OK understood. Could you add a brief comment somewhere here with this same information to avoid any confusion or doubt in the future?

src/lib/arch/riscv/Instruction_execute.cc
FinnWilkinson: In my opinion, given the user should know the ISA their binary targets it should be implicit which instruction_execute is referred to, plus it decreases the length of the prefix.

Interested to hear others thoughts on this - I'm happy to include ISA specific prefixes where necessary if thats the consensus @jj16791 @rahahahat @ABenC377 @JosephMoore25

src/lib/arch/riscv/Instruction_execute.cc
FinnWilkinson: Are there a fixed number of rounding modes? And are they always encoded in the same bits for every instruction? 

If so then a simple general function could be implemented which does a bitwise mask on the instruction bytes and returns the rounding mode based on this? If it can be this simple then it would be interesting to see what the performance hit is

src/lib/arch/riscv/Instruction_execute.cc
FinnWilkinson: Could you update the comment to make this blocking factor clearer?

src/lib/arch/riscv/Instruction.cc
dANW34V3R: After offline discussion this will be left as is, but another PR should be created to clean up InstructionGroup logic

dANW34V3R: See issue #347

src/lib/arch/riscv/Instruction_decode.cc
dANW34V3R: Removed for cleanliness but may need to be re-implemented as Capstone updates and supports more extensions

src/lib/arch/riscv/Instruction_execute.cc
dANW34V3R: Updated comment

src/include/simeng/arch/Architecture.hh
dANW34V3R: The main reason I changed this was to play around with the control status registers, however I think the concept off what I am doing will persist e.g. updating "system" registers with information from the core. 1) should be an easy enough fix e.g. ```updateSytemRegisters``` or ```updateSystemRegistersFromCore```. 

2) becomes a bit harder, I can see your point but now have thought about the execution of it. I think the current function (and how it may expand in the future) could be seen as decently agnostic - from the core, pass in all the information it has and then the ISA can choose what it does with that information. Performance wise, I can't see the unused parameters having any noticeable impact.

The other suggested option is to pass the core object itself to the ISA and probe for information. This would require creating a new getter in the core per new value required (not terrible, but slightly more complicated than passing as a parameter). Currently the ISA has access to the register file set, passed as a parameter, and updates the register file itself. If the core were the sole parameter then there would be no access to the register file parameter without making it a public field which seems dangerous to me as the core now doesn't get to control how this is updated. Another option would be to pass both the register file set and the core object into the functions which seems a bit odd but I guess would work. Probing would theoretically be slower as you need to make a function call per value on every cycle (resp. to having access directly) although I don't think this will be particularly noticeable.

@FinnWilkinson @jj16791 Thoughts on how to proceed?

FinnWilkinson: We could pass the core object to a `updateSystemRegisters` function and then make the Archiecture class a Friend class of Core. This would allow the architecture to access protected and private member variables of Core simply by doing `core.piv_mem_var`

src/lib/arch/riscv/InstructionMetadata.cc
FinnWilkinson: Sorry, this mistake is on me... the prefix should actually be `[SimEng:InstructionMetadata]`

src/lib/arch/riscv/Instruction_execute.cc
FinnWilkinson: Could you change all these comments "TODO: set csr flag when Zicsr implementation is complete" to add clarity?

src/lib/arch/riscv/Instruction_execute.cc
dANW34V3R: Implementation done

src/lib/models/emulation/Core.cc
dANW34V3R: Have removed this parameter in light of the coming update to this function and the RISC-V CSR implementation. I'll keep this and the following related comments in mind when implementing this

src/lib/models/outoforder/Core.cc
dANW34V3R: See above

src/lib/arch/riscv/Instruction_execute.cc
jj16791: I'd vote to keep it without the ISA in the prefix as it's implied by the config file and I see it as redundant information for the user. 

However, as an aside, if we do switch to this sort of inclusion then it would also make sense to add a similar thing for core types, all files inside the arch-specific folders (e.g. `ExceptionHandler`), etc.

jj16791 (changes requested):
Some failed stages on Jenkins which need to be solved and some minor comments.

jj16791:
In addition to the last comment, I get an assertion trigger in the aarch64 test suite on the Mac Studio:
```
2: [ RUN      ] AArch64/InstStore.st1_multi_struct/emulation
2:
2: [SimEng:ExceptionHandler] Encountered execution not-yet-implemented exception
2: [SimEng:ExceptionHandler]   Generated by instruction:
2: [SimEng:ExceptionHandler]     0x000000000000001c: 00 00 00 00     udf #0
2: [SimEng:ExceptionHandler]       opcode ID: 5643
2: [SimEng:Core] Halting due to fatal exception
2:
2: [SimEng:ExceptionHandler] Encountered execution not-yet-implemented exception
2: [SimEng:ExceptionHandler]   Generated by instruction:
2: [SimEng:ExceptionHandler]     0x0000000000000024: 00 00 00 00     udf #0
2: [SimEng:ExceptionHandler]       opcode ID: 5643
2: [SimEng:Core] Halting due to fatal exception
2:
2: [SimEng:ExceptionHandler] Encountered execution not-yet-implemented exception
2: [SimEng:ExceptionHandler]   Generated by instruction:
2: [SimEng:ExceptionHandler]     0x000000000000001c: 00 00 00 00     udf #0
2: [SimEng:ExceptionHandler]       opcode ID: 5643
2: [SimEng:Core] Halting due to fatal exception
2: Assertion failed: (bytes > 0 && "Attempted to access an uninitialised RegisterValue"), function getAsVector, file RegisterValue.hh, line 94.
2/4 Test #2: regression-aarch64-test ..........Subprocess aborted***Exception:   4.77 sec
```
 Please resolve if this persists after future commits.

FinnWilkinson:
> In addition to the last comment, I get an assertion trigger in the aarch64 test suite on the Mac Studio:
> 
> ```
> 2: [ RUN      ] AArch64/InstStore.st1_multi_struct/emulation
> 2:
> 2: [SimEng:ExceptionHandler] Encountered execution not-yet-implemented exception
> 2: [SimEng:ExceptionHandler]   Generated by instruction:
> 2: [SimEng:ExceptionHandler]     0x000000000000001c: 00 00 00 00     udf #0
> 2: [SimEng:ExceptionHandler]       opcode ID: 5643
> 2: [SimEng:Core] Halting due to fatal exception
> 2:
> 2: [SimEng:ExceptionHandler] Encountered execution not-yet-implemented exception
> 2: [SimEng:ExceptionHandler]   Generated by instruction:
> 2: [SimEng:ExceptionHandler]     0x0000000000000024: 00 00 00 00     udf #0
> 2: [SimEng:ExceptionHandler]       opcode ID: 5643
> 2: [SimEng:Core] Halting due to fatal exception
> 2:
> 2: [SimEng:ExceptionHandler] Encountered execution not-yet-implemented exception
> 2: [SimEng:ExceptionHandler]   Generated by instruction:
> 2: [SimEng:ExceptionHandler]     0x000000000000001c: 00 00 00 00     udf #0
> 2: [SimEng:ExceptionHandler]       opcode ID: 5643
> 2: [SimEng:Core] Halting due to fatal exception
> 2: Assertion failed: (bytes > 0 && "Attempted to access an uninitialised RegisterValue"), function getAsVector, file RegisterValue.hh, line 94.
> 2/4 Test #2: regression-aarch64-test ..........Subprocess aborted***Exception:   4.77 sec
> ```
> 
> Please resolve/investigate if this persists after future commits. Doesn't happen in `dev` for me.

This is one of the tests which was failing for me and that I fixed in PR #339 (see [here](src/lib/arch/aarch64/InstructionMetadata.cc))

src/include/simeng/arch/riscv/Architecture.hh
```diff
@@ -14,6 +14,19 @@ namespace simeng {
...
+  RISCV_SYSREG_FFLAGS = 0x001,
+  RISCV_SYSREG_FRM = 0x002,
+  RISCV_SYSREG_FCSR = 0x003,
+
+  RISCV_SYSREG_CYCLE = 0xC00,
+  RISCV_SYSREG_TIME = 0xC01,
+  RISCV_SYSREG_INSTRET = 0xC02,
+
+} riscv_sysreg;
+
```
FinnWilkinson, line 29: Are system registers being included in this PR still?

src/include/simeng/arch/Architecture.hh
FinnWilkinson: Resolving as change removed

FinnWilkinson (changes requested):

src/include/simeng/arch/riscv/Architecture.hh
```diff
@@ -95,6 +108,12 @@ class Architecture : public arch::Architecture {
 
   /** A reference to a Linux kernel object to forward syscalls to. */
   kernel::Linux& linux_;
+
+  /** System Register of Processor Cycle Counter. */
+  simeng::Register cycleSystemReg_;
+
+  /** System Register of Processor Retired Counter. */
+  simeng::Register retiredSystemReg_;
```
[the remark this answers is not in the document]
dANW34V3R: No in this PR, but potentially will be used in CSR PR, will remove for cleanliness. The Huawei implementation could be different also

src/lib/arch/riscv/Instruction_execute.cc
dANW34V3R: Somehow missed on find and replace. Should be done now

src/lib/arch/riscv/Architecture.cc
```diff
@@ -16,6 +16,10 @@ std::forward_list<InstructionMetadata> Architecture::metadataCache;
 
 Architecture::Architecture(kernel::Linux& kernel, YAML::Node config)
     : linux_(kernel) {
+  // Set initial rounding mode for F/D extensions
+  // TODO set fcsr accordingly when Zicsr extension supported
```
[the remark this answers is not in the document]
dANW34V3R: It is in the map but doesn't actually do anything atm. Waiting for the CSR extension to be implemented

src/lib/arch/riscv/InstructionMetadata.cc
```diff
@@ -252,6 +254,184 @@ void InstructionMetadata::alterPseudoInstructions(const cs_insn& insn) {
...
+               "[SimEng:InstructionMetadata] Unimplemented psuedoinstruction "
+               "csrr");
+      } else if (strcmp(mnemonic, "csrs") == 0) {
+        assert(false &&
+               "[SimEng:InstructionMetadata] Unimplemented psuedoinstruction "
+               "csrs");
+      } else if (strcmp(mnemonic, "frcsr") == 0) {
+        assert(false &&
+               "[SimEng:InstructionMetadata] Unimplemented psuedoinstruction "
+               "rdtime");
```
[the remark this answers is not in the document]
dANW34V3R: It is along the lines of an unimplemented instruction so yes should probably end the simulation. Will make them cause a NYI exception.

As for assertions only triggering in Debug mode I completely agree, these can go unnoticed when not in the correct mode. However, as a wider issue, this style has been used throughout and there are many places that probably should end the simulation but are rather only caught in debug mode. I have seen this causing problems with some of the work we have received from outside the group (as well as sometime inside) which is why I made a point about developing in Debug mode a while back. Something that needs some discussion @jj16791 @JosephMoore25 @ABenC377 @rahahahat

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +903,787 @@ void Instruction::execute() {
...
 
+      // Control and Status Register extension (Zicsr)
+
+      // Currently do not read-modify-write ATOMICALLY
+      // Left mostly unimplemented due to Capstone being unable to disassemble
+      // CSR addresses. Some partial functionality is implemented for
+      // correctness of other extensions
+    case Opcode::RISCV_CSRRW: {  // CSRRW rd,csr,rs1
+
+      if (metadata.operands[1].reg == 2) {
```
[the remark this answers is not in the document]
dANW34V3R: There should be some more complex functionality to update the value stored in the system register given by ```operands[1].reg```. As the Zicsr extension hasn't been implemented currently this doesn't happen and Rd is set to 0 as dummy logic to allow progression. Full execution logic will be provided in the Zicsr PR

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +903,787 @@ void Instruction::execute() {
...
+    }
+    case Opcode::RISCV_CSRRWI: {  // CSRRWI rd,csr,imm
+      executionNYI();
+      break;
+    }
+    case Opcode::RISCV_CSRRS: {  // CSRRS rd,csr,rs1
+      // dummy implementation
+      results[0] = RegisterValue(static_cast<uint64_t>(0), 8);
+      break;
+    }
```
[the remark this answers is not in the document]
dANW34V3R: One of our codes (I can't remember which one) will use this instruction, but the value didn't matter. This implementation allowed progression. None of the other CSR instructions have been used

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +903,787 @@ void Instruction::execute() {
...
+
+      if (metadata.operands[1].reg == 2) {
+        // Update CPP rounding mode but not floating point CSR as currently no
+        // implementation
+
+        // TODO this won't work properly in OoO core as rounding mode could be
+        // updated before instructions earlier in program order execute
+        // causing them to be rounded incorrectly. Rounding mode needs to be
+        // set on commit and all subsequent instructions must be performed
+        // with updated mode. e.g. this should be performed atomically
```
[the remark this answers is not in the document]
dANW34V3R: Will make an attempted implementation

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +903,787 @@ void Instruction::execute() {
...
+        res = fmaxf(rs1, rs2);
+      }
+
+      results[0] = RegisterValue(NanBoxFloat(res), 8);
+      break;
+    }
+
+    case Opcode::RISCV_FMADD_D: {  // FMADD.D rd,rs1,rs2,rs3
+      // The fused multiply-add instructions must set the invalid operation
+      // exception flag when the multiplicands are ∞ and zero, even when the
```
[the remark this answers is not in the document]
dANW34V3R: This is copy pasted from the spec. Will change

src/lib/arch/riscv/Instruction_execute.cc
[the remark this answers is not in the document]
FinnWilkinson: Understood. A brief comment such as `TODO: implement fully when Zicsr extension is supported` would be good

FinnWilkinson (changes requested):
This is looking good. Noticed a few more things on most recent pass through

src/lib/arch/riscv/Instruction.cc
```diff
@@ -31,7 +37,10 @@ Instruction::Instruction(const Architecture& architecture,
 Instruction::Instruction(const Architecture& architecture,
                          const InstructionMetadata& metadata,
                          InstructionException exception)
-    : architecture_(architecture), metadata(metadata) {
+    : architecture_(architecture),
+      metadata(metadata),
+      exception_(metadata.metadataException_) {
+  exceptionEncountered_ = metadata.metadataExceptionEncountered_;
```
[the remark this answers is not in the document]
dANW34V3R: Yeah, after the comment above I realised that this implementation is completely redundant so have removed. I tried using the initialiser list syntax but this caused errors as ```exceptionEncountered_``` is a member of the parent class.

src/lib/arch/riscv/InstructionMetadata.cc
```diff
@@ -252,10 +254,200 @@ void InstructionMetadata::alterPseudoInstructions(const cs_insn& insn) {
...
+        aliasNYI();
+        assert(false &&
+               "[SimEng:InstructionMetadata] Unimplemented psuedoinstruction "
+               "fscsr");
+      } else if (strcmp(mnemonic, "fscsr") == 0) {
+        aliasNYI();
+        // 2 pseudoinstructions with same name but different number of registers
+        assert(false &&
+               "[SimEng:InstructionMetadata] Unimplemented psuedoinstruction "
+               "fscsr");
```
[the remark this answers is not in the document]
dANW34V3R: Good spot, updated

src/lib/arch/riscv/ExceptionHandler.cc
```diff
@@ -640,6 +641,64 @@ bool ExceptionHandler::init() {
...
+              break;
+          }
+        }
+
+        // Dummy logic to allow progression. Set Rd to 0
+        stateChange = {ChangeType::REPLACEMENT, {destinationRegs[0]}, {0ull}};
+        break;
+      default:
+        // Make simulation stop
+        break;
```
dANW34V3R, line 699: Need to fill out this case

src/include/simeng/arch/riscv/Instruction.hh
```diff
@@ -158,19 +164,25 @@ class Instruction : public simeng::Instruction {
...
   /** The maximum number of source registers any supported RISC-V instruction
    * can have. */
-  static const uint8_t MAX_SOURCE_REGISTERS = 2;
+  static const uint8_t MAX_SOURCE_REGISTERS = 3;
   /** The maximum number of destination registers any supported RISC-V
    * instruction can have. */
   static const uint8_t MAX_DESTINATION_REGISTERS = 1;
 
+  /* Return the array of operand values */
+  const std::array<RegisterValue, MAX_SOURCE_REGISTERS> getOperands() const;
```
FinnWilkinson, line 187: Unless I'm missing something this function is un-implemented? And also seems redundant

src/lib/arch/riscv/Instruction.cc
```diff
@@ -14,14 +14,20 @@ const Register Instruction::ZERO_REGISTER = {RegisterType::GENERAL, 0};
...
 }
 
 Instruction::Instruction(const Architecture& architecture,
                          const InstructionMetadata& metadata, uint8_t latency,
                          uint8_t stallCycles)
-    : architecture_(architecture), metadata(metadata) {
+    : architecture_(architecture),
+      metadata(metadata),
+      exception_(metadata.metadataException_) {
+  exceptionEncountered_ = metadata.metadataExceptionEncountered_;
```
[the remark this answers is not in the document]
FinnWilkinson: If the constructors are un-used then I would vote to remove them.

In AArch64 there is also the concept of not yet implemented alias'. It may be worth checking out how that is implemented to be consistent across the project, and will prevent the need for a "metadata exception"

src/include/simeng/arch/riscv/Instruction.hh
dANW34V3R: It is implemented at the bottom of ```instruction_execute.cc``` which is not the correct place for it so moved to ```instruction.cc```. It is used in he exception handler by the new functionality line 653

src/lib/arch/riscv/ExceptionHandler.cc
```diff
@@ -633,14 +634,71 @@ bool ExceptionHandler::init() {
...
         std::cout << "\n[SimEng:ExceptionHandler] Unrecognised syscall: "
                   << syscallId << std::endl;
         return fatal();
     }
 
+    return concludeSyscall(stateChange);
+  } else if (exception == InstructionException::AtomicOperation) {
+    // TODO this may complete in only a single cycle and not the value from the
+    // config file. The instruction should be completely reissued and allowed to
+    // execute with latency
```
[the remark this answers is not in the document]
dANW34V3R: Yes, my thought with the comment is that it will apply the latency when going through the EU. This is when the instruction is logically only raising the exception and while other instructions are being executed OoO at the same time. The instruction should be performing its actual logic once the pipeline has been "serialised" (once it is at the head of the ROB). In the code this happens in the exception handler which doesn't (as far as I am aware) account for the latency. It should only be a minor accuracy inconsistency, but an inconsistency none the less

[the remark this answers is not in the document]
FinnWilkinson: I see. Maybe the comment prefix should be `NOTE` then rather than `TODO`?

src/lib/arch/riscv/Instruction.cc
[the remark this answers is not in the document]
FinnWilkinson: Replying to [this comment](https://github.com/UoB-HPC/SimEng/pull/303#issuecomment-1838937835) : 

I see. More information being printed for NYI alias' would be beneficial. @jj16791 your thoughts on this?

src/lib/arch/riscv/ExceptionHandler.cc
```diff
@@ -633,14 +634,71 @@ bool ExceptionHandler::init() {
...
+
+    // Retrieve metadata, operand values and destination registers from
+    // instruction
+    auto metadata = instruction_.getMetadata();
+    auto operands = instruction_.getOperands();
```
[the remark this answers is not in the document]
dANW34V3R: Ahh yeah, I see your point now. I added a function which is already there. Removed getOperands and using getSourceOperands here

src/include/simeng/arch/riscv/Instruction.hh
```diff
@@ -47,7 +48,9 @@ enum class InstructionException {
   SupervisorCall,
   HypervisorCall,
   SecureMonitorCall,
-  NoAvailablePort
+  NoAvailablePort,
+  IllegalInstruction,
+  AtomicOperation
```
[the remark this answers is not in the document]
dANW34V3R: @jj16791 @JosephMoore25

src/lib/arch/riscv/ExceptionHandler.cc
dANW34V3R: Done

src/lib/arch/riscv/Instruction.cc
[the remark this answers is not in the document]
dANW34V3R: I can update the AArch64 implementation to follow this if we think this is a good idea

jj16791 (changes requested):

src/include/simeng/arch/riscv/Instruction.hh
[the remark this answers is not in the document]
JosephMoore25: I agree it probably shouldn't be `ProcessStateUpdate`, given we already have `ProcessStateChange`, and that it should also change. I'm not sure what cases this exception gets generated in. If it is due to a process state change, then even `ProcessStateFlush` could be a little more descriptive.  That being said, this could imply you're clearing the process state. Further thought required...

JosephMoore25 (changes requested):
Not much to say that hasn't already been mentioned - it looks 98% of the way there. My biggest concern is the rounding modes. We want to ensure correctness while maintaining speed, and I'm slightly worried that the current implementation doesn't account for both.

That being said, I don't want this to stop the PR. Given this is a current roadblock, if it's a *correct* implementation then I'm happy with it once the remaining changes are made. 

Sail may overwrite some of this work later down the line, so imo progressing this PR would allow big features in other PRs to be added on the sooner side that will have more impact than perfecting this.

src/lib/arch/riscv/Instruction.cc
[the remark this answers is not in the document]
dANW34V3R: Should I copy over the functionality? And instead of reusing the ```ExecutionNotYetImplemented``` exception should I add a new ```AliasNotImplemented``` exception to provide more precise detail?
@jj16791 @FinnWilkinson

src/include/simeng/arch/riscv/Instruction.hh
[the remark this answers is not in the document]
dANW34V3R: The original intent was specifically for atomic operations (csr instructions, like this one, are meant to execute atomically) which may require some more specific logic later down the line regarding the ```NOTE```/```TODO`` at the top of the conditional body. I'm fine changing it now but I reckon it will likely change again in the near future when we have a clearer idea of how this will actually work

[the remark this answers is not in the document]
dANW34V3R: Thumbs up for ```PipelineFlush```, thumbs down for ```AtomicOperation``` will change to whatever has the majority vote

src/lib/arch/riscv/ExceptionHandler.cc
```diff
@@ -633,14 +634,71 @@ bool ExceptionHandler::init() {
...
+            case 3:  // RUP Round up (+infinity)
+              fesetround(FE_UPWARD);
+              break;
+            case 4:  // RMM Round to nearest, ties to max magnitude
+              // FE_TONEAREST ties towards even but no other options available
+              // in fenv
+              fesetround(FE_TONEAREST);
+              break;
+            default:
+              // Invalid Case
```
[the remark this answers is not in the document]
dANW34V3R: The first part of the comment is a quote from the spec, I've added some quotes to try and make this more clear but can add more info if needed. The second part of the comment is a little more explanation and the solution. Nothing should happen when an incorrect dynamic rounding mode is set, an exception should only be raised if another instruction attempts to use it. The implementation I have in mind requires being able to set the CSR (which hasn't been implemented yet) so the comment serves as a reminder to implement this later.

In any case, it would require very poor code generation from the compiler to actually hit this case.

src/lib/arch/riscv/ExceptionHandler.cc
```diff
@@ -823,6 +881,12 @@ void ExceptionHandler::printException(const Instruction& insn) const {
     case InstructionException::NoAvailablePort:
       std::cout << "unsupported execution port";
       break;
+    case InstructionException::IllegalInstruction:
+      std::cout << "illegal instruction";
+      break;
+    case InstructionException::AtomicOperation:
```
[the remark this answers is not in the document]
dANW34V3R: I have used this a simple way to give a print out if the execution logic of an instruction raises this exception but doesn't have a case in the ```switch``` in the exception handler. See the default case, line 681 ```riscv/ExceptionHandler.cc```

src/lib/arch/riscv/ExceptionHandler.cc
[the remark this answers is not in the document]
jj16791: That makes more sense now, thank you for the clarification.

src/lib/arch/riscv/ExceptionHandler.cc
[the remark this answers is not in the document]
jj16791: Oh I see, I misread the switch case bounds

src/lib/arch/riscv/Instruction.cc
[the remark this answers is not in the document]
dANW34V3R: Implemented. As for removing unused constructors this is also the case for the AArch64 Instruction constructor which takes an exception. Should this be removed also @FinnWilkinson @jj16791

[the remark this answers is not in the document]
jj16791: The second constructor which takes an exception is needed for things like:
```
    uop = std::make_shared<Instruction>(*this, metadataCache.front(),
                                        InstructionException::MisalignedPC);
```
so shouldn't be removed

src/lib/arch/aarch64/InstructionMetadata.hh
```diff
@@ -84,6 +84,13 @@ struct InstructionMetadata {
   /** The number of explicit operands. */
   uint8_t operandCount;
 
+  /** The current exception state of this instruction. */
+  simeng::arch::aarch64::InstructionException metadataException_ =
```
jj16791, line 88: `simeng::arch::aarch64::` is unnecessary given the `namespace` usage

jj16791 (changes requested):

src/lib/arch/aarch64/InstructionMetadata.hh
dANW34V3R: Of course, think this has remained after an initial implementation. Removed

src/lib/arch/riscv/Instruction.cc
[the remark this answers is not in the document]
dANW34V3R: Ah yes, this is the case in RISC-V also. Seems like IDEs have trouble finding this usage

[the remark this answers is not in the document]
dANW34V3R: I can't find a usage of the second RISC-V constructor so removed

src/lib/arch/riscv/Instruction_execute.cc
```diff
@@ -827,6 +903,787 @@ void Instruction::execute() {
...
+      executionNYI();
+      break;
+    }
+
+      // TODO "Apart from transfer operations ... all other floating-point
+      // operations on narrower n-bit operations, n < FLEN, check if the input
+      // operands are correctly NaN-boxed, i.e., all upper FLEN−n bits are 1.
+      // If so, the n least-significant bits of the input are used as the
+      // input value, otherwise the input value is treated as an n-bit
+      // canonical NaN."
```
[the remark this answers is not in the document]
dANW34V3R: This is another quote from the spec which talks about functionality that is needed for correctness but an implementation could be slow (and possibly never used if the code being run is "correct", which I believe we do assume?). I didn't know exactly how the rest of the F/D extension would look at the time of adding this comment so it hasn't been tackled. I'll give an implementation a go again and we can see if we want to keep it. I don't think it will have too much impact on performance

src/lib/arch/riscv/Instruction_execute.cc
dANW34V3R: @rahahahat @ABenC377 @JosephMoore25

src/lib/arch/riscv/InstructionMetadata.cc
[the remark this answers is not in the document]
dANW34V3R: This should be written down somewhere and discussed on Friday so that the comment can be resolved. Just not sure where

src/lib/arch/aarch64/InstructionMetadata.hh
```diff
@@ -84,6 +84,12 @@ struct InstructionMetadata {
   /** The number of explicit operands. */
   uint8_t operandCount;
 
+  /** The current exception state of this instruction. */
+  InstructionException metadataException_ = InstructionException::None;
+
+  /** Whether an exception has been encountered. */
+  bool metadataExceptionEncountered_ = false;
```
FinnWilkinson, line 91: These should be private with appropriate getters given they should only be modified by `InstructionMetadata.cc`

src/lib/arch/riscv/InstructionMetadata.hh
```diff
@@ -52,30 +53,40 @@ struct InstructionMetadata {
...
+
   /** The number of explicit operands. */
   uint8_t operandCount;
 
+  /** The current exception state of this instruction. */
+  InstructionException metadataException_ = InstructionException::None;
+
+  /** Whether an exception has been encountered. */
+  bool metadataExceptionEncountered_ = false;
+
```
FinnWilkinson, line 83: These should be private with appropriate getters given they should only be modified by `InstructionMetadata.cc`

src/lib/arch/riscv/ExceptionHandler.cc
```diff
@@ -823,6 +887,12 @@ void ExceptionHandler::printException(const Instruction& insn) const {
     case InstructionException::NoAvailablePort:
       std::cout << "unsupported execution port";
       break;
+    case InstructionException::IllegalInstruction:
+      std::cout << "illegal instruction";
+      break;
+    case InstructionException::PipelineFlush:
+      std::cout << "unknown atomic operation";
```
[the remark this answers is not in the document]
FinnWilkinson: Adding a TODO comment will help us to track this in the future (hopefully!)

FinnWilkinson (changes requested):
Looking very good. Last few minor details

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -224,6 +224,7 @@ enum class InstructionException {
   EncodingUnallocated,
   EncodingNotYetImplemented,
   ExecutionNotYetImplemented,
+  AliasNotYetImplemented,
```
FinnWilkinson, line 227: Could you ensure there are tests implemented to ensure this exception gets triggered correctly with the expected result / console printout

test/regression/riscv/instructions/float.cc
```diff
@@ -200,6 +200,30 @@ TEST_P(InstFloat, FDIV_S) {
...
+    fdiv.s fa5, fa3, fa3  # 1 / 1 = 1
+    fdiv.s fa6, fa4, fa3  # Incorrect NaN box should be caught by fdiv and
+                          # canonical NaN used as input, NaN / 1 = NaN
+   )");
+
+  EXPECT_EQ(getFPRegister<uint64_t>(13), 0xffffffff3f800000);
+  EXPECT_EQ(getFPRegister<uint64_t>(14), 0xfff7ffff3f800000);
+
+  EXPECT_EQ(getFPRegister<uint64_t>(15), 0xffffffff3f800000);
+  EXPECT_EQ(getFPRegister<uint64_t>(16), 0xffffffff7fc00000);
```
FinnWilkinson, line 226: I presume `0xffffffff7fc00000` is the hex code for NAN?

test/regression/riscv/instructions/float.cc
```diff
@@ -648,6 +771,32 @@ TEST_P(InstFloat, FCVT_WU_S) {
...
+                          # canonical NaN used as input, fcvt NaN = 2^32 − 1
+)");
+
+  EXPECT_EQ(getFPRegister<uint64_t>(13), 0xffffffff3f800000);
+  EXPECT_EQ(getFPRegister<uint64_t>(14), 0xfff7ffff3f800000);
+
+  EXPECT_EQ(getGeneralRegister<uint64_t>(5), 0x0000000000000001);
+  EXPECT_EQ(getGeneralRegister<uint32_t>(6),
+            (uint64_t)pow(2, 32) - 1);  // Expected result from spec
+  EXPECT_EQ(getGeneralRegister<uint64_t>(6), 0xFFFFFFFFFFFFFFFF);
```
FinnWilkinson, line 799: Why do you check register 6 twice; once as uint32 and once as uint64?

test/regression/riscv/instructions/float.cc
dANW34V3R: This is the hex code for a single precision NaN which has been NaN boxed to fit in a double precision register. Comment needed?

src/include/simeng/arch/aarch64/Instruction.hh
FinnWilkinson: Resolving as implementing this is being done as part of PR #352

src/include/simeng/arch/riscv/Instruction.hh
```diff
@@ -42,12 +44,15 @@ enum class InstructionException {
...
+  AliasNotYetImplemented,
   MisalignedPC,
   DataAbort,
   SupervisorCall,
   HypervisorCall,
   SecureMonitorCall,
-  NoAvailablePort
+  NoAvailablePort,
+  IllegalInstruction,
+  PipelineFlush
```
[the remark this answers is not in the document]
FinnWilkinson: Resolving as implementing this is being done as part of PR #352

test/regression/riscv/instructions/float.cc
FinnWilkinson: Could be useful? Or use a static constexpr / define ?

test/regression/riscv/instructions/float.cc
dANW34V3R: Strictly speaking only the hex check is necessary. The cast shouldn't be there so removed. The instructions produces a result of size word (32 bits) and the spec describes this as being the value 2^32-1 so this is what the first check is for, this is a bit more human readable and easier to align with the spec. The result should also be sign extended, so the full 64 bits are checked against the hex. I could reduce this to one check and describe the rest in a short comment

test/regression/riscv/instructions/float.cc
dANW34V3R: constexpr sounds good

test/regression/riscv/instructions/float.cc
FinnWilkinson: Two checks is fine, was just curious. Removing the cast though was needed yeah

src/lib/arch/riscv/Instruction_execute.cc
dANW34V3R: ISA printed in SimEng metadata so no need  to be within individual error messages

FinnWilkinson (approved):
There are _some_ minor consistency issues with CoreInstance changes, but as this class is soon being changed / removed its not worth blocking this PR for them

FinnWilkinson (dismissed):

src/include/simeng/CoreInstance.hh
```diff
@@ -75,6 +75,8 @@ class CoreInstance {
   /** Getter for the set simulation mode in a string format. */
   const std::string getSimulationModeString() const;
 
+  const std::string getISAString() const;
```
JosephMoore25, line 78: Add quick comment: `/** Getter for the ISA in string format */`
Just makes this file more consistent, even if it's self explanatory

src/include/simeng/CoreInstance.hh
```diff
@@ -169,6 +171,8 @@ class CoreInstance {
    */
   std::string modeString_ = "Emulation";
 
+  std::string instructionSetArchString_ = "";
```
JosephMoore25, line 174: `/** A string format for the instruction set in use */`

JosephMoore25 (changes requested):
Looks good, nice work. Two minor comments that should take minimal time for consistency and then I'm happy to approve

docs/sphinx/developer/arch/supported/riscv.rst
```diff
@@ -78,3 +83,17 @@ An example of this would be the pseudoinstruction ``not rd, rs``. This is implem
 This must be fixed in the ``InstructionMetadata`` constructor. A new entry should be added to the switch statement and the pseudoinstruction mnemonic checked. The correct set of operands can then be set. A couple of helper functions are used for common operand fixes.
 
 To ensure all pseudoinstructions are accounted for, the table in chapter 25 of the `RISC-V Unprivileged specification <https://riscv.org/technical/specifications/>`_ should be checked. It is recommended to implement all pseudoinstructions for all currently implemented instructions.
+
+Rounding Modes
+**************
+
+RISC-V floating point instructions can use either static or dynamic rounding modes. The former embedded as 3 bits within the instruction encoding, and the later held as 3 bits of the ``fcsr`` system register.
```
jj16791, line 90: *latter

jj16791 (changes requested):
In agreeance with @FinnWilkinson approval comment but the upcoming SimInfo PR will change those changes so best to re-discuss implementation on that PR or on the future CoreInstance PR.

jj16791 (dismissed):

JosephMoore25 (approved):
Looks good overall, nice work

FinnWilkinson (approved):

JosephMoore25 (approved):

jj16791 (approved):

FinnWilkinson (approved):

PR #306, opened 2023-03-31
Fix Incorrect Loop Buffer Behaviour

dANW34V3R:
When running Tealeaf compiled to RISC-V a series of memcpy's are performed. The main body of a memcpy is a short loop, loading from one memory location and storing to another byte by byte. As this short loop is repeated it causes the loop buffer to be filled which is then used to supply decode. On one of these calls to memcpy, the loop buffer is filled just as the loop should finish iterating.

In the original implementation when the loop buffer is in the state FILLING the fetch unit fetches the final branch of this loop. As this is at the loop boundary address the loop buffer eagerly changes state to SUPPLYING and stops fetching more instructions. It disregards the branch predictors' "not taken" prediction which should cause the instruction at pc+4 to be fetched; instead the fetch tick is completed for this cycle and in the next cycle the loop buffer supplies decode (along with the not taken prediction given when filling) which then ultimately causes an infinite loop which grows SimEngs memory until the host system memory fills and the OS kills it.

To fix this, when the loop buffer should finish filling the branch prediction is checked to determine whether the buffer should be used to supply decode or not.

### DIFF
--- a/src/lib/pipeline/FetchUnit.cc
+++ b/src/lib/pipeline/FetchUnit.cc
@@ -141,12 +141,21 @@ void FetchUnit::tick() {
           {encoding, bytesRead, pc_, macroOp[0]->getBranchPrediction()});
 
       if (pc_ == loopBoundaryAddress_) {
-        // loopBoundaryAddress_ has been fetched whilst filling the loop buffer.
-        // Stop filling as loop body has been recorded and begin to supply
-        // decode unit with instructions from the loop buffer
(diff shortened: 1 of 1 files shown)


jj16791 (approved):

FinnWilkinson (approved):

rahahahat (approved):

jj16791:
#re-run tests

jj16791:
#rerun tests

PR #313, opened 2023-04-25
Config File update

FinnWilkinson:


configs/DEMO_RISCV.yaml
```diff
@@ -39,8 +36,15 @@ Branch-Predictor:
   Fallback-Static-Predictor: "Always-Taken"
   Branch-Predictor:
   BTB-bitlength: 16
-LSQ-L1-Interface:
-  Access-Latency: 4
+Memory-Hierarchy:
+  Cache-Line-Width: 128
+  L1-Data:
```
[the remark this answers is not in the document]
FinnWilkinson: Why would each cache need a cache line width? Isn't there a system wide one?

test/integration/VirtMemTests.cc
```diff
@@ -13,15 +13,16 @@ const std::string configStr =
...
     "Process-Image: {Heap-Size: 100000, Stack-Size: 100000, Mmap-Size: "
-    "200000}, Simulation-Memory: {Size: 500000}, CPU-Info: "
+    "200000}, Memory-Hierarchy: {Cache-Line-Width: 64, L1-Data: "
+    "{Access-Latency: 1, Size: 500000}}, CPU-Info: "
     "{Generate-Special-Dir: False}}";
 
 TEST(VirtMemTest, MmapSysCallNoAddressNoFile) {
   Config::set(configStr.c_str());
 
   // Create simulation memory.
```
rahahahat, line 23: Maybe expand the comment by saying that we use cache params for simulation memory just so that testing is faster

rahahahat: Please expand in each test

src/include/simeng/Config.hh
```diff
@@ -16,14 +16,15 @@ class Config {
...
-   "RAS-entries: 5, Fallback-Static-Predictor: 2}, LSQ-L1-Interface: "         \
-   "{Access-Latency: 4, Exclusive: False, Load-Bandwidth: 32, "                \
+   "10485760}, Register-Set: {GeneralPurpose-Count: 154, "                     \
+   "FloatingPoint/SVE-Count: 90, Predicate-Count: 17, Conditional-Count: "     \
+   "128, Matrix-Count: 2}, Pipeline-Widths: {Commit: 4, FrontEnd: 4, "         \
+   "LSQ-Completion: 2}, Queue-Sizes: {ROB: 180, Load: 64, Store: 36}, "        \
+   "Branch-Predictor: {BTB-Tag-Bits: 11, Saturating-Count-Bits: 2, "           \
+   "Global-History-Length: 10, RAS-entries: 5, Fallback-Static-Predictor: "    \
+   "2}, Memory-Hierarchy: {Cache-Line-Width: 256, L1-Data: {Access-Latency: "  \
+   "4, Size: 65536}, DRAM: {Access-Latency: 200, Size: 2684354560}}, "         \
```
rahahahat, line 26: Simulation memory size was 104857600 and the L1 cache size is 65536, and as far as I can see for all tests L1 cache params are used for all instances `simeng::memory::Mem`. Are you sure that this decrement in size won't cause any problems with the test suite?

src/tools/simeng/main.cc
```diff
@@ -84,9 +84,10 @@ int main(int argc, char** argv) {
 
   // Get the memory size from the YAML config file.
   const size_t memorySize =
-      Config::get()["Simulation-Memory"]["Size"].as<size_t>();
-  const uint64_t latency =
-      Config::get()["LSQ-L1-Interface"]["Access-Latency"].as<uint64_t>();
+      Config::get()["Memory-Hierarchy"]["DRAM"]["Size"].as<size_t>();
+  const uint16_t latency =
```
rahahahat, line 88: Add a comment stating why L1 cache latency us used for DRAM

test/regression/RegressionTest.cc
```diff
@@ -33,7 +33,7 @@ void RegressionTest::run(const char* source, const char* triple,
   Config::set(config);
 
   const size_t memorySize =
-      Config::get()["Simulation-Memory"]["Size"].as<size_t>();
+      Config::get()["Memory-Hierarchy"]["L1-Data"]["Size"].as<size_t>();
```
rahahahat, line 36: Add a comment why size of memory is L1-Size and not DRAM-size

src/lib/ModelConfig.cc
```diff
@@ -421,20 +410,44 @@ void ModelConfig::validate() {
...
+  std::vector<std::string> subSubFields = {"Access-Latency", "Size"};
+  nodeChecker<uint16_t>(configFile_[root][subFields[0]], subFields[0],
+                        std::make_pair(1, UINT16_MAX), ExpectedValue::UInteger,
+                        64);
+  nodeChecker<uint16_t>(configFile_[root][subFields[1]][subSubFields[0]],
+                        subSubFields[0], std::make_pair(1, UINT16_MAX),
+                        ExpectedValue::UInteger, 1);
+  // Default L1-Data size is 1024 * 32 = 32 KiB
+  uint16_t cacheLineLen = configFile_[root][subFields[0]].as<uint16_t>();
+  nodeChecker<uint64_t>(
```
rahahahat, line 425: Should we allow such large L1 cacheSizes? surely uint16_t will suffice?

FinnWilkinson: Yes, something like the M1 firestorm core has 128KiB of L1 Data cache per core

test/integration/VirtMemTests.cc
FinnWilkinson: Changed config to use DRAM rather than L1 to avoid confusion

src/include/simeng/Config.hh
FinnWilkinson: This is the default config file used for simulation, not the test suite. The config files used for the test suite have been configured with appropriate memory. Additionally, test suite classes have been updated to use the DRAM option rather than L1-Data to avoid confusion

src/tools/simeng/main.cc
FinnWilkinson: Done

test/regression/RegressionTest.cc
FinnWilkinson: Changed config to use DRAM rather than L1 to avoid confusion

src/tools/simeng/main.cc
rahahahat: can we not use DRAM here as well, as mentioned we can specify access latency of DRAM as whatever we want through the config?
lets keep it to 5 till we have a good cache hierarchy setup?

configs/a64fx.yaml
```diff
@@ -41,8 +38,15 @@ Branch-Predictor:
...
   Fallback-Static-Predictor: "Always-Taken"
-LSQ-L1-Interface:
-  Access-Latency: 5
+Memory-Hierarchy:
+  Cache-Line-Width: 256
+  L1-Data:
+    Access-Latency: 5
+    Size: 65536 # 64 KiB
+  DRAM:
+    Access-Latency: 200
```
rahahahat, line 47: Can we not put the access latency of DRAM to 5 for the time being as that's the only thing being used currently? Since there  are caches we can just put a comment in each .yaml file stating values not in use yet

FinnWilkinson: Done

src/tools/simeng/main.cc
FinnWilkinson: Done

rahahahat (approved):

PR #315, opened 2023-05-02
Unaigned PC exception fix

FinnWilkinson:
After the updates made for multi-threading-support and memory-overhaul, the programByteLength definition (used by the fetch unit to determine when to not fetch) had changed to the actual length of the process's region rather than the end of the .data section that it had been before.

This meant that any PC value which exceeded the data section (i.e. via speculated fetches on incorrect branch predictions) would be fetched and processed by the decoder, where in some cases when the address branched to was not properly aligned and as such triggered an unaligned PC exception.

To fix this, the context associated to each process now contains `dataSectionEnd`, which is then used by the fetch unit to determine if a PC is valid before sending a request to memory.

### DIFF
--- a/src/lib/OS/Process.cc
+++ b/src/lib/OS/Process.cc
@@ -396,7 +396,7 @@ void Process::initContext(
   context_.TID = TID_;
   context_.pc = entryPoint_;
   context_.sp = stackPtr;
-  context_.progByteLen = getProcessImageSize();
+  context_.dataSectionEnd = getHeapStart();
   // Initialise all registers to 0
   size_t numTypes = regFileStructure.size();
   context_.regFile.reserve(numTypes);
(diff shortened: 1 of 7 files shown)


jj16791 (approved):

src/lib/OS/Process.cc
```diff
@@ -396,7 +396,7 @@ void Process::initContext(
   context_.TID = TID_;
   context_.pc = entryPoint_;
   context_.sp = stackPtr;
-  context_.progByteLen = getProcessImageSize();
+  context_.dataSectionEnd = getHeapStart();
```
rahahahat, line 399: With the virtual memory setup, there is a padding between of `PAGE_SIZE` bytes between the maximum address of the `.data` section and heap. Do you think it can lead to problems in the future? Would `getHeapStart - PAGE_SIZE` work?

FinnWilkinson: Good spot, changed

PR #317, opened 2023-05-15
[WIP] 32 bit support

dANW34V3R:
This is a temporary PR to review initial 32 bit support provided by Huawei. Information regarding the changes made is available in README_RV32.md

### DIFF
--- a/src/include/simeng/RegisterValue.hh
+++ b/src/include/simeng/RegisterValue.hh
@@ -26,10 +26,16 @@ class RegisterValue {
    * number of bytes (defaulting to the size of the template type). */
   template <class T,
             typename std::enable_if_t<!std::is_pointer_v<T>, T>* = nullptr>
-  RegisterValue(T value, uint16_t bytes = sizeof(T)) : bytes(bytes) {
+  RegisterValue(T value, uint16_t bytes = sizeof(T), bool relaxFor32 = true) : bytes(bytes) {
+    relaxedFor32bit_ = relaxFor32;
+    std::memset(this->value, 0, MAX_LOCAL_BYTES);
     if (isLocal()) {
       T* view = reinterpret_cast<T*>(this->value);
-      view[0] = value;
+      if (sizeof(T) > bytes) { // e.g. when T is int64 and bytes is 4
+        std::memcpy(this->value, &value, bytes);
+      } else {
+        view[0] = value;
+      }
 
       if (bytes > sizeof(T)) {
         // Zero the remaining bytes not set by the provided value
@@ -90,11 +96,16 @@ class RegisterValue {
    * the specified datatype. */
   template <class T>
   const T* getAsVector() const {
-    static_assert(alignof(T) <= 8 && "Alignment over 8 bytes not guranteed");
+    static_assert(alignof(T) <= 8 && "Alignment over 8 bytes not guaranteed");
     assert(bytes > 0 && "Attempted to access an uninitialised RegisterValue");
-    assert(sizeof(T) <= bytes &&
-           "Attempted to access a RegisterValue as a datatype larger than the "
-           "data held");
+    assert((sizeof(T) <= bytes || (bytes == 4 && sizeof(T) == 8)) && "Attempted"
+           " to access a RegisterValue as a datatype larger than the "
+           "data held" );
+    if(!relaxedFor32bit_) { // maybe #ifdef if it makes slower?
+      assert(sizeof(T) <= bytes &&
+        "Attempted to access a RegisterValue as a datatype larger than the "
+        "data held");
+    }
     if (isLocal()) {
       return reinterpret_cast<const T*>(value);
     } else {
@@ -129,6 +140,9 @@ class RegisterValue {
   /** The underlying local member value. Aligned to 8 bytes to prevent
    * potential alignment issue when casting. */
   alignas(8) char value[MAX_LOCAL_BYTES];
+
+  /** Switch for different assert checking */
+  bool relaxedFor32bit_;
 };
 
 }  // namespace simeng
--- a/src/include/simeng/arch/Architecture.hh
+++ b/src/include/simeng/arch/Architecture.hh
@@ -14,6 +14,12 @@ using MacroOp = std::vector<std::shared_ptr<Instruction>>;
 
 namespace arch {
 
+/** Modes. Assume only has 32-bit and 64-bit. */
+enum arch_mode {
+  ARCH_32BIT=1,
+  ARCH_64BIT=0
+};
+
 /** The types of changes that can be made to values within the process state. */
 enum class ChangeType { REPLACEMENT, INCREMENT, DECREMENT };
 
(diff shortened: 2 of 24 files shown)


jj16791 (commented):
A good start to integrating the rv32 ISA. Most of the riscv namespace changes look good, I'm unable to comment on the validity of the logic/functionality due to my lacking knowledge of the ISA. There are some changes which I feel could be combined (e.g. the logic for ELF32/ELF64) to reduce code duplication and improve readability.

I don't think we'd be happy with upstreaming the tracing logic in its current format. We've had some internal discussions regarding its implementation but haven't had time to settle on a decision on how to do it. Also, the use of 32/64 bit architecture variables in the parent architecture class doesn't match our current idea of an agnostic arch class. Will have to think about moving this notion to be isolated to the riscv arch class or re-evaluating the content of the parent arch class.

The changes to the emulation core seem good (Other than the RISCV-specific logic which I'm aware will be removed) and the new pipeline buffer class looks promising.

Some of the changes to the `RegisterValue.hh` header are confusing, we'll need a better explanation for the changes made before moving forward with any decision.

Finally, there's a general need for code clean-up/proper formatting applied throughout the project.

src/include/simeng/arch/Architecture.hh
```diff
@@ -14,6 +14,12 @@ using MacroOp = std::vector<std::shared_ptr<Instruction>>;
 
 namespace arch {
 
+/** Modes. Assume only has 32-bit and 64-bit. */
+enum arch_mode {
```
rahahahat, line 18: Please use the enum class to establish an explicit type, and parent it on uint8_t to reduce size usage.
Also please capitalise arch_mode. 

You can static_cast the enum values to be used as an index or anything else.

src/include/simeng/RegisterValue.hh
```diff
@@ -26,10 +26,16 @@ class RegisterValue {
    * number of bytes (defaulting to the size of the template type). */
   template <class T,
             typename std::enable_if_t<!std::is_pointer_v<T>, T>* = nullptr>
-  RegisterValue(T value, uint16_t bytes = sizeof(T)) : bytes(bytes) {
+  RegisterValue(T value, uint16_t bytes = sizeof(T), bool relaxFor32 = true) : bytes(bytes) {
```
rahahahat, line 29: I think the below statement should be added to this constructor to avoid the case when the bytes argument provided is less than `sizeof(T)`

```
bytes = std::max(bytes, sizeof(T));
```
I think having bytes being less than `sizeof(T)` doesn't make sense and is dangerous; As it will lead to data truncation. Which I'm not sure why would be needed if a smaller data type can be used.  Even if something like this is intentional I still think it should not be allowed.

src/include/simeng/RegisterValue.hh
```diff
@@ -90,11 +96,16 @@ class RegisterValue {
    * the specified datatype. */
   template <class T>
   const T* getAsVector() const {
-    static_assert(alignof(T) <= 8 && "Alignment over 8 bytes not guranteed");
+    static_assert(alignof(T) <= 8 && "Alignment over 8 bytes not guaranteed");
     assert(bytes > 0 && "Attempted to access an uninitialised RegisterValue");
-    assert(sizeof(T) <= bytes &&
-           "Attempted to access a RegisterValue as a datatype larger than the "
-           "data held");
+    assert((sizeof(T) <= bytes || (bytes == 4 && sizeof(T) == 8)) && "Attempted"
```
rahahahat, line 101: This seems dangerous. 

Why has the assert condition been changed? Surely it doesn't make sense to assign `bytes` to be less than sizeof(T), because we can always use a smaller-sized T. 

I see however that some constructors also don't respect this condition. Please see [this](https://github.com/UoB-HPC/SimEng/pull/317/files#r1309825421).

src/include/simeng/RegisterValue.hh
rahahahat: A similar statement should be added in the constructor `  RegisterValue(const char* ptr, uint16_t bytes, uint16_t capacity)`

src/include/simeng/RegisterValue.hh
```diff
@@ -90,11 +96,16 @@ class RegisterValue {
...
+           " to access a RegisterValue as a datatype larger than the "
+           "data held" );
+    if(!relaxedFor32bit_) { // maybe #ifdef if it makes slower?
+      assert(sizeof(T) <= bytes &&
```
rahahahat, line 105: This will not work as assertions are disabled in the RELEASE build. So it will be a nop afaik.

src/include/simeng/RegisterValue.hh
rahahahat, line 101: Why has the second condition been added in the assert and why is it specific to only `bytes == 4 && sizeof(T) == 8`? Surely the first expression in the assert i.e. `sizeof(T) <= bytes` incorporates the second expression.

src/include/simeng/RegisterValue.hh
rahahahat: I don't understand why this has been added? Also, I think this is related to [ this](https://github.com/UoB-HPC/SimEng/pull/317/files#r1309922102) comment.

src/include/simeng/RegisterValue.hh
```diff
@@ -129,6 +140,9 @@ class RegisterValue {
   /** The underlying local member value. Aligned to 8 bytes to prevent
    * potential alignment issue when casting. */
   alignas(8) char value[MAX_LOCAL_BYTES];
+
+  /** Switch for different assert checking */
+  bool relaxedFor32bit_;
```
rahahahat, line 145: Could you please give additional reasoning for the addition of this new class variable and it's usage? 
Also please see comments [1](https://github.com/UoB-HPC/SimEng/pull/317/files#r1309922102) & [2](https://github.com/UoB-HPC/SimEng/pull/317/files#r1309914613) first.

src/include/simeng/RegisterValue.hh
```diff
@@ -26,10 +26,16 @@ class RegisterValue {
...
   template <class T,
             typename std::enable_if_t<!std::is_pointer_v<T>, T>* = nullptr>
-  RegisterValue(T value, uint16_t bytes = sizeof(T)) : bytes(bytes) {
+  RegisterValue(T value, uint16_t bytes = sizeof(T), bool relaxFor32 = true) : bytes(bytes) {
+    relaxedFor32bit_ = relaxFor32;
+    std::memset(this->value, 0, MAX_LOCAL_BYTES);
     if (isLocal()) {
       T* view = reinterpret_cast<T*>(this->value);
-      view[0] = value;
+      if (sizeof(T) > bytes) { // e.g. when T is int64 and bytes is 4
```
rahahahat, line 34: What would be the use case for such an addition? This will cause a truncation during reinterpret_cast. This might work for a specific use case but could cause problems in the future if something like this is allowed.

Why can't you use int32 instead?

src/include/simeng/RegisterValue.hh
rahahahat: Please[ see](https://github.com/UoB-HPC/SimEng/pull/317/files#r1309948928)

PR #332, opened 2023-07-15
Update AArch64 NEON/SVE Stores

FinnWilkinson:
All contiguous and interleaved vector stores have been updated to generate as few MemoryAccessTarget's as possible.

Some More vector loads of the same type have also been updated.

Statement added to the Developer Docs to try ensure new instructions being supported  follow the same mantra.

This also has a side effect of a large performance increase


docs/sphinx/developer/arch/supported/aarch64.rst
```diff
@@ -211,9 +211,13 @@ Loads and stores
...
+Address generation is expected to generate one or more instances of ``MemoryAccessTarget``, containing an address and the number of bytes to access. The same variables described above (``operands``, ``metadata``) are available to use to generate these addresses.
 
-Once the addresses have been generated, they should be supplied in a vector to the ``setMemoryAddresses`` helper function.
+Once the addresses have been generated, they should be supplied in a vector to the ``setMemoryAddresses`` helper function. 
+
+For loads, data can be read from the ``memoryData`` vector in ``Instruction_execute.cc``, with each index holding a ``RegisterValue`` for a corresponding ``MemoryAccessTarget``. For stores, a ``RegisterValue`` must be placed in each index of the ``memoryData`` vector, again one per ``MemoryAccessTarget`` generated.
+
+Concerning vector loads and stores (i.e. NEON, SVE, or SME), an effort should be made to create as few ``MemoryAccessTarget``'s as possible, with the exception of gather/scatter instructions where one ``MemoryAccessTarget`` should be created per vector element.
```
jj16791, line 220: I'm not sure of the best way to get my point across or a good replacement, but I think something like this would be better. "To best match modelled hardware, contiguous Load and Store instructions used one ``MemoryAccessTarget`` per destination/source register respectively whilst non-contiguous variants use multiple."

jj16791 (changes requested):

docs/sphinx/developer/arch/supported/aarch64.rst
rahahahat: Saying just one MemoryAccessTarget would be inaccurate right? given that it all depends on the predicate for SVE loads/stores? 
This is only true for NEON right? Even then we sometimes generate more than one MemoryAccessTarget, but the generated ones target a larger address range

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1631,6 +1634,44 @@ class sveHelp {
     }
     return {out, 256};
   }
+
+  /** Helper function for SVE instructions store instructions.
+   * T represents the type of operands (e.g. for zn.d, T = uint64_t).
+   * This function merges store data for contiguous active elements, and return
```
rahahahat, line 1640: returns a vector containing store data for all merged contiguous active elements.

docs/sphinx/developer/arch/supported/aarch64.rst
rahahahat: We could divide the description into two separate sentences i.e one of SVE/SME (as they are predicated) and one for NEON.
For SVE/SME you could say 

> An effort must be made to merge contiguous active elements as a single load/store by using predicate information. This has been done in helper functions `sve_merge_store_data` and `generatePredicatedContiguousAddressBlocks`. Depending upon the SVE/SME instruction to be implemented either these helpers could be used or similar logic be implemented.

rahahahat (changes requested):

docs/sphinx/developer/arch/supported/aarch64.rst
FinnWilkinson: I think a combination of the two could work. We are merging all of them to better match hardware, but a more verbose explanation can be done for SVE/SME for sure. 

I would change the SVE/SME one though to say "into a single load/store by evaluating the predicate information." And to also put in which of `Instruction_Address` and `Instruction_execute` these helper functions are used.

Also "Depending upon the SVE/SME instruction to be implemented, these helpers can be used or similar logic can be implemented if they are not suitable."

src/include/simeng/arch/aarch64/helpers/sve.hh
FinnWilkinson: Additional to this, the comment should be formatted to match all other helper functions.

jj16791 (approved):

rahahahat (approved):

PR #336, opened 2023-08-14
Various memory centric bug fixes

jj16791:
Bug fixes for:
 - Use of incorrectly speculated loaded values through forward operand calls
 - Continuation of memory accesses to the L1I interface whilst the loop buffer was in use causing  high host memory usage
 - Memory leak from SSTSimEng aggregated write request objects

### DIFF
--- a/src/lib/models/outoforder/Core.cc
+++ b/src/lib/models/outoforder/Core.cc
@@ -46,6 +46,7 @@ Core::Core(MemoryInterface& instructionMemory, MemoryInterface& dataMemory,
           [this](auto regs, auto values) {
             dispatchIssueUnit_.forwardOperands(regs, values);
           },
+          [](auto uop) { uop->setCommitReady(); },
           config["LSQ-L1-Interface"]["Exclusive"].as<bool>(),
           config["LSQ-L1-Interface"]["Load-Bandwidth"].as<uint16_t>(),
           config["LSQ-L1-Interface"]["Store-Bandwidth"].as<uint16_t>(),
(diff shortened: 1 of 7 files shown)


src/lib/models/outoforder/Core.cc
```diff
@@ -46,6 +46,7 @@ Core::Core(MemoryInterface& instructionMemory, MemoryInterface& dataMemory,
           [this](auto regs, auto values) {
             dispatchIssueUnit_.forwardOperands(regs, values);
           },
+          [](auto uop) { uop->setCommitReady(); },
```
FinnWilkinson, line 49: Why is raiseException being made to just setting commit ready?

src/lib/pipeline/LoadStoreQueue.cc
```diff
@@ -100,6 +104,13 @@ void LoadStoreQueue::startLoad(const std::shared_ptr<Instruction>& insn) {
   if (ld_addresses.size() == 0) {
     // Early execution if not addresses need to be accessed
     insn->execute();
+
+    if (insn->exceptionEncountered()) {
+      // Exception; don't pass insn to completedLoads_
+      raiseException_(insn);
+      return;
+    }
```
FinnWilkinson, line 112: This may be answered in my previous comment, but if raise exception is just setting commit ready, why do we need to use a call-back function type approach?

FinnWilkinson (changes requested):
Would like some clarification on new LSQ function please

src/lib/pipeline/LoadStoreQueue.cc
jj16791: That logic may not be appropriate for all core types using the LSQ (e.g. inorder vs OoO core)

FinnWilkinson (approved):

sst/SimEngMemInterface.cc
```diff
@@ -196,6 +196,7 @@ void SimEngMemInterface::requestWrite(const MemoryAccessTarget& target,
   for (StandardMem::Request* req : requests) {
     sstMem_->send(req);
   }
+  delete aggrReq;
```
dANW34V3R, line 199: I assume the ```new``` is required for SST to function correctly?

ABenC377 (approved):
👌🏻

rahahahat:
Looks good to me!

PR #337, opened 2023-08-14
Model accuracy improvements

jj16791:
Update to a64fx configs for improved model accuracy and application of LSQ restrictions altered to reflect updates of how contiguous addresses are generated.


configs/a64fx.yaml
```diff
@@ -245,6 +247,123 @@ Latencies:
     - STORE_ADDRESS_SVE
     Execution-Latency: 6
     Execution-Throughput: 1
+  11:
+    Instruction-Groups:
```
FinnWilkinson, line 251: Can a comment be added explaining what all these opcodes are (i.e. NEON and SVE indexed FMLA) and why they are being handled in this way as opposed to the normal way

configs/a64fx.yaml
```diff
@@ -215,7 +215,9 @@ Latencies:
   6:
     Instruction-Groups:
     - VECTOR_SIMPLE
+    - FP_MUL
     - SVE_SIMPLE
+    - SVE_MUL
```
FinnWilkinson, line 220: Why are these flags added twice? they are present 3 lines down...

configs/a64fx_SME.yaml
```diff
@@ -154,11 +137,6 @@ Reservation-Stations:
     Dispatch-Rate: 2
```
FinnWilkinson, line 137: What does dispatch rate represent? If it is how many instructions can be issued per cycle, why for RSs with only one connected EU is is more than 1?

configs/a64fx_SME.yaml
```diff
@@ -294,13 +266,130 @@ Latencies:
     Execution-Latency: 6
     Execution-Throughput: 1
   11:
+    Instruction-Groups:
+    - ~1922,
```
FinnWilkinson, line 270: Similar to my comment in `a64fx.yaml` a brief comment here would be good

configs/a64fx_SME.yaml
```diff
@@ -294,13 +266,130 @@ Latencies:
...
     - SME_SIMPLE_LOGICAL
     - SME_SIMPLE_CMP
     # Same as SVE
     Execution-Latency: 4
     Execution-Throughput: 1
-  12:
+  13:
     Instruction-Groups:
     - SME_SIMPLE
     - SME_DIV_OR_SQRT
```
FinnWilkinson, line 395: Given `-SME` has been placed above with SVE, these groups should also probably be moved to be along side the SVE counterparts above.

I think it would also be better to match how SVE groups are used - i.e. don't use the catch-all of `-SME` at all, and instead place the following in the correct latency group (match to SVE):

- SME_SIMPLE
- SME_SIMPLE_ARTH
- SME_SIMPLE_ARTH_NOSHIFT
- SME_SIMPLE_LOGICAL
- SME_SIMPLE_LOGICAL_NOSHIFT
- SME_SIMPLE_CMP
- SME_SIMPLE_CVT
- SME_MUL
- SME_DIV_OR_SQRT

src/lib/ModelConfig.cc
```diff
@@ -513,7 +513,10 @@ void ModelConfig::validate() {
           // AArch64_INSTRUCTION_LIST_END
           boundChecker(configFile_[root][i]["Instruction-Opcode"][opcodeIndex],
                        (std::string(latNum) + std::string(grpNum)),
-                       std::make_pair(0, 4516), ExpectedValue::UInteger);
+                       std::make_pair(
+                           0, static_cast<int>(
+                                  AARCH64Opcode::AArch64_INSTRUCTION_LIST_END)),
+                       ExpectedValue::UInteger);
```
FinnWilkinson, line 519: This formatting seems weird? looks like an extra line is added but I'm not sure if this is just github being weird...

src/lib/pipeline/LoadStoreQueue.cc
```diff
@@ -422,55 +422,57 @@ void LoadStoreQueue::tick() {
...
+            memory_.requestRead(req, itInsn->insn->getSequenceId());
           }
+
+          // Remove processed address from queue
+          addressQueue.pop();
+        }
+        // Remove entry from vector iff all of its requests have been
+        // scheduled
+        if (addressQueue.size() == 0) {
+          itInsn = itReq->second.erase(itInsn);
```
FinnWilkinson, line 475: I'm not 100% sure on why this needed changing? Are you able to briefly explain why? Thanks

FinnWilkinson (changes requested):

configs/a64fx.yaml
rahahahat: +1

src/lib/pipeline/LoadStoreQueue.cc
```diff
@@ -422,55 +422,57 @@ void LoadStoreQueue::tick() {
...
+        auto& addressQueue = itInsn->reqAddresses;
+        while (addressQueue.size()) {
+          const simeng::MemoryAccessTarget req =
+              addressQueue.front();  // Speculatively increment count of this
+                                     // request type
+          reqCounts[isStore]++;
+
+          // Ensure the limit on the number of permitted operations is adhered
+          // to
+          if (reqCounts[isStore] + reqCounts[!isStore] > totalLimit_) {
```
rahahahat, line 436: Why don't we decrement reqCounts[isStore] in this if statement, if we do it in the next one?

configs/a64fx.yaml
jj16791: A merge error

configs/a64fx_SME.yaml
jj16791: It's what you said. Won't effect anything due to the notion of ports being busy but I'll change it for clarity

configs/a64fx_SME.yaml
jj16791: So are you saying to duplicate all SVE group latency settings with an SME entry? All this config stuff was just copied over from the ModSim work so it should be what you defined previously.

src/lib/ModelConfig.cc
jj16791: This si the way clang formatting formats it

src/lib/pipeline/LoadStoreQueue.cc
jj16791: This change was done so that the limits were applied to individual address accesses rather than individual uops. We could have a case whereby a uop with 8 accesses (8 separate memory flows) would be considered as a single access from the LSQ restrictions pov.

src/lib/pipeline/LoadStoreQueue.cc
jj16791: This case will cause us to stop considering any accesses whilst the second condition is for only 1 type of access (i.e. Load or Store). The decrement is meaningless in the former case but we have to decrement the speculative increment in the second case to avoid the first case from being incorrectly evaluated as true.

For example:

Allow for 2 LD/1 ST/2 accesses overall per cycle.

2 STs go through, second one is prevented due to limits applied (without decrement the `exceededLimits[accessType::STORE]` would equal 2).
A LD should be able to go through as only 1 access has previously (the first ST) but with the decrement `reqCounts[isStore] + reqCounts[!isStore] > totalLimit_` would evaluate to true which is incorrect.

We could have a decrement in the first case but it would serve no purpose hence why it was omitted. I don't think its inclusion provides any further code clarity especially given a comment outlining my above explanation is given in the second condition body above the `reqCounts[isStore]--;`.

configs/a64fx.yaml
jj16791: Added a brief comment

configs/a64fx_SME.yaml
jj16791: Added a brief comment

configs/a64fx_SME.yaml
FinnWilkinson: Yes. For this example implementation of SME, all latencies are identical to their SVE counterparts. I should have done this before in the ModSim work - but have only spotted it now on review

configs/a64fx.yaml
FinnWilkinson: Sorry to add to this. Could a `/** NOTE: */` or similar be added too saying that any changes to Capstone will likely break this functionality (as Opcode values will change)

src/lib/pipeline/LoadStoreQueue.cc
FinnWilkinson: I see, so for instructions like an FP32 Gather, previously it would treat this as 1 access rather than the 8 individual accesses that it should do (assuming 8 elements per FP32 vector)?

And just to clarify - a normal SVE vector load (assuming `ptrue p0`) will still be done as a single access / flow?

configs/a64fx_SME.yaml
jj16791: Made the change, please double check in case I've misunderstood

src/lib/pipeline/LoadStoreQueue.cc
jj16791: Yes exactly. The restrictions were being placed on the entries in `requestLoadQueue_` or `requestStoreQueue_` rather than the addresses in each entry. Given our updates to the SVE contiguous accesses, they only have 1 address in each of those request queue entries so yes, 1 flow

configs/a64fx.yaml
dANW34V3R: I think the Capstone enum "names" can be used instead of the raw numbers  which <em>may</em> prevent breakages if Capstone changes. I think I successfully tried this a while ago but I could be wrong.

dANW34V3R: https://uob-hpc.github.io/SimEng/developer/arch/supported/riscv.html#instruction-groups

See bottom paragraph of this section

src/lib/pipeline/LoadStoreQueue.cc
```diff
@@ -422,55 +422,57 @@ void LoadStoreQueue::tick() {
...
+          // Request a read from the memory interface if the requestQueue_
+          // entry represents a read
+          if (!isStore) {
+            memory_.requestRead(req, itInsn->insn->getSequenceId());
           }
+
+          // Remove processed address from queue
+          addressQueue.pop();
+        }
+        // Remove entry from vector iff all of its requests have been
```
dANW34V3R, line 472: Is the second "f" in "iff" intentional to mean "if and only if"?

jj16791: I'll remove the second f, it makes sense but isn't necessary

dANW34V3R (approved):

configs/a64fx.yaml
FinnWilkinson: Given this is in the YAML file and not in the code itself, we can't use the enum name as (I'm fairly sure) you can't convert a string to an enum name quickly. Resolving as initial issues have been amended

configs/a64fx_SME.yaml
FinnWilkinson: Looks good. Thanks for updating

ABenC377 (approved):
👍🏻

src/lib/pipeline/LoadStoreQueue.cc
rahahahat: Great explanation.  To prevent this question from being asked again, can you please add some extra detail to the comment? i.e. Per cycle limits exceeded, no more requests can be scheduled this cycle.

rahahahat (approved):

src/lib/pipeline/LoadStoreQueue.cc
rahahahat: Or something like Per cycle limits exceeded, more request can be scheduled next cycle.

jj16791: As stated, there's already a comment for the reverse of this explanation in the second conditional body hence adding another comment is overly verbose and makes the code look messier.

Issue #329, opened 2023-07-06
Incorrect logic on multiple SMSTART calls


FinnWilkinson:
As part of the SME AArch64 extension, Streaming SVE Mode was introduced. This is enabled via `SMSTART` and disabled by `SMSTOP`. For any SME instruction to be executed the core must be in Streaming SVE mode. 

`SMSTART`/`SMSTOP` is composed of 2 parts: `SM` and `ZA` (not to be confused with the register `za`). When `SM` goes from an enabled state (=1) to a disabled state (=0) or vice versa then all Vector (`z`, and subsequently `v`, `b`, `h`, `s`, `d` regs) and Predicate (`p`) registers are set to 0. Similarly, when `ZA` goes from an enabled state (=1) to a disabled state (=0) or vice versa then the `za` matrix register is zeroed out.

Currently within SimEng, on any execution of `SMSTART` or `SMSTOP` the register clearing functionality is performed for SM or ZA or both; which ever is appropriate for the instruction variation executed. However, this zeroing functionality should only be performed when the state changes from 1 to 0 or 0 to 1, not on any call.

For example, the trace below is what currently would happen in SimEng

```
SMSTART                    // Enter Streaming SVE mode. Zero all vector, fp, scalar, neon, predicate, and za matrix registers
dup z0.s, #3               // Duplicate the number 3 in all elements of z0 :  `z0.s = {3, 3, 3, 3}`
dup z1.s, #2               // Duplicate the number 2 in all elements of z1 : `z1.s = {2, 2, 2, 2}`
SMSTART                  // Enter Streaming SVE mode, zero all vector, fp, scalar, neon, predicate, and za matrix registers
add z2.s, z1.s, z0.s   // Add z0.s and z1.s element wise : `z2.s = {0, 0, 0, 0}`
SMSTOP                    // Exit Streaming mode. Zero all vector, fp, scalar, neon, predicate, and za matrix registers
```

However, the following should happen instead - effectively making the second SMSTART have no affect:
```
SMSTART                    // Enter Streaming SVE mode. Zero all vector, fp, scalar, neon, predicate, and za matrix registers
dup z0.s, #3               // Duplicate the number 3 in all elements of z0 :  `z0.s = {3, 3, 3, 3}`
dup z1.s, #2               // Duplicate the number 2 in all elements of z1 : `z1.s = {2, 2, 2, 2}`
SMSTART                  // Already in Streaming SVE mode. Do nothing
[...]

PR #339, opened 2023-10-30
SME SVCR logic fix

FinnWilkinson:
This PR for the most part addresses issue #329 - correcting the SVCR and Streaming Mode logic introduced with SME. This is summaried as:

-  If a program is in Streaming Mode and calls SMSTART again, ZA and Vector registers are **not** zeroed out. 
- SVE Vector registers (and their aliases such as NEON and Scalar registers) are only zeroed out when SVCR.SM bit changes
- SME's ZA register is only zeroed out when SVCR.ZA bit changes
- SVCR can now be correctly updated using a regulat `msr svcr, xt` instruction in addition to `smstart`/`smstop`/`msr svcr, #imm`
- More tests surrounding the SVCR logic have been added

Additional to this, logic for the NEON muti-struct store with post index instructions introduced in PR #335 has been corrected via a small change to the metadata to correctly identify if a `#imm` or `xt` is used as said post index.

NOTE: this branch is based on `macos-sst-build-fix` and thus contains the fixes present in PR #338. This PR will be rebased to `dev` once PR #338 has been merged and should be merged into `dev` after PR #338.

### DIFF
--- a/src/lib/arch/aarch64/ExceptionHandler.cc
+++ b/src/lib/arch/aarch64/ExceptionHandler.cc
@@ -649,16 +649,15 @@ bool ExceptionHandler::init() {
   } else if (exception == InstructionException::StreamingModeUpdate ||
              exception == InstructionException::ZAregisterStatusUpdate ||
              exception == InstructionException::SMZAUpdate) {
-    // Retrieve register file structure from architecture
-    auto regFileStruct =
-        instruction_.getArchitecture().getRegisterFileStructures();
-    // Retrieve metadata from architecture
-    auto metadata = instruction_.getMetadata();
     // Get Architecture
     const Architecture& arch = instruction_.getArchitecture();
+    // Retrieve register file structure from architecture
+    auto regFileStruct = arch.getRegisterFileStructures();
+    // Retrieve metadata from architecture
+    auto metadata = instruction_.getMetadata();
 
     uint64_t newSVCR = 0;
-    const uint64_t currSVCR = instruction_.getArchitecture().getSVCRval();
+    const uint64_t currSVCR = arch.getSVCRval();
 
     // Check if exception was called by AArch64_MSR (msr systemreg, xt) or
     // AArch64_MSRpstatesvcrImm1 (msr svcr<sm|za|smza>, #imm)
--- a/src/lib/arch/aarch64/ExceptionHandler.cc
+++ b/src/lib/arch/aarch64/ExceptionHandler.cc
@@ -654,52 +654,71 @@ bool ExceptionHandler::init() {
         instruction_.getArchitecture().getRegisterFileStructures();
     // Retrieve metadata from architecture
     auto metadata = instruction_.getMetadata();
+    // Get Architecture
+    const Architecture& arch = instruction_.getArchitecture();
 
-    // Update SVCR value
-    const uint64_t svcrBits = static_cast<uint64_t>(metadata.operands[0].svcr);
-    const uint8_t imm = metadata.operands[1].imm;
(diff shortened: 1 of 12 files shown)


JosephMoore25:
Looks mostly good, but tiny code inconsistency.

You define "const Architecture& arch = instruction_.getArchitecture();" but then still use instruction_.getArchitecture() a couple lines later and then switch to arch a bit later. Probs best to just stick to one or the other for the sake of consistency.

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -654,52 +654,71 @@ bool ExceptionHandler::init() {
...
     auto metadata = instruction_.getMetadata();
+    // Get Architecture
+    const Architecture& arch = instruction_.getArchitecture();
 
-    // Update SVCR value
-    const uint64_t svcrBits = static_cast<uint64_t>(metadata.operands[0].svcr);
-    const uint8_t imm = metadata.operands[1].imm;
-    const uint64_t currSVCR = instruction_.getArchitecture().getSVCRval();
     uint64_t newSVCR = 0;
+    const uint64_t currSVCR = instruction_.getArchitecture().getSVCRval();
```
ABenC377, line 661: can use arch variable set above

test/regression/aarch64/Exception.cc
```diff
@@ -76,6 +119,13 @@ TEST_P(Exception, svcr) {
   )");
   CHECK_NEON(0, uint64_t, fillNeon<uint64_t>({0}, VL / 8));
 
+  RUN_AARCH64(R"(
+    # Ensure z regs get enabled when SM enabled
+    smstart sm
+    dup z0.d, #3
```
ABenC377, line 125: Lines 124 and 125 are the other way around than the surrounding tests (i.e., smstart before dup).   Is this intended?

ABenC377 (approved):
Can 'instruction_.getArchitecture()' in ExceptionHandler.cc line 661 be replaced with the 'arch' variable declared above it?

FinnWilkinson:
@JosephMoore25 @ABenC377 Good spot, will change this now

JosephMoore25 (approved):

src/lib/arch/aarch64/Instruction.cc
```diff
@@ -36,6 +36,11 @@ InstructionException Instruction::getException() const { return exception_; }
 const span<Register> Instruction::getOperandRegisters() const {
   return {const_cast<Register*>(sourceRegisters.data()), sourceRegisterCount};
 }
+
+const span<RegisterValue> Instruction::getSourceOperands() const {
+  return {const_cast<RegisterValue*>(operands.data()), operands.size()};
+}
+
```
dANW34V3R, line 43: Any specific reason to be returning a span rather than a vector as in the AArch64/Instruction.hh implementation of operands, possibly preventing this awkward conversion. The RISC-V implementation has operands as an array so conversion would be necessary either way

dANW34V3R (commented):
One comment

src/lib/arch/aarch64/Instruction.cc
FinnWilkinson: Using span as a return type was copied from existing code in the class (see `getOperandRegisters()`). 

In terms of why a span - if we return a vector then we create a new copy of the data, whilst a span (made up of a pointer and length) will just point to said data.

dANW34V3R (approved):

ABenC377 (approved):

JosephMoore25 (approved):

jj16791 (approved):

Issue #322, opened 2023-05-28
Use raw string literal for `DEFAULT_CONFIG` in  ModelConfig.hh 


tom91136:
Let's replace the following macro:

```c++
#define DEFAULT_CONFIG                                                         \
  ("{Core: {ISA: AArch64, Simulation-Mode: inorderpipelined, "                 \
   "Clock-Frequency: 2.5, Timer-Frequency: 100, Micro-Operations: True, "      \
   "Vector-Length: 512, Streaming-Vector-Length: 512}, Fetch: "                \
   "{Fetch-Block-Size: 32, Loop-Buffer-Size: 64, Loop-Detection-Threshold: "   \
   "4}, Process-Image: {Heap-Size: 10485760, Stack-Size: 1048576}, "           \
   "Register-Set: {GeneralPurpose-Count: 154, FloatingPoint/SVE-Count: 90, "   \
   "Predicate-Count: 17, Conditional-Count: 128, Matrix-Count: 2}, "           \
   "Pipeline-Widths: {Commit: 4, FrontEnd: 4, LSQ-Completion: 2}, "            \
   "Queue-Sizes: {ROB: 180, Load: 64, Store: 36}, Branch-Predictor: "          \
   "{BTB-Tag-Bits: 11, Saturating-Count-Bits: 2, Global-History-Length: 10, "  \
   "RAS-entries: 5, Fallback-Static-Predictor: 2}, L1-Data-Memory: "           \
   "{Interface-Type: Flat}, L1-Instruction-Memory: {Interface-Type: Flat}, "   \
   "LSQ-L1-Interface: {Access-Latency: 4, Exclusive: False, Load-Bandwidth: "  \
   "32, Store-Bandwidth: 16, Permitted-Requests-Per-Cycle: 2, "                \
   "Permitted-Loads-Per-Cycle: 2, Permitted-Stores-Per-Cycle: 1}, Ports: "     \
   "{'0': {Portname: Port 0, Instruction-Group-Support: [1, 8, 14]}, '1': "    \
   "{Portname: Port 1, Instruction-Group-Support: [0, 14]}, '2': {Portname: "  \
   "Port 2, Instruction-Group-Support: [1, 8, 71]}, '3': {Portname: Port 4, "  \
   "Instruction-Group-Support: [67]}, '4': {Portname: Port 5, "                \
   "Instruction-Group-Support: [67]}, '5': {Portname: Port 3, "                \
[...]

PR #343, opened 2023-11-03
Global siminfo

jj16791:
This PR looks to replace the use of Config.hh and YAML-CPP with SimInfo.hh and a single header yaml library ryml. The SimInfo class acts as a central point for all model configuration values both defined manually and from a read-in yaml config file.

Improvements to adding new config options have been included through a new expectation struct that places requirements on a specific config option and a default value. Through this, default config files can be created and are especially useful in the test suite.

An aside, there seem to be a lot of line changes but ~33K of those are the new header-only ryml library


FinnWilkinson:
Does this PR invalidate issue #322 ?

src/include/simeng/config/SimInfo.hh
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  /** A utility function to get a value, of a specified type, from a config
+   * option. */
+  template <typename T>
+  static T getValue(ryml::ConstNodeRef node) {
+    T val;
+    node >> val;
+    return val;
+  }
```
FinnWilkinson, line 67: I may be missing somethig here, but given we store `validatedConfig` in SimInfo anyway, could we not have the argument of `getValue` be some string or struct of strings rather than require all classes that want to access a config value to either store a copy of the config or be passed one as an argument?

For example, have `config::SimInfo::getValue<bool>("LSQ-L1-Interface/Exclusive")` or `config::SimInfo::getValue<bool>({"LSQ-L1-Interface", "Exclusive"})`?

This would reduce the number of copies of the config tree needing to be passed around the project, and allow all classes to automatically have access to the config file via `SimInfo.hh`.

FinnWilkinson: This could also help with another issue [raised previously](https://github.com/UoB-HPC/SimEng/pull/307#discussion_r1200792935); as forward declaring this single function along with `class SimInfo` should remove any circular dependancy from an implementation using `archInfo.hh` files in each ISAs `arch` directory

jj16791: I had thought of this prior but hit many dead ends. Given that you can have any depth regarding the YAML hierarchy, I'm unsure of a good way to do this. 

The string-only param has the problem of excluding certain characters from YAML config options.

We could do a vector of strings instead but it doesn't seem user-friendly. However, there's an argument that the current solution also has an issue of "user-friendlyness".

jj16791:
> Does this PR invalidate issue #322 ?

The direct define mention yes but there are some new defines which can benefit from the suggested change

src/include/simeng/config/SimInfo.hh
jj16791: I guess it's a question of 
```
config::SimInfo::getValue<int>(config["Pipeline-Widths"]["FrontEnd"]
```
vs 
```
config::SimInfo::getValue<int>({"Pipeline-Widths", "FrontEnd"})
```

jj16791: One small other point would be that without the current method, you'd have to convert non-string values to strings before passing them as part of the param. E.g. having to convert the iteration count of a for loop from an int to a string.

Possible a minor detail but all this is for the sake of readability

FinnWilkinson: I would argue that a vector of strings is no less user friendly than the current setup - you still need to know the exact names of the fields anyway; you're just encasing it in a vector rather than in `[ ]`. Plus, in my opinion, not having to pass `config` to every class that would need information from the config file is a big plus, and also eliminates n copies of the same object being passed, created, or stored in the project as currently is the case.

I'm also not 100% sure on your final point about converting things to strings... wouldn't this be the case currently too? As field names are passed always as `["field name here"]`. Please correct me if this isn't the case though!

jj16791: I agree with the moot user-friendly point hence my previous comment
> However, there's an argument that the current solution also has an issue of "user-friendlyness".

Also, we're passing the config tree around as a const reference so not sure it has the detrimental impact of n actual copies.

Further to the n copies point, if you wanted to truly eliminate all copies then you'd have to replace all accesses to the config which don't use the `config::SimInfo::getValue` utility which were removed in the previous PR due to readability. For example,

```
 ryml::ConstNodeRef config = config::SimInfo::getConfig();
 config["Branch-Predictor"]["BTB-Tag-Bits"] >> btbBits_;
 config["Branch-Predictor"]["Saturating-Count-Bits"] >> satCntBits_;
 config["Branch-Predictor"]["Global-History-Length"] >> globalHistoryLength_;
 config["Branch-Predictor"]["RAS-entries"] >> rasSize_;
```
would become,
```
btbBits_ = config::SimInfo::getValue<uint64_t>({"Branch-Predictor", "BTB-Tag-Bits"});
satCntBits_ = config::SimInfo::getValue<uint8_t>({"Branch-Predictor", "Saturating-Count-Bits"});
globalHistoryLength_ = config::SimInfo::getValue<uint16_t>({"Branch-Predictor", "Global-History-Length"});
rasSize_ = config::SimInfo::getValue<uint64_t>({"Branch-Predictor", "RAS-entries"});
```
which gets worse when you consider each access has to be defined from the root of the YAML tree (e.g. you can't define the point of access as `ryml::ConstNodeRef port_node = config["Latencies"][i];` and then extract from there).

For the last point. Let's say you're doing something in a loop with iteration variable `i`, then we'd have:
```
config::SimInfo::getValue<bool>(config["Execution-Units"][i]["Pipelined"])
```
vs
```
config::SimInfo::getValue<bool>({"Execution-Units", std::to_string(i), "Pipelined"})
```

configs/a64fx_SME.yaml
```diff
@@ -1,10 +1,8 @@
----
 # The following resources where utilised to create the config file and naming schemes:
 # https://github.com/fujitsu/A64FX
 
 Core:
   Simulation-Mode: outoforder
-  ISA: AArch64
```
ABenC377, line 7: Why is the ISA parameter being deleted here but not in the other .yaml files?

src/include/simeng/GenericPredictor.hh
```diff
@@ -46,7 +46,7 @@ class GenericPredictor : public BranchPredictor {
 
   /** A 2^bits length vector of pairs containing a satCntBits_-bit saturating
    * counter and a branch target. */
-  std::vector<std::pair<uint8_t, uint64_t>> btb_;
+  std::vector<std::pair<uint16_t, uint64_t>> btb_;
```
ABenC377, line 49: why do we need 16 bits for a saturating counter?  Should only use two or three.

src/include/simeng/arch/Architecture.hh
```diff
@@ -95,16 +90,14 @@ class Architecture {
...
   virtual uint8_t getMaxInstructionSize() const = 0;
 
-  /** Returns the physical register structure as defined within the config
-   * file
+  /** Returns the physical register structure as defined within the config file
    */
-  virtual std::vector<RegisterFileStructure> getConfigPhysicalRegisterStructure(
-      YAML::Node config) const = 0;
+  virtual std::vector<RegisterFileStructure>
+  getConfigPhysicalRegisterStructure() const = 0;
```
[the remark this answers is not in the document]
FinnWilkinson: We could possibly store the physical register file structure in SimInfo too (along with the quantaties as a seperate structure if needed)?

src/include/simeng/config/SimInfo.hh
```diff
@@ -0,0 +1,235 @@
...
+#include "simeng/config/ModelConfig.hh"
+#include "simeng/config/yaml/ryml.hh"
+
+namespace simeng {
+namespace config {
+
+#define DEFAULT_STR "Default"
+
+/** Enum representing the possible simulation modes. */
+enum class simMode { emulation, inorder, outoforder };
```
FinnWilkinson, line 19: If this (along with the enum class for ISA type) was defined in the `simeng` namespace rather than the `simeng::config` namespace then checking against these values would be much cleaner. i.e. `... == simMode::emulation` rather than `config::simMode::emulation`

FinnWilkinson: Also, having this enum class named `SimMode` and having each enum begin with a capital letter (i.e. `Emulation`) would be more consistent with the rest of the code base

src/lib/CoreInstance.cc
```diff
@@ -280,8 +266,10 @@ void CoreInstance::createCore() {
 
 void CoreInstance::createSpecialFileDirectory() {
   // Create the Special Files directory if indicated to do so in Config
-  if (config_["CPU-Info"]["Generate-Special-Dir"].as<bool>() == true) {
-    simeng::SpecialFileDirGen SFdir = simeng::SpecialFileDirGen(config_);
+  bool generate;
+  config_["CPU-Info"]["Generate-Special-Dir"] >> generate;
```
FinnWilkinson, line 270: Can' this be replaced with `config::SimInfo::getGenSpecFiles()`?

src/lib/CoreInstance.cc
```diff
@@ -280,8 +266,10 @@ void CoreInstance::createCore() {
...
+  if (generate) {
+    simeng::SpecialFileDirGen SFdir = simeng::SpecialFileDirGen();
```
FinnWilkinson, line 272: is the `simeng::` needed given this is also in the simeng namespace?

src/lib/CoreInstance.cc
```diff
@@ -33,16 +33,15 @@ CoreInstance::~CoreInstance() {
...
   // Check to see if either of the instruction or data memory interfaces should
   // be created. Don't create the core if either interface is marked as External
   // as they must be set manually prior to the core's creation.
 
   // Convert Data-Memory's Interface-Type value from a string to
   // simeng::MemInterfaceType
-  std::string dType_string =
-      config_["L1-Data-Memory"]["Interface-Type"].as<std::string>();
+  std::string dType_string;
+  config_["L1-Data-Memory"]["Interface-Type"] >> dType_string;
```
FinnWilkinson, line 44: The use of SimInfo's `getValue` function seems vacant in this file. Given it exists we should use it to avoid these multi-line value accesses. Apply throughout

src/lib/SpecialFileDirGen.cc
```diff
@@ -4,19 +4,20 @@
...
+  config["CPU-Info"]["Socket-Count"] >> socket_count;
+  config["CPU-Info"]["SMT"] >> smt;
+  config["CPU-Info"]["BogoMIPS"] >> bogoMIPS;
+  config["CPU-Info"]["Features"] >> features;
+  config["CPU-Info"]["CPU-Implementer"] >> cpu_implementer;
+  config["CPU-Info"]["CPU-Architecture"] >> cpu_architecture;
+  config["CPU-Info"]["CPU-Variant"] >> cpu_variant;
+  config["CPU-Info"]["CPU-Part"] >> cpu_part;
+  config["CPU-Info"]["CPU-Revision"] >> cpu_revision;
+  config["CPU-Info"]["Package-Count"] >> package_count;
```
FinnWilkinson, line 20: No use of SimInfo's `getValue` function which allows for a more conventional way of setting values (more readable to do `var_a = SimInfo::getValue<int>(config[...]);`

FinnWilkinson: This is a re-occuring issue throughout - please apply this throughout the project

src/lib/arch/aarch64/Architecture.cc
```diff
@@ -26,15 +22,23 @@ Architecture::Architecture(kernel::Linux& kernel, YAML::Node config)
...
-  systemRegisterMap_[ARM64_SYSREG_FPSR] = systemRegisterMap_.size();
-  systemRegisterMap_[ARM64_SYSREG_TPIDR_EL0] = systemRegisterMap_.size();
-  systemRegisterMap_[ARM64_SYSREG_MIDR_EL1] = systemRegisterMap_.size();
-  systemRegisterMap_[ARM64_SYSREG_CNTVCT_EL0] = systemRegisterMap_.size();
-  systemRegisterMap_[ARM64_SYSREG_PMCCNTR_EL0] = systemRegisterMap_.size();
-  systemRegisterMap_[ARM64_SYSREG_SVCR] = systemRegisterMap_.size();
+  std::vector<uint64_t> sysRegs = config::SimInfo::getSysRegVec();
+  for (size_t i = 0; i < sysRegs.size(); i++) {
+    systemRegisterMap_[sysRegs[i]] = systemRegisterMap_.size();
+  }
```
FinnWilkinson, line 41: What is the need for this map given `getSysRegVecIndex()` in SimInfo?

src/lib/arch/aarch64/Architecture.cc
```diff
@@ -297,36 +296,50 @@ void Architecture::updateSystemTimerRegisters(RegisterFileSet* regFile,
 }
 
 std::vector<RegisterFileStructure>
-Architecture::getConfigPhysicalRegisterStructure(YAML::Node config) const {
+Architecture::getConfigPhysicalRegisterStructure() const {
+  // Get register counts
+  ryml::ConstNodeRef regConfig = config::SimInfo::getConfig()["Register-Set"];
+  uint16_t gpCnt;
```
FinnWilkinson, line 302: Use of `Count` rather than `Cnt` is preferable in my opinion. Won't die on this hill though so change at your discression

src/lib/arch/aarch64/MicroDecoder.cc
```diff
@@ -10,8 +10,9 @@ std::unordered_map<uint32_t, std::vector<Instruction>>
     MicroDecoder::microDecodeCache;
 std::forward_list<InstructionMetadata> MicroDecoder::microMetadataCache;
 
-MicroDecoder::MicroDecoder(YAML::Node config)
-    : instructionSplit_(config["Core"]["Micro-Operations"].as<bool>()) {}
+MicroDecoder::MicroDecoder() {
+  config::SimInfo::getConfig()["Core"]["Micro-Operations"] >> instructionSplit_;
+}
```
FinnWilkinson, line 15: use of `getValue` would allow `instructionSplit_` to be defined in the initialiser list again

src/lib/arch/riscv/Architecture.cc
```diff
@@ -26,19 +27,29 @@ Architecture::Architecture(kernel::Linux& kernel, YAML::Node config)
 
   cs_option(capstoneHandle, CS_OPT_DETAIL, CS_OPT_ON);
 
+  // Generate zero-indexed system register map
+  for (size_t i = 0; i < config::SimInfo::getSysRegVec().size(); i++) {
+    systemRegisterMap_[config::SimInfo::getSysRegVec()[i]] =
+        systemRegisterMap_.size();
+  }
```
FinnWilkinson, line 34: Similar comments made in AArch64 Architecture.cc apply to this file also

src/lib/arch/riscv/InstructionMetadata.hh
```diff
@@ -2,7 +2,8 @@
 
 #include <string>
 
-#include "capstone/capstone.h"
+#include "simeng/arch/riscv/Architecture.hh"
+#include "simeng/arch/riscv/Instruction.hh"
```
FinnWilkinson, line 6: Why are these headers now required?

src/include/simeng/config/SimInfo.hh
FinnWilkinson: In a similar vein as comment 1, removing SimInfo from the `config` namespace (it can still be held in the config directory) will make accessing SimInfo more readable & intuative

src/lib/config/ModelConfig.cc
```diff
@@ -0,0 +1,997 @@
...
+  // validate it to ensure correctness for the simulation
+  setExpectations(true);
+  constructDefault(expectations_, configTree_.root_id());
+  validate();
+}
+
+void ModelConfig::constructDefault(ExpectationNode expectations,
+                                   size_t root_id) {
+  // Iterate over the expectations supplied
+  for (const auto& chld : expectations.getChildren()) {
```
FinnWilkinson, line 103: Personal preference for `child` over `chld` but not a required change to be made.

If change is made, please apply throughout file

src/lib/config/ModelConfig.cc
```diff
@@ -0,0 +1,997 @@
...
+
+  expectations_["Core"].addChild(
+      ExpectationNode::createExpectation<std::string>("emulation",
+                                                      "Simulation-Mode"));
+  expectations_["Core"]["Simulation-Mode"].setValueSet(
+      std::vector<std::string>{"emulation", "inorderpipelined", "outoforder"});
+
+  expectations_["Core"].addChild(
+      ExpectationNode::createExpectation<float>(1.f, "Clock-Frequency"));
+  expectations_["Core"]["Clock-Frequency"].setValueBounds(0.f, 10.f);
```
FinnWilkinson, line 250: Proposing a change to `Clock-Frequency-GHz` for additional clarity

src/lib/config/ModelConfig.cc
```diff
@@ -0,0 +1,997 @@
...
+
+  // Early check on ["Core"]["Clock-Frequency"] as values are needed to inform
+  // the expected lower bound of the ["Core"]["Timer-Frequency"] value
+  uint64_t tFreqUpperBound = 1000;
```
FinnWilkinson, line 254: Linking the `tFreqUpperBound` and upper bound of clock frequency via a single variable would make updating this ceiling much easier if required, and less prone to errors

src/lib/config/ModelConfig.cc
```diff
@@ -0,0 +1,997 @@
...
+                << std::endl;
+      exit(1);
+    }
+
+    tFreqUpperBound = cFreq * 1000;
+  }
+
+  expectations_["Core"].addChild(
+      ExpectationNode::createExpectation<uint64_t>(100, "Timer-Frequency"));
+  expectations_["Core"]["Timer-Frequency"].setValueBounds<uint64_t>(
```
FinnWilkinson, line 275: Proposing a change to `Timer-Frequency-MHz` for additional clarity?

src/lib/config/ModelConfig.cc
```diff
@@ -0,0 +1,997 @@
...
+        512, "Vector-Length", true));
+    expectations_["Core"]["Vector-Length"].setValueSet(
+        std::vector<uint64_t>{128, 256, 384, 512, 640, 768, 896, 1024, 1152,
+                              1280, 1408, 1536, 1664, 1792, 1920, 2048});
+
+    expectations_["Core"].addChild(ExpectationNode::createExpectation<uint64_t>(
+        512, "Streaming-Vector-Length", true));
+    expectations_["Core"]["Streaming-Vector-Length"].setValueSet(
+        std::vector<uint64_t>{128, 256, 384, 512, 1024, 2048});
+  }
```
FinnWilkinson, line 294: I feel as though the default values of VL and SVL should be 128, as it makes more sense to go with the lower bound rather than a64fx's value of VL. (Won't die on this hill...)

src/lib/config/ModelConfig.cc
```diff
@@ -0,0 +1,997 @@
...
+
+  // Fetch
+  expectations_.addChild(ExpectationNode::createExpectation("Fetch"));
+
+  expectations_["Fetch"].addChild(
+      ExpectationNode::createExpectation<uint64_t>(32, "Fetch-Block-Size"));
+  expectations_["Fetch"]["Fetch-Block-Size"].setValueSet(std::vector<uint64_t>{
```
FinnWilkinson, line 301: Fetch-block-size could be `uint32_t` given the max value is 65536

src/lib/config/ModelConfig.cc
```diff
@@ -0,0 +1,997 @@
...
+  // Iterate over the expectations supplied
+  for (const auto& chld : expectations.getChildren()) {
+    std::string key = chld.getKey();
+    ExpectedType type = chld.getType();
+    // If the key is a wildcard , then change it to be an appropriate value
+    // in the resultant config file and its type to be valueless
+    if (key == wildcard) {
+      key = "0";
+      type = ExpectedType::Valueless;
+    }
```
FinnWilkinson, line 111: What is the reasoning behind changing the key from `*` to `0`?

FinnWilkinson (changes requested):

src/include/simeng/config/SimInfo.hh
FinnWilkinson: Just realised in my full review that we pass around as reference, so yup I agree that is moot.

Given what you've said, I agree keeping it as is would be worthwhile, however, I do think we should use `SimInfo::getValue<uint64_t>` etc as much as possible mainly as I think `var = SimInfo::getValue<uint64_t>()` is more intuative than `config[...][...] >> var`

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+  Double,
+  Float,
+  Integer,
+  String,
+  UInteger,
+  Valueless
+};
+
+/** String to represent a wildcard node key. */
+const std::string wildcard = "*";
```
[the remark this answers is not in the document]
jj16791: There's a comment in the class itself above `bool isWild_` denoting what a while node is. I'll look to add a comment somewhere in the code explaining what a wildcard node is and then reference it in all "wildcard" associated comments. No point explaining it in multiple places and looks messy.

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+  /** A templated struct to store a boolean value denoting whether a passed
+   * typename T belongs to one types represented in ExpectedType. */
+  template <typename T>
+  struct is_expected_type
+      : std::integral_constant<bool, std::is_same<bool, T>::value ||
+                                         std::is_same<double, T>::value ||
+                                         std::is_same<float, T>::value ||
+                                         std::is_same<int64_t, T>::value ||
+                                         std::is_same<std::string, T>::value ||
+                                         std::is_same<uint64_t, T>::value> {};
```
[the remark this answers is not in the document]
jj16791: Both were kept inside the ExpectationNode class as that's the only scope they're used in. Unlike ExpectedType, wildcard, and ValidationResult which are needed by other classes of the config namespace.

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
FinnWilkinson: Thats fair - I will admit that this comment was made before looking through the whole class (I should have checked) and if the `isWild` comment is good then I agree only 1 point of explanation is needed

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
FinnWilkinson: Understood. If their scope is only within the class then perhaps they should be declared under `private` or `protected`?

src/include/simeng/arch/Architecture.hh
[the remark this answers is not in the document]
jj16791: Moved all the physical register information to the ArchInfo files

src/include/simeng/arch/aarch64/ArchInfo.hh
```diff
@@ -0,0 +1,58 @@
...
+    return sysRegisterEnums_;
+  }
+
+  /** Get the structure of the architecture register fileset(s). */
+  std::vector<simeng::RegisterFileStructure> getArchRegStruct(
+      ryml::ConstNodeRef config) override {
+    // Given some register quantities rely on Config file arguments (SME
+    // relies on SVL), it is possible that if the config was to change the
+    // register quantities would be incorrect. This function provides a way to
+    // reset the Architectural register structure.
```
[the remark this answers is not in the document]
jj16791: Comment no longer exists

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+ * config option. */
+enum class ExpectedType {
+  Bool,
+  Double,
+  Float,
+  Integer,
+  String,
+  UInteger,
+  Valueless
+};
```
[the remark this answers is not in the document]
jj16791: Added to comment above DataTypeVariant as ExpectedType doesn't depend on DataTypeVariant unlike DataTypeVariant depending on ExpectedType

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+   * instance. The instance created is one with a value and a key. A default
+   * value, key, type, and a bool denoting whether the node is optional are
+   * provided to the underlying constructor. */
+  template <typename T>
+  static ExpectationNode createExpectation(T defaultValue, std::string key,
+                                           bool optional = false) {
+    // Ensure templated type is of an expected type
+    static_assert(is_expected_type<T>::value &&
+                  "[SimEng:ModelConfig] Unexpected type given to "
+                  "ExpectationNode::create");
```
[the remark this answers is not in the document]
jj16791: Correct but should be [SimEng:ExpectationNode] as we use the class name in the prefixes

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
jj16791: Given their use in ExpectationNode's public scope, that wouldn't be possible. Unless there's some niche C++ logic to allow this which I don't know of

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+    return node;
+  }
+
+  /** Constructor for ExpectationNode instances. */
+  ExpectationNode(std::string key, ExpectedType type, bool optional)
+      : nodeKey_(key), type_(type), isOptional_(optional) {
+    if (nodeKey_ == wildcard) isWild_ = true;
+  }
+
+  ExpectationNode(){};
```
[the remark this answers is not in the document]
jj16791: The default constructor is needed to provide a root node when creating the tree-like structure via the addition of expectations

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+    ExpectationNode node =
+        ExpectationNode(key, ExpectedType::Valueless, optional);
+    return node;
+  }
+
+  /** Constructor for ExpectationNode instances. */
+  ExpectationNode(std::string key, ExpectedType type, bool optional)
+      : nodeKey_(key), type_(type), isOptional_(optional) {
+    if (nodeKey_ == wildcard) isWild_ = true;
+  }
```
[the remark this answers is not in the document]
jj16791: I think I understand what you're saying but we need the default constructor to be public so we can't force their usage

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+  T getDefault() const {
+    return getByType<T>(defaultValue_);
+  }
+
+  /** A getter function to retrieve the value type of a node. */
+  ExpectedType getType() const { return type_; }
+
+  /** A getter function to retrieve the child ExpectationNode instances of this
+   * node. */
+  std::vector<ExpectationNode> getChildren() const { return nodeChildren_; }
```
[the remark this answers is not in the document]
jj16791: You can't return a const here as we call `validateConfigNode` via the children nodes which cannot be qualified as a const given we possibly alter the passed node with default values. We could pass a reference but then would have to look the const qualifier of the getter function

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+
+  /** A getter function to retrieve whether the expectations should be applied
+   * to a sequence of config values. */
+  bool isSequence() const { return isSequence_; }
+
+  /** A getter function to retrieve whether the node is wild. */
+  bool isWild() const { return isWild_; }
```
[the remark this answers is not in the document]
jj16791: Gone through and replaced instance of just wild

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+                  "[SimEng:ModelConfig] Unexpected type given to "
+                  "ExpectationNode::getByType");
+
+    // Value existence check
+    if (variant.valueless_by_exception()) {
+      std::cerr << "[SimEng:ModelConfig] No value in passed variant within "
+                   "ExpectationNode with key "
+                << nodeKey_ << std::endl;
+      exit(1);
+    }
```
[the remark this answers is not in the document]
jj16791: You're correct in that this should be private as it's only used internally for the class. Also moved `typeToString`, `setDefaultValue` and `validateConfigNodeWithType` for the same reasoning.

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
jj16791: Comment made above class

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
FinnWilkinson: Looking at an existing in `DispatchIssueUnit`, structs such as `dependencyEntry` are only used within the scope of the DispatchIssueUnit class, but are declared above the class itself in the header file but still in the pipeline namespace.

To be consistent in the codebase, I stand by my first comment if these two definitions are required in ExceptionNode's public scope

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
FinnWilkinson: If the default constructor is needed to be public then thats ok - but this constructor could still be made private to ensure users will use the `createExpectation` functions?

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
FinnWilkinson: On this, should the `Expectations.hh` file be renamed to `ExpectationNode.hh` to be consistent with the file naming present in the rest of the project?

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
FinnWilkinson: Could add a brief comment above it to describe this?

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
FinnWilkinson: I may be missing something, but `validateConfigNode` and `validateConfigNodeWithType` don't seem to change the ExpectationNode itself, but rather just the ryaml::NodeRef passed into these functions?

If this isn't the case though - I feel as though the latter would be preferable (removing const but returning a reference to the children) to avoid multiple copies + creation of vectors?

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+        return "string";
+      case static_cast<size_t>(ExpectedType::UInteger):
+        return "unsigned integer";
+    }
+    return "unknown";
+  }
+
+  /** Setter function to set the default value for this node's associated config
+   * option. */
+  void setDefaultValue(DataTypeVariant var) { defaultValue_ = var; }
```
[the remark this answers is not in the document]
jj16791: Added checks for data type matching in all "setter" functions

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    return {false, "Success"};
+  }
+
+  /** Search through the held children to find a node with the key `childKey`.
+   * If no `childKey` can be found, then it is considered to be fatal for the
+   * simulation. However, if a wild node is present within the children,
+   * then return said child. */
+  ExpectationNode& operator[](std::string childKey) {
```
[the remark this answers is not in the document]
jj16791: The reference returned is modified/altered in many places within `ModelConfig.cc` especially within `postValidation()`

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+   * true, then the config values lower in the YAML hierarchy from the
+   * associated config option are a sequence of values (a set of values with no
+   * key). All values are validated against this node's expectations. */
+  bool isSequence_ = false;
+
+  /** Whether this instance of ExpectationNode is wild. If true, then when
+   * indexing this instance of ExpectationNode, any passed key will match. This
+   * is primarily used when one config option has many "child" values or YAML
+   * structures which follow the same pattern. */
+  bool isWild_ = false;
```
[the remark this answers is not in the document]
jj16791: Added comment elsewhere in code to resolve this

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+   */
+  std::vector<DataTypeVariant> expectedSet_;
+
+  /** Whether a value bounds have been defined as part of the expectation for
+   * the associated config option. */
+  bool definedBounds_ = false;
+
+  /** The value bounds the associated config option is expected to lie between.
+   */
+  std::pair<DataTypeVariant, DataTypeVariant> expectedBounds_;
```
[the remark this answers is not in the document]
jj16791: You can have a sequence with a set or a bound (kind of the point of the sequence classification, to map the value or bounds onto an unknown amount of children). I do like the idea of a union though

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+  /** Add a child node to the vector of children within this node. */
+  void addChild(ExpectationNode chld) {
+    // Ensure that if the new child is wild, one does not already exist in this
+    // instance's children
+    if (chld.getKey() == wildcard) {
+      if (hasWild_) {
+        std::cerr << "[SimEng:ModelConfig] Attempted to add multiple wild "
+                     "nodes to the same ExpectationNode instance of key "
+                  << nodeKey_ << std::endl;
+        exit(1);
```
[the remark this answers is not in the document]
jj16791: The way you want this is only somewhat possible. Due to the wildcard stuff, you'd never be able to print "Ports::1::Portname", only "Ports::*::Portname" which doesn't match your intent as far as I can tell. This is because that "1" doesn't exist in the expectation tree, only the config tree passed in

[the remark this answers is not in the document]
jj16791: The feasibility of doing it with the wildcard stuff issue is okay. We'd have to build a cumulative string which is passed down to each child when the parent adds it to its list of children.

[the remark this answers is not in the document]
jj16791: I'm not 100% sure on your intent here as the original question is flawed (i.e. "Ports::1::Portname" doesn't exist) so I don't if you meant to be applying this printout to the expectationTree checking or configTree validation

src/include/simeng/config/ModelConfig.hh
```diff
@@ -0,0 +1,121 @@
...
+#include <variant>
+
+#include "simeng/config/Expectations.hh"
+#include "simeng/config/yaml/ryml.hh"
+
+namespace simeng {
+namespace config {
+
+/** Enum representing the possible ISAs. */
+enum class ISA { AArch64, RV64 };
```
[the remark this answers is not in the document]
jj16791: ModelConfig has no link to SimInfo (unlike SimInfo which includes ModelConfig) and given the ISA usage in the ModelConfig header file, I don't know how we'd do this. I could be missing a trick here so if you validate a solution then I'd be happy to move it over

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
FinnWilkinson: That was a bad example on my part. What I meant was printing the cumulative string to this node, i.e. `[Parent][Some-Child][Some-grandchild]` or in the case of a wildcard `[Parent][*][some-grandchild]`

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+  }
+
+  /** Search through the held children to find a node with the key `childKey`.
+   * If no `childKey` can be found, then it is considered to be fatal for the
+   * simulation. However, if a wild node is present within the children,
+   * then return said child. */
+  ExpectationNode& operator[](std::string childKey) {
+    int wildIndex = -1;
+    // Search children for childKey and record latest wildcard children
+    for (size_t chld = 0; chld < nodeChildren_.size(); chld++) {
```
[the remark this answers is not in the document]
jj16791: changed

src/include/simeng/config/SimInfo.hh
```diff
@@ -0,0 +1,235 @@
...
+    return val;
+  }
+
+  /** A getter function to retrieve the config file path. */
+  static std::string getConfigPath() { return getInstance()->configFilePath_; }
+
+  /** A setter function to set the config file path. */
+  static std::string setConfigPath(std::string path) {
+    return getInstance()->configFilePath_ = path;
+  }
```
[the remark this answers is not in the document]
jj16791: Yes it should

src/include/simeng/config/SimInfo.hh
```diff
@@ -0,0 +1,235 @@
...
+
+  /** A getter function to retrieve whether or not the special files
+   * directories should be generated. */
+  static const bool getGenSpecFiles() {
+    return getInstance()->genSpecialFiles_;
+  }
+
+  /** A function used to reset the architectural register file structure. */
+  static void resetArchRegs() { getInstance()->resetArchRegStruct(); }
```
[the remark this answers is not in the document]
jj16791: Function no longer eixsts

src/include/simeng/config/SimInfo.hh
```diff
@@ -0,0 +1,235 @@
...
+
+  /** The validated model config file represented as a ryml:Tree. */
+  ryml::Tree validatedConfig_;
+
+  /** The ModelConfig instance used to create and maintain the model config
+   * file. */
+  simeng::config::ModelConfig mdlCnf_;
+
+  /** The path of the model config file. Defaults to "Default". */
+  std::string configFilePath_ = "Default";
```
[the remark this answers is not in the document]
jj16791: Updated and moved `#define DEFAULT_STR "Default"` to be in the correct scope

src/include/simeng/config/SimInfo.hh
jj16791: The changing of simMode to SimMode and the capitalising of each entry are the only points I agree with here. 

The structure of the SimEng framework is built upon namespaces to identify and separate regions of the codebase. The folder structure is also reminiscent of this ideal. Moving SimInfo and the enum classes outside of the config namespace is not desirable as they serve the specific purpose of configuring the simulation. I don't see how the change from `config::SimMode` to `SimMode` improves cleanliness or readability and if anything, given we are using the config namespace, removing SimInfo from it would arguably make it less intuitive given the namespace usage throughout the framework.

src/lib/CoreInstance.cc
jj16791: Yes, good spot

src/lib/CoreInstance.cc
jj16791: Removed all instances of `simeng::`

src/lib/arch/aarch64/Architecture.cc
jj16791: `getSysRegVecIndex` is currently unused (perhaps should be removed?) as the getting of the sys reg tag has to be done without a critical failure to the simulation. Incorrectly speculated instructions can possibly generate an invalid sys reg tag

src/lib/arch/aarch64/Architecture.cc
jj16791: That's fine by me given prior changes in comments further up

src/lib/arch/aarch64/MicroDecoder.cc
jj16791: Unlike other instances commented on, changing this one now as it allows for instructionSplit_ to be `const`

src/lib/arch/riscv/InstructionMetadata.hh
jj16791: Supplies access to `arch::riscv::groupInheritance` and `arch::riscv::Opcode::RISCV_INSTRUCTION_LIST_END` in `ModelConfig.cc` (approach chosen for consistency with already integrated aarch64 work)

src/lib/config/ModelConfig.cc
jj16791: Changed

src/lib/config/ModelConfig.cc
jj16791: Somewhat agree with this but where does it end regarding the other config options? Should we also have stuff like `Load-Bandwidth-Bytes`

src/lib/config/ModelConfig.cc
jj16791: I'm not sure what is being asked here. Are you talking about the default cases?

src/lib/config/ModelConfig.cc
jj16791: Given ``, DataTypeVariantonly 64bit int/uints are supported at the moment

src/lib/config/ModelConfig.cc
jj16791: You're constructing the default config Tree and you don't want the config node key value to be `*` in the resultant tree. In the comment above, it is explained as "then change it to be an appropriate value in the resultant config file"

src/include/simeng/config/Expectations.hh
```diff
@@ -0,0 +1,403 @@
...
+  ExpectedType type_ = ExpectedType::Valueless;
+
+  /** Whether the config option associated with this node is optional. */
+  bool isOptional_ = false;
+
+  /** Whether the config option associated with this node is a sequence. If
+   * true, then the config values lower in the YAML hierarchy from the
+   * associated config option are a sequence of values (a set of values with no
+   * key). All values are validated against this node's expectations. */
+  bool isSequence_ = false;
```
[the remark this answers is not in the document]
jj16791: Added a better comment

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
jj16791: I'll change to be a ref

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
jj16791: Yes, I follow the logic now

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
FinnWilkinson: Understood 👍

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
FinnWilkinson: If these functions do not alter any member variables of the class they belong to then they should be able to be made `const` - even if they change any argument passed in

src/include/simeng/config/SimInfo.hh
FinnWilkinson: Thats fair enough

src/include/simeng/config/ModelConfig.hh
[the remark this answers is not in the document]
FinnWilkinson: I'd hadn't realised that `ModelConfig.hh` didn't include `SimInfo.hh`....

src/lib/arch/aarch64/Architecture.cc
FinnWilkinson: Given it is unused then I would say that `getSysRegVecIndex` should be removed

src/include/simeng/config/Expectations.hh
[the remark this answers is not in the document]
jj16791: See edited comments

src/lib/arch/riscv/InstructionMetadata.hh
FinnWilkinson: I don't think this PR should address this, and I am happy with the reason provided above - but including headers for small items such as these creates messy dependancies in the project, something which (I think) should be rectified and cleaned up at some-point

src/lib/config/ModelConfig.cc
FinnWilkinson: I personally don't object to being more verbose with all config options, but specifically about `Clock-Frequency`; given `Timer-Frequency` doesn't abide by the same units I feel that clarification could be useful whereas all bandwidths are given in bytes (and are generally the accepted standard unit for those metrics, but so it GHz and clock frequency....)

src/lib/arch/riscv/InstructionMetadata.hh
jj16791: The entire include logic based around `InstructionMetadata` needs to be redone. It's an artefact of the old SimEng setups that has been hidden away. Only realised how bad it got during this rebase onto dev

src/lib/config/ModelConfig.cc
FinnWilkinson: Is this something worth expanding? i.e. having `uint8_t` etc in `DataTypeVariant`. I'm not swung either way

dANW34V3R (changes requested):
Lots of things improved greatly in this PR e.g. definition of base config used for testing. Some minor comments but I generally disagree with Finn on the use of the ">>" operator in most places. I think it is much cleaner and easier to read while using the getter function adds a lot more unnecessary syntax. While ">>" requires 2 lines in lots of places, I don't think this is an issue as the getter line often becomes too long and splits over 2 lines anyway

src/include/simeng/config/SimInfo.hh
```diff
@@ -0,0 +1,218 @@
...
+    validatedConfig_ = modelConfig_.getConfig();
+    extractValues();
+  }
+
+  /** Gets the static instance of the SimInfo class. */
+  static std::unique_ptr<SimInfo>& getInstance() {
+    static std::unique_ptr<SimInfo> SimInfoClass = nullptr;
+    if (SimInfoClass == nullptr) {
+      SimInfoClass = std::unique_ptr<SimInfo>(new SimInfo());
+    }
```
ABenC377, line 139: Why is there an if statement here -- won't it always have the value of 'nullptr' as we've set it to that in the line above?

src/include/simeng/config/ExpectationNode.hh
```diff
@@ -0,0 +1,513 @@
...
+
+  /** Whether a value bounds have been defined as part of the expectation for
+   * the associated config option. */
+  bool definedBounds_ = false;
+
+  /** The value bounds the associated config option is expected to lie between.
+   */
+  std::pair<DataTypeVariant, DataTypeVariant> expectedBounds_;
+
+  /** The instances of ExpectationNode's held within this node. Considered to be
```
ABenC377, line 507: Don't think there should be an apostrophe in ExpectationNodes

src/lib/CoreInstance.cc
ABenC377: I fall on Finn's side of the fence here.  The getValue function is easier for me to read and understand than the >> operator.

src/lib/GenericPredictor.cc
```diff
@@ -4,17 +4,23 @@
...
+  ryml::ConstNodeRef config = config::SimInfo::getConfig();
+  config["Branch-Predictor"]["BTB-Tag-Bits"] >> btbBits_;
+  // Create branch prediction structures
+  config["Branch-Predictor"]["Saturating-Count-Bits"] >> satCntBits_;
+  config["Branch-Predictor"]["Global-History-Length"] >> globalHistoryLength_;
+  config["Branch-Predictor"]["RAS-entries"] >> rasSize_;
+  // Calculate the saturation counter boundary between weakly taken and
+  // not-taken. `(2 ^ num_sat_cnt_bits) / 2` gives the weakly taken state
+  // value
+  uint8_t weaklyTaken = std::pow(2, (satCntBits_ - 1));
```
ABenC377, line 17: Is there a benefit to using 'std::pow' over '<<' when the power is 2?  I'd have thought '<<' would be faster.

src/lib/arch/aarch64/Architecture.cc
```diff
@@ -88,24 +96,28 @@ Architecture::Architecture(kernel::Linux& kernel, YAML::Node config)
...
-      YAML::Node group_node = config["Ports"][i]["Instruction-Group-Support"];
-      for (size_t j = 0; j < group_node.size(); j++) {
-        uint16_t group = group_node[j].as<uint16_t>();
-        uint8_t newPort = static_cast<uint8_t>(i);
+      ryml::ConstNodeRef group_node =
+          config["Ports"][i]["Instruction-Group-Support-Nums"];
+      for (size_t j = 0; j < group_node.num_children(); j++) {
+        uint16_t group;
+        group_node[j] >> group;
+        uint16_t newPort = static_cast<uint16_t>(i);
```
ABenC377, line 120: Do we need to go from eight to 16 bits for newPort?

src/lib/arch/riscv/Architecture.cc
```diff
@@ -70,24 +81,28 @@ Architecture::Architecture(kernel::Linux& kernel, YAML::Node config)
...
-      YAML::Node group_node = config["Ports"][i]["Instruction-Group-Support"];
-      for (size_t j = 0; j < group_node.size(); j++) {
-        uint16_t group = group_node[j].as<uint16_t>();
-        uint8_t newPort = static_cast<uint8_t>(i);
+      ryml::ConstNodeRef group_node =
+          config["Ports"][i]["Instruction-Group-Support-Nums"];
+      for (size_t j = 0; j < group_node.num_children(); j++) {
+        uint16_t group;
+        group_node[j] >> group;
+        uint16_t newPort = static_cast<uint16_t>(i);
```
ABenC377, line 105: Does this need to be uint16_t rather than uint8_t?

src/lib/models/outoforder/Core.cc
```diff
@@ -16,77 +16,92 @@ namespace outoforder {
...
+          config::SimInfo::getValue<uint16_t>(
+              config["LSQ-L1-Interface"]["Permitted-Requests-Per-Cycle"]),
+          config::SimInfo::getValue<uint16_t>(
+              config["LSQ-L1-Interface"]["Permitted-Loads-Per-Cycle"]),
+          config::SimInfo::getValue<uint16_t>(
+              config["LSQ-L1-Interface"]["Permitted-Stores-Per-Cycle"])),
       fetchUnit_(fetchToDecodeBuffer_, instructionMemory, processMemorySize,
-                 entryPoint, config["Fetch"]["Fetch-Block-Size"].as<uint16_t>(),
+                 entryPoint,
+                 config::SimInfo::getValue<uint32_t>(
```
ABenC377, line 67: Do we need to go from uint16_t to uint32_t?

src/include/simeng/config/SimInfo.hh
FinnWilkinson: As we initialise a static unique_ptr to a new SimInfo object, the variable will retian its state between all function calls. So, on the first call to the function `SimInfoClass` will indeed be nullptr, but on all subsequent calls the same `SimInfoClass` vaiable has already been instantiated.

ABenC377: gotcha -- makes sense

src/lib/CoreInstance.cc
jj16791: Config getter functionality has been changed to new custom ryml `as` function as discussed offline.

src/lib/SpecialFileDirGen.cc
jj16791: Config getter functionality has been changed to new custom ryml `as` function as discussed offline.

CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.13)
 include(FetchContent)
+include(CheckCXXCompilerFlag)
```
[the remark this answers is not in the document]
jj16791: Used by `check_cxx_compiler_flag`

src/include/simeng/arch/aarch64/InstructionGroups.hh
```diff
@@ -92,12 +92,19 @@ const uint16_t LOAD_SME = 82;
 const uint16_t STORE_ADDRESS_SME = 83;
 const uint16_t STORE_DATA_SME = 84;
 const uint16_t STORE_SME = 85;
+const uint16_t ALL = 86;
+const uint16_t NONE = 87;
```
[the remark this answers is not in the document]
jj16791: I guess not. Included to provide flexible group definitions but thinking about it, I can think of a use-case. Will remove.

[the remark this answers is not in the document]
jj16791: A replacement for when ALL is appropriate (e.g. default value for EU blocking groups)

src/include/simeng/arch/riscv/InstructionGroups.hh
```diff
@@ -19,11 +19,16 @@ const uint16_t STORE_INT = 9;
 const uint16_t LOAD = 10;
 const uint16_t STORE = 11;
 const uint16_t BRANCH = 12;
+const uint16_t ALL = 13;
+const uint16_t NONE = 14;
```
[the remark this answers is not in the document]
jj16791: See other comment of similar questioning

test/integration/ConfigTest.cc
```diff
@@ -0,0 +1,307 @@
+#include <fstream>
+#include <iostream>
+
+#include "gtest/gtest.h"
```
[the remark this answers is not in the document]
jj16791: "Integration" testing is a standardised form of testing similar to regression and unit. It suits the current tests held in that folder well given it tests all classes within the config namespace

test/unit/ISATest.cc
```diff
@@ -8,16 +8,16 @@ namespace {
...
+      "Timer-Frequency: 100, Micro-Operations: True, Vector-Length: 512, "
+      "Streaming-Vector-Length: 512}, CPU-Info: {Generate-Special-Dir: "
+      "False}}");
+
   std::unique_ptr<simeng::arch::Architecture> isa =
-      std::make_unique<simeng::arch::aarch64::Architecture>(kernel, config);
+      std::make_unique<simeng::arch::aarch64::Architecture>(kernel);
 
-  EXPECT_GT(isa->getRegisterFileStructures().size(), 0);
+  EXPECT_GT(isa->getNumSystemRegisters(), 0);
```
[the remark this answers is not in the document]
jj16791: Code no longer exists. Can't test logic anymore as it doesn't exist in the class

src/include/simeng/config/ExpectationNode.hh
```diff
@@ -0,0 +1,513 @@
...
+ *
+ * can be represented by
+ *
+ *                             |--["value_node_A"]
+ * ["parent_node"]-[wildcard]--|--["value_node_B"]
+ *                             |--["value_node_C"]
+ */
+class ExpectationNode {
+ public:
+  /** NOTE: The index of the ExpectedType enum matches that of DataTypeVariant.
```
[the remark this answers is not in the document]
jj16791: Good point.

src/include/simeng/pipeline/FetchUnit.hh
```diff
@@ -39,7 +39,7 @@ class FetchUnit {
   /** Construct a fetch unit with a reference to an output buffer, the ISA, and
    * the current branch predictor, and information on the instruction memory. */
   FetchUnit(PipelineBuffer<MacroOp>& output, MemoryInterface& instructionMemory,
-            uint64_t programByteLength, uint64_t entryPoint, uint8_t blockSize,
+            uint64_t programByteLength, uint64_t entryPoint, uint32_t blockSize,
```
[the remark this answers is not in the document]
jj16791: Because in the config we support up to 65536 in the block size. Not sure how this was changed prior.

src/lib/pipeline/FetchUnit.cc
```diff
@@ -6,7 +6,7 @@ namespace pipeline {
 FetchUnit::FetchUnit(PipelineBuffer<MacroOp>& output,
                      MemoryInterface& instructionMemory,
                      uint64_t programByteLength, uint64_t entryPoint,
-                     uint8_t blockSize, const arch::Architecture& isa,
+                     uint32_t blockSize, const arch::Architecture& isa,
```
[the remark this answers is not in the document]
jj16791: See other comment

src/include/simeng/RegisterFileSet.hh
```diff
@@ -30,6 +30,10 @@ struct RegisterFileStructure {
   uint16_t bytes;
   /** The number of registers. */
   uint16_t quantity;
+  /** Check for the equality of two RegisterFileStructure structs. */
+  bool operator==(const RegisterFileStructure& other) const {
```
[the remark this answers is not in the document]
jj16791: Line 34 of ConfigTest.cc `EXPECT_EQ(simeng::config::SimInfo::getArchRegStruct(), archRegStruct);`

src/lib/GenericPredictor.cc
jj16791: From some old issues, we had with a compiler. I've changed it to use `<<` and altered some of the private member variable data widths to better reflect their possible values

src/lib/models/outoforder/Core.cc
jj16791: Miscalculated bit needs, don't need 17 so can narrow to 16bits

src/lib/arch/aarch64/Architecture.cc
jj16791: Was changed as we've used more than 256 ports in simulations prior

src/lib/arch/riscv/Architecture.cc
jj16791: Was changed as we've used more than 256 ports in simulations prior

src/lib/config/ModelConfig.cc
jj16791: I didn't due to the extra code size needed to support this but thinking about the added validation when having errors due to the provided due to invalid narrowing of config values (in addition to the bound/set expectations), I think it's worth it

src/lib/config/ModelConfig.cc
jj16791: Understand now, thank you for explaining. Unified values under one variable.

src/lib/config/ModelConfig.cc
```diff
@@ -0,0 +1,1007 @@
...
+  expectations_["Core"].addChild(
+      ExpectationNode::createExpectation<std::string>("emulation",
+                                                      "Simulation-Mode"));
+  expectations_["Core"]["Simulation-Mode"].setValueSet(
+      std::vector<std::string>{"emulation", "inorderpipelined", "outoforder"});
+
+  const float clockFreqUpperBound = 10.f;
+  expectations_["Core"].addChild(
+      ExpectationNode::createExpectation<float>(1.f, "Clock-Frequency"));
+  expectations_["Core"]["Clock-Frequency"].setValueBounds(0.f,
```
FinnWilkinson, line 268: Does a minimum clock frequency of 0GHz make sense? Perhaps the lower-bound should be 1mhz or 0.001 GHz?

src/lib/GenericPredictor.cc
```diff
@@ -4,17 +4,25 @@
...
-                .as<uint16_t>(),
-            0}),
-      satCntBits_(
-          config["Branch-Predictor"]["Saturating-Count-Bits"].as<uint64_t>()),
-      globalHistoryLength_(
-          config["Branch-Predictor"]["Global-History-Length"].as<uint64_t>()),
-      rasSize_(config["Branch-Predictor"]["RAS-entries"].as<uint64_t>()) {
+GenericPredictor::GenericPredictor() {
+  ryml::ConstNodeRef config = config::SimInfo::getConfig();
+  btbBits_ = config["Branch-Predictor"]["BTB-Tag-Bits"].as<uint8_t>();
```
FinnWilkinson, line 9: There seems to be some mixing and matching regarding if `config` is passed as a constructor / function argument (with default value set to `config::SimInfo::getConfig()`) or if it is fetched at the start of the constructor / function.

For example, Linux process passes config as an argument with a default value, but generic predictor fetches config at the start of the function.

Passing config as an argument with a defualt value does allow for the initialiser list to be utilised more, so I would probably have this method as our goto

ABenC377 (dismissed):
👍🏻

src/lib/config/ModelConfig.cc
```diff
@@ -0,0 +1,1007 @@
...
+    expectations_["Core"].addChild(ExpectationNode::createExpectation<uint64_t>(
+        128, "Vector-Length", true));
+    expectations_["Core"]["Vector-Length"].setValueSet(
+        std::vector<uint64_t>{128, 256, 384, 512, 640, 768, 896, 1024, 1152,
+                              1280, 1408, 1536, 1664, 1792, 1920, 2048});
+
+    expectations_["Core"].addChild(ExpectationNode::createExpectation<uint64_t>(
+        128, "Streaming-Vector-Length", true));
+    expectations_["Core"]["Streaming-Vector-Length"].setValueSet(
+        std::vector<uint64_t>{128, 256, 384, 512, 1024, 2048});
```
FinnWilkinson, line 311: 384 is not a valid SME Streaming Vector length.

ABenC377 (dismissed):
👍🏻

src/lib/config/ModelConfig.cc
jj16791: Good spot, accidentally left in

src/include/simeng/arch/aarch64/MicroDecoder.hh
```diff
@@ -21,7 +20,7 @@ struct OpType {
 class MicroDecoder {
  public:
   /** Construct a micro decoder for splitting relevant instructons. */
-  MicroDecoder(YAML::Node config);
+  MicroDecoder(ryml::ConstNodeRef config = config::SimInfo::getConfig());
   ~MicroDecoder();
```
FinnWilkinson, line 24: Add blank line between constructor and destructor

src/lib/GenericPredictor.cc
```diff
@@ -4,17 +4,24 @@
...
-          config["Branch-Predictor"]["Global-History-Length"].as<uint64_t>()),
-      rasSize_(config["Branch-Predictor"]["RAS-entries"].as<uint64_t>()) {
+GenericPredictor::GenericPredictor(ryml::ConstNodeRef config) {
+  btbBits_ = config["Branch-Predictor"]["BTB-Tag-Bits"].as<uint8_t>();
+  // Create branch prediction structures
+  satCntBits_ =
+      config["Branch-Predictor"]["Saturating-Count-Bits"].as<uint8_t>();
+  globalHistoryLength_ =
+      config["Branch-Predictor"]["Global-History-Length"].as<uint16_t>();
+  rasSize_ = config["Branch-Predictor"]["RAS-entries"].as<uint16_t>();
```
FinnWilkinson, line 14: These 4 variables could be added to initialiser list

src/lib/arch/aarch64/Architecture.cc
```diff
@@ -26,15 +21,19 @@ Architecture::Architecture(kernel::Linux& kernel, YAML::Node config)
...
   cs_option(capstoneHandle, CS_OPT_DETAIL, CS_OPT_ON);
 
+  // Initialise SVE and SME vector lengths
+  VL_ = config["Core"]["Vector-Length"].as<uint64_t>();
+  SVL_ = config["Core"]["Streaming-Vector-Length"].as<uint64_t>();
+
+  // Initialise virtual counter timer increment frequency
+  vctModulo_ = (config["Core"]["Clock-Frequency-GHz"].as<float>() * 1e9) /
+               (config["Core"]["Timer-Frequency-MHz"].as<uint32_t>() * 1e6);
+
```
FinnWilkinson, line 31: VL_, SVL_, and vctModule_ could be re-added to initialiser list

src/lib/arch/riscv/Architecture.cc
tom91136: I think at some point it might be nice to use something like:
```cpp
template <typename I, typename J> static std::enable_if_t<std::is_signed_v<I> && std::is_signed_v<J>, I> int_cast(J value) {
  if (value < std::numeric_limits<I>::min() || value > std::numeric_limits<I>::max()) {
    throw std::out_of_range("out of range");
  }
  return static_cast<I>(value);
}

template <typename I, typename J> static std::enable_if_t<std::is_signed_v<I> && std::is_unsigned_v<J>, I> int_cast(J value) {
  if (value > static_cast<std::make_unsigned_t<I>>(std::numeric_limits<I>::max())) {
    throw std::out_of_range("out of range");
  }
  return static_cast<I>(value);
}

template <typename I, typename J> static std::enable_if_t<std::is_unsigned_v<I> && std::is_signed_v<J>, I> int_cast(J value) {
  if (value < 0 || static_cast<std::make_unsigned_t<J>>(value) > std::numeric_limits<I>::max()) {
    throw std::out_of_range("out of range");
  }
  return static_cast<I>(value);
}

template <typename I, typename J> static std::enable_if_t<std::is_unsigned_v<I> && std::is_unsigned_v<J>, I> int_cast(J value) {
  if (value > std::numeric_limits<I>::max()) {
    throw std::out_of_range("out of range");
  }
  return static_cast<I>(value);
}
```

And then we can do:
```
uint16_t newPort = int_cast<uint16_t>(i);
```
So that it doesn't blow up if it does go out of bounds.

FinnWilkinson (dismissed):
All looks good, few minor changes but not critical

src/lib/arch/riscv/Architecture.cc
dANW34V3R: I think this is starting to go against our "simple cpp" matra

src/lib/GenericPredictor.cc
jj16791: Good spot. Also changed in `SpecialFilDenGen`

src/include/simeng/RegisterFileSet.hh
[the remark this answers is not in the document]
dANW34V3R: Ah, implicit use

dANW34V3R (changes requested):
```.as``` looks great. Found 1 typo

**Review by FinnWilkinson — DISMISSED** [review]

File: configs/a64fx_SME.yaml [hunk 64]
```diff
@@ -1,14 +1,11 @@
...
-  # Clock Frequency is in GHz.
-  Clock-Frequency: 1.8
-  # Timer-Frequency is in MHz.
-  Timer-Frequency: 100
+  ISA: AArch64]
```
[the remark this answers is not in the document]
jj16791: Something like this will be flagged once the current planned updates to the Jenkins CI have been implemented. Fix the typo

dANW34V3R (approved):

FinnWilkinson (approved):

ABenC377 (dismissed):

FinnWilkinson (changes requested):
Docs read well, only a few minor points

docs/sphinx/developer/arch/supported/aarch64.rst
```diff
@@ -43,7 +43,7 @@ The above diagram describes the instruction groups currently implemented for the
 
 This hierarchy-based naming convention has been chosen to provide the user with greater control over the number of instructions grouped under one name, whilst also remaining intuitive. A variety of combinations/instruction scopes can be defined through this method and only uses a small set of easily interpreted operation descriptions.
 
-If the supplied instruction groups don't provide a small enough scope, a Capstone opcode can be used instead (found in ``SimEng/build/_deps/capstone-lib-src/arch/AArch64/AArch64GenInstrInfo.inc``) with the format ``~{CAPSTONE_OPCODE}``.
+If the supplied instruction groups don't provide a small enough scope, a numerical Capstone opcode can be used instead (found in ``SimEng/build/_deps/capstone-lib-src/arch/AArch64/AArch64GenInstrInfo.inc``).
```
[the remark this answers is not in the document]
FinnWilkinson: Given we dropped the usage of `~` prefix, this is fine. Resolving

docs/sphinx/developer/models/index.rst
```diff
@@ -55,3 +55,159 @@ The current existing processors have supplied configuration files:
...
+
+
+Adding model configuration options
+----------------------------------
+
+SimEng utilises a scheme which defines value expectations on mandatory and optional configuration options. The value expectations consist of:
+
+- A default value to be used when creating an in-code default config file
+- A boolean to describe whether the config option is mandatory or optional
+- Either a set of values or an upper and lower bound which the supplied config option must conform to
```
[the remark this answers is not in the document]
FinnWilkinson: Resolved offline - no indentation required

FinnWilkinson (approved):

docs/sphinx/developer/models/index.rst
```diff
@@ -55,3 +55,159 @@ The current existing processors have supplied configuration files:
...
+      ExpectationNode::createExpectation<uint16_t>(1, "Execution-Latency"));
+    expectations_["Latencies"][wildcard]["Execution-Latency"]
+      .setValueBounds<uint16_t>(1, UINT16_MAX);
+
+When validating all config options under the "Latencies" key, the ``wildcard`` node created will be used for each option.
+
+Sequence Nodes
+**************
+
+Similar to the motivation behind the ``wildcard`` node usage, a value expectation can be set to be applied to all values in a YAML sequence/list/array/etc. By calling ``.setAsSequence()`` on an instance of ``ExpectationNode``, its value expectations will be applied to a YAML sequence of values. Below is an example of when you might use this.
```
dANW34V3R, line 179: Does ```sequence/list/...``` require double quotes?

dANW34V3R (commented):
Minor comments

docs/sphinx/developer/models/index.rst
```diff
@@ -55,3 +55,159 @@ The current existing processors have supplied configuration files:
...
+
+Which is used for those nodes whose corresponding config option has no value (only a key and children).
+
+.. code-block:: text
+
+        2.
+            template <typename T>
+            createExpectation(T defaultValue, std::string key, bool optional = false)
+
+Which is used for those nodes whose corresponding config option do have a value. The function is templated to support different config value types (e.g. T = string, integer, floating point, etc).
```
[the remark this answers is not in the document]
jj16791: Yeah, I agree it doesn't read well. Changing to "does"

docs/sphinx/assets/expectations.png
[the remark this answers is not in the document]
jj16791: Correcting RV64 to be rv64 in diagram

docs/sphinx/developer/models/index.rst
dANW34V3R: No quotes required, not "in-code" language

ABenC377 (commented):
Only one question

docs/sphinx/developer/models/index.rst
```diff
@@ -55,3 +55,159 @@ The current existing processors have supplied configuration files:
...
+        ISA: AArch64
+        Simulation-Mode: outoforder
+
+The ``ExpectationNode`` tree structure would be:
+
+.. image:: ../../assets/expectations.png
+  :width: 350
+  :alt: ExpectationNode structure example
+
+The tree structure formed with ``ExpectationNode`` instances is constructed through the ``addChild(...)`` utility of the ``ExpectationNode`` class. Starting at a root node, ``addChild(...)`` is used to create various branches of the tree which map to the various config options as described in the above diagram. For those nodes which hold config values, either ``setValueBounds(...)`` or ``setValueSet(...)`` can be used to set the config value restrictions.
```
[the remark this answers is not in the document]
jj16791: It should be "is", well spotted

ABenC377 (approved):

src/lib/arch/aarch64/Architecture.cc
```diff
@@ -51,12 +47,13 @@ Architecture::Architecture(kernel::Linux& kernel, YAML::Node config)
...
-    YAML::Node port_node = config["Latencies"][i];
+  for (size_t i = 0; i < config["Latencies"].num_children(); i++) {
+    ryml::ConstNodeRef port_node = config["Latencies"][i];
     uint16_t latency = port_node["Execution-Latency"].as<uint16_t>();
     uint16_t throughput = port_node["Execution-Throughput"].as<uint16_t>();
-    for (size_t j = 0; j < port_node["Instruction-Group"].size(); j++) {
-      uint16_t group = port_node["Instruction-Group"][j].as<uint16_t>();
+    for (size_t j = 0; j < port_node["Instruction-Group-Nums"].num_children();
+         j++) {
+      uint16_t group = port_node["Instruction-Group-Nums"][j].as<uint16_t>();
```
FinnWilkinson, line 56: I may have missed this, but where does `"Instruction-Group-Nums"` come from? Added on config validation?

src/tools/simeng/main.cc
```diff
@@ -82,13 +84,25 @@ int main(int argc, char** argv) {
...
   std::cout << "[SimEng] Workload: " << executablePath;
   for (const auto& arg : executableArgs) std::cout << " " << arg;
   std::cout << std::endl;
-  std::cout << "[SimEng] Config file: " << configFilePath << std::endl;
-  std::cout << "[SimEng] ISA: " << coreInstance->getISAString() << std::endl;
+  std::cout << "[SimEng] Config file: "
+            << simeng::config::SimInfo::getConfigPath() << std::endl;
+  std::cout << "[SimEng] ISA: " << simeng::config::SimInfo::getISAString()
+            << std::endl;
+  std::cout << "[SimEng] Local Special File directory: ";
```
FinnWilkinson, line 96: Should the print out say "Auto-Generated Special File Directory" given the used files will always be local?

FinnWilkinson (commented):
A few Minor points

src/lib/arch/aarch64/Architecture.cc
jj16791: It stores the string to int conversions created in `ModelConfig::postValidation()`

src/tools/simeng/main.cc
jj16791: Good point and I'm happy with the suggested change

FinnWilkinson (approved):

ABenC377 (approved):

dANW34V3R (approved):

PR #353, opened 2023-12-08
Added .cache to gitignore

rahahahat:
This directory is used by clangd and any tool that uses clangd.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@
 obj
 build
 xcode-build
+cmake-build-debug
+cmake-build-release
 docs/sphinx/_build
 binaries
 install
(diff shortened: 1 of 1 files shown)


FinnWilkinson (dismissed):

dANW34V3R (commented):
Previously I have attempted to add the files `cmake-build-debug` and `cmake-build-release` which are the default build folder names produced by CLion. This was rejected at the time but does fall in line with some of the other files excluded by this .gitignore. If this PR is approved then these should be added also

dANW34V3R (approved):

FinnWilkinson (approved):

jj16791 (approved):

PR #364, opened 2023-12-18
Correct CSRRW State Change

### LINKED ISSUE #359: CloverLeaf RISC-V Infinity Issue
When running CloverLeaf compiled with GCC 12.2 targeting rv64g and applying the fix from register-bug-fixes the following output is given:

> Clover Version Infinity
>        MPI Version
>    Task Count      1
> 
> 
> Clover Version Infinity
>        MPI Version
>    Task Count      1
> 
>  Output file clover.out opened. All output will go there.
>  Step       1 time    Infinity control    sound    timestep   Infinity       1,       1 x  Infinity y  Infinity
>  Wall clock                   Infinity
>  Average time per cell                   Infinity
>  Step time per cell                      Infinity
>  Step       2 time    Infinity control    sound    timestep   Infinity       1,       1 x  Infinity y  Infinity
>  Wall clock                   Infinity
>  Average time per cell                   Infinity
>  Step time per cell                      Infinity

This does not occur before commit: 73ced0f1 on the RISCV_F branch

dANW34V3R:
Closes #359 

This PR fixes issue #359 causing CloverLeaf compiled with GCC 12.2 to produce Inf for all floating point values. At beginning of the program execution `fsflags Ra` is called repeatedly. This is a pseudoinstruction for `csrrw x0, fflags, Ra` which discards the result in the zero register. 

The code in the execution handler blindly updated the first value held in the destination register span without checking if it is valid first which is not the case when the only destination is X0. This caused overwriting of registers incorrectly. A simple check is introduced to fix this.

Also in this PR:

- More accurate dummy value returned for the fsrm instruction
- Removal of unused variables in execute

### DIFF
--- a/src/lib/arch/riscv/ExceptionHandler.cc
+++ b/src/lib/arch/riscv/ExceptionHandler.cc
@@ -649,7 +649,7 @@ bool ExceptionHandler::init() {
     auto operands = instruction_.getSourceOperands();
     auto destinationRegs = instruction_.getDestinationRegisters();
 
-    uint8_t rm = 0b110; // Set to invalid rounding mode
+    uint8_t rm = 0b110;  // Set to invalid rounding mode
     uint64_t result = 0;
 
     ProcessStateChange stateChange;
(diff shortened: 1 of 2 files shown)


dANW34V3R:
rerun tests

dANW34V3R:
#rerun tests

jj16791 (approved):

JosephMoore25 (approved):

FinnWilkinson (approved):

PR #366, opened 2023-12-22
Optimise logic in ReorderBuffer::commitMicroOps

rahahahat:


src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -34,39 +35,31 @@ void ReorderBuffer::reserve(const std::shared_ptr<Instruction>& insn) {
...
-        if (!buffer_[index]->isWaitingCommit()) {
-          return;
-        } else if (buffer_[index]->isLastMicroOp()) {
-          // all microOps must be in ROB for the commit to be valid
-          validForCommit = true;
-        }
-      }
-      if (!validForCommit) return;
+  size_t index = 0;
+  uint8_t ins_cnt = 0;
```
rahahahat, line 42: @jj16791 Can the number of microops generated exceed 256?

FinnWilkinson:
Please add a description, assignee, labels, and ensure the status on the associated github project is correct

src/lib/pipeline/ReorderBuffer.cc
FinnWilkinson: I think this is the number of micro-ops for a single macro-op rather than total micro-ops in the ROB.

In such a case, the max number of micro-ops generated per macro op is implementation defined. The only candidate I can think of for currently supportable instructions is the AArch64 gather/scatter loads IF the vector length is 2048-bits and you are doing a byte load (i.e. loading 256 byte elements). 

However, checking the spec the gather / scatter instructions which load/store single bytes only permit the source/destination register to be of the `.S` or `.D` variant - meaning that there would be a max of 64 or 32 elements per vector respectively to scatter or gather.

In summary - no, the number of micro-ops per macro-op is very very unlikely to exceed 256.

rahahahat: Sorry, I should've specified i did mean number of micro-ops per macro-op. 

> In summary - no, the number of micro-ops per macro-op is very very unlikely to exceed 256.
When you say very very unlikely, that means there it could exceed 256 right? If that's the case the data type above needs to be incremented otherwise the logic is broken

rahahahat:
> Please add a description, assignee, labels, and ensure the status on the associated github project is correct

I've added labels and asignee. I don't think the description is necessary as the title communicates everything that is done. I've just changed one function.

src/lib/pipeline/ReorderBuffer.cc
FinnWilkinson: Currently it could never exceed 256, but in the future there is a extremely small chance it could. For now I would keep it the same

.gitignore
```diff
@@ -2,6 +2,7 @@
 .vscode
 .idea
 .DS_Store
+.cache
```
FinnWilkinson, line 5: This is duplicated from PR #353, please remove it from this PR or close the other PR

src/lib/pipeline/ReorderBuffer.cc
FinnWilkinson, line 42: To avoid confusion akin to the other comment left on this variable, perhaps rename this to `uopCount` or `microOpCount`?

src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -34,39 +35,31 @@ void ReorderBuffer::reserve(const std::shared_ptr<Instruction>& insn) {
...
-        buffer_[firstOp]->setCommitReady();
-      }
+  while (index < bsize) {
+    auto& uop = buffer_[index];
+    uint64_t mop_id = uop->getInstructionId();
+
+    if (mop_id == insnId) {
+      if (!uop->isWaitingCommit()) return;
+      ins_cnt++;
+    } else if (ins_cnt && mop_id != insnId) {
```
FinnWilkinson, line 51: `mop_id && insnID` in this else if seems redundant given that this else if can only be reached if this is true

src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -34,39 +35,31 @@ void ReorderBuffer::reserve(const std::shared_ptr<Instruction>& insn) {
...
     }
+    index++;
   }
+
+  index = index - ins_cnt;
+  for (int x = 0; x < ins_cnt; x++) {
+    buffer_[index]->setCommitReady();
+    index++;
+  }
+
```
FinnWilkinson, line 62: Generally looks good. I think some comments akin to the previous implementation, plus one explaining that the `setCommitReady` logic is only reached iff all micro-ops of a macro-op are waitingCommit, would be beneficial.

src/lib/pipeline/ReorderBuffer.cc
FinnWilkinson: After a full review, there would be no harm in making this a `uint16` to "future proof" this

src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -34,39 +35,31 @@ void ReorderBuffer::reserve(const std::shared_ptr<Instruction>& insn) {
...
+      if (!uop->isWaitingCommit()) return;
+      ins_cnt++;
+    } else if (ins_cnt && mop_id != insnId) {
+      break;
     }
+    index++;
   }
+
+  index = index - ins_cnt;
+  for (int x = 0; x < ins_cnt; x++) {
```
FinnWilkinson, line 58: Could `x` be given the same type as `ins_cnt`

FinnWilkinson (changes requested):
All looks fine, just a few comments made. 

Could you add to the description of this PR the performance improvement seen on your local system and on the Mac Studio? (in Release mode). Looking at the Jenkin's performance pipeline results there doesn't seem to be much improvement (with the exception of `triad_gcc_a64fx`). There are also 2 regressions...

dANW34V3R (changes requested):
This seems like a very "minimal gains" optimisation e.g. no fundamental algorithmic change, rather less repeated array loads/getter calls in the high level language (which may not happen once compiled). I would need to see actual performance gains before approval. Some comments explaining what is going on also needed (took me a bit of time to work it all through and understand)

jj16791:
#rerun tests

FinnWilkinson:
Closing as stale / not planned

PR #370, opened 2024-01-16
Config file bug cleanup

JosephMoore25:
This PR aims to squash the bugs being investigated that appear through certain parameter values in the config file.

The current known issues are:

[FIXED] `Vector-Length` being greater than `Load-Bandwidth` or `Store-Bandwidth` causes progression to halt with no error. A simple check is needed in `ModelConfig.cc` to ensure this is not the case, and to error out if it is the case.

[FIXED] `Fetch-Block-Size` being greater than 128 (despite being a valid value) causes a large memory leak. This was due to the precision of `bufferedBytes_` and `bufferOffset` being 8 bit, as was the `bytesAvailable` parameter of the function `predecode`. This led to an infinite loop of instruction fetches.

[TODO, not in this PR] Setting `GeneralPurpose-Count` or `FloatingPoint/SVE-Count` to a small value above architectural register values in OoO/MicroOp mode causes indefinite stalls due to not being able to rename all destination registers. This is to be handled in a different PR due to this not being a trivial fix.

[DONE] Investigate further parameter values to check for unexpected behaviour. Nothing is noticably wrong at this time.


src/lib/config/ModelConfig.cc
```diff
@@ -886,6 +886,18 @@ void ModelConfig::postValidation() {
   // Record any unlinked port names
   for (const auto& prt : portnames)
     invalid_ << "\t- " << prt << " has no associated reservation station\n";
+
+  // Check for valid LSQ-L1-Interface Load and Store Bandwidth
```
dANW34V3R, line 890: A slightly more descriptive comment would be nice e.g. something along the lines of "Ensure load/store bandwidth is high enough for **vector unit**"

src/lib/config/ModelConfig.cc
```diff
@@ -886,6 +886,18 @@ void ModelConfig::postValidation() {
...
+  if (configTree_["Core"]["Vector-Length"].as<uint16_t>() / 8 >
+      configTree_["LSQ-L1-Interface"]["Load-Bandwidth"].as<uint16_t>()) {
+    invalid_ << "\t- Vector-Length (bits) must be greater than Load-Bandwidth "
+                "(bytes)\n";
+  } else if (configTree_["Core"]["Vector-Length"].as<uint16_t>() / 8 >
```
dANW34V3R, line 895: Should this be a separate if rather than else if? Would allow for both issues to printed to the user on the first attempt. No concerns for performance as this only runs once at the start of the program

dANW34V3R (commented):
Minor comments, but concepts look good

dANW34V3R:
It _could_ be worth adding tests for the maximum and minimum bound of each config option e.g. just run the default program for these options and make sure it runs to completion. This would be far from covering all potential issues as all options are dependent but would omit basic issues in the future. This would probably take a non-trivial amount of time so something that could be added to the CI pipeline. Thoughts @jj16791 @FinnWilkinson @ABenC377 @JosephMoore25 ?

JosephMoore25:
@dANW34V3R Ideally this would be tested. Problem is for ~23 parameters that would be 46 tests. In the case that one breaks, bounds would need to be set on the program. A timeout would be needed in the case of broken progression, and I guess memory leaks _should_ eventually crash - these things may be slightly easier to monitor manually. 

It may be sufficient to do this once before each release - I can provide an internal tool to run these tests automatically.

Our options are:

1) Add tests to the pipeline. Pipeline will become noticably longer, but each push we confirm the bounds of all individual parameters (but can not guarantee dependent parameter combinations still)

2) Use an internal tool to run this test semi-automated, to be run before releases

3) Manually test bounds when changing code that may affect these (especially precision of variables)

We should take a vote/discussion on this, but the outcome is unlikely to change this PR.  @FinnWilkinson @jj16791 @dANW34V3R 

I'm under the impression that 2 should be sufficient and 3 should always be done anyway. These limits are often not touched through normal use of SimEng, but instead when designing extreme CPUs, hence not running into an issue until now.

src/lib/config/ModelConfig.cc
JosephMoore25: Good idea. Implemented and works as expected

src/lib/config/ModelConfig.cc
JosephMoore25: Changed to make a little more clear

dANW34V3R (dismissed):

FinnWilkinson:
> It _could_ be worth adding tests for the maximum and minimum bound of each config option e.g. just run the default program for these options and make sure it runs to completion. This would be far from covering all potential issues as all options are dependent but would omit basic issues in the future. This would probably take a non-trivial amount of time so something that could be added to the CI pipeline. Thoughts @jj16791 @FinnWilkinson @ABenC377 @JosephMoore25 ?

This would be nice to do yes, but I think we would need to use some other benchmark other than the default one to ensure all potential issues are discovered. For example, without running a code using SVE instructions, one of the bugs fixed by this PR wouldn't have been discovered

FinnWilkinson:
> @dANW34V3R Ideally this would be tested. Problem is for ~23 parameters that would be 46 tests. In the case that one breaks, bounds would need to be set on the program. A timeout would be needed in the case of broken progression, and I guess memory leaks _should_ eventually crash - these things may be slightly easier to monitor manually.
> 
> It may be sufficient to do this once before each release - I can provide an internal tool to run these tests automatically.
> 
> Our options are:
> 
> 1. Add tests to the pipeline. Pipeline will become noticably longer, but each push we confirm the bounds of all individual parameters (but can not guarantee dependent parameter combinations still)
> 2. Use an internal tool to run this test semi-automated, to be run before releases
> 3. Manually test bounds when changing code that may affect these (especially precision of variables)
> 
> We should take a vote/discussion on this, but the outcome is unlikely to change this PR. @FinnWilkinson @jj16791 @dANW34V3R
> 
> I'm under the impression that 2 should be sufficient and 3 should always be done anyway. These limits are often not touched through normal use of SimEng, but instead when designing extreme CPUs, hence not running into an issue until now.

We run a semi-automated script on the project before release to check that all exisiting benchmarks still functionally work, so doing this for something else wouldn't be too bad (in my opinion). However, upkeeping such a tool would require a bit more work given it is subject to change much more often than the list of supported benchmarks.

An alternative would be to implement unit / integration tests to ensure all the bounds are functioning correctly - this option would need to be a PR in itself though

src/lib/config/ModelConfig.cc
```diff
@@ -886,6 +886,19 @@ void ModelConfig::postValidation() {
   // Record any unlinked port names
   for (const auto& prt : portnames)
     invalid_ << "\t- " << prt << " has no associated reservation station\n";
+
+  // Ensure LSQ-L1-Interface Load/Store Bandwidth is large enough for vector
+  // unit
```
FinnWilkinson, line 891: This comment doesn't really make sense to me. The load/store bandwidth needs to be large enough to accomodate a full vector load in a single request request; its to do with the vector length, not the vector unit arrangement

src/lib/config/ModelConfig.cc
```diff
@@ -886,6 +886,19 @@ void ModelConfig::postValidation() {
...
+  if (configTree_["Core"]["Vector-Length"].as<uint16_t>() / 8 >
+      configTree_["LSQ-L1-Interface"]["Load-Bandwidth"].as<uint16_t>()) {
+    invalid_ << "\t- Vector-Length (bits) must be greater than Load-Bandwidth "
+                "(bytes)\n";
+  }
+  if (configTree_["Core"]["Vector-Length"].as<uint16_t>() / 8 >
+      configTree_["LSQ-L1-Interface"]["Store-Bandwidth"].as<uint16_t>()) {
+    invalid_ << "\t- Vector-Length (bits) must be greater than Store-Bandwidth "
+                "(bytes)\n";
+  }
```
FinnWilkinson, line 901: Similar checks should also be made for Streaming vector length, as this can a) be larger than the (sve) vector length, ad b) increase the size of Sve vector loads/stores when streaming mode is enabled.

FinnWilkinson (changes requested):

dANW34V3R:
> We run a semi-automated script on the project before release to check that all exisiting benchmarks still functionally work, so doing this for something else wouldn't be too bad (in my opinion). However, upkeeping such a tool would require a bit more work given it is subject to change much more often than the list of supported benchmarks.

Check on release seems like a possible solution to get the tests run but not unnecessarily over checking. I can see a way where it could be automated by checking the values set by functions like `setValueBounds` and `setValueSet` which would mean no manual upkeep.

> An alternative would be to implement unit / integration tests to ensure all the bounds are functioning correctly - this option would need to be a PR in itself though

This would be option 1 I guess as all GTests are run in the CI pipeline on every push to an open PR. Check on release seems like a better option if these tests take some time to run

src/lib/config/ModelConfig.cc
JosephMoore25: I've changed this comment in the latest commit to make this correct and more informative

src/lib/config/ModelConfig.cc
JosephMoore25: Good idea, I've added these checks in the latest commit.

FinnWilkinson (approved):

src/lib/config/ModelConfig.cc
```diff
@@ -910,6 +910,35 @@ void ModelConfig::postValidation() {
     invalid_ << "\t- Only a 'Flat' L1-Instruction-Memory Interface-Type is "
                 "supported. Interface-Type used is "
              << l1iType << "\n";
+
+  // Ensure LSQ-L1-Interface Load/Store Bandwidth is large enough to accomodate
```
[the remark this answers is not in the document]
jj16791: If we do this, then we can move all/some of this logic out of `postValidation` and into the appropriate `setValueBounds` call. 

I say some as there's an argument to be made about contextualising an errored lower bound with the current additions to `invalid_` (if we moved all logic into the `setValueBounds` call, then the user wouldn't be told the lower bound is x because the vector-length is y). There are arguments for both sides though to be honest. Pretty trivial to understand the lower bound when considering our target audience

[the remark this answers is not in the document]
jj16791: If we keep some logic here, then the additions to `invalid` should contain the values of `Vector-Length` and `Store-Bandwidth` rather than just stating the keys

jj16791 (changes requested):

src/lib/config/ModelConfig.cc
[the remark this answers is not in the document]
JosephMoore25: Fixed this, as well as fixed some tests that broke with the new checks added (default values caused invalid CPUs) so all previously working tests still pass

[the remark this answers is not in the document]
JosephMoore25: I get your point. The largest load that seems to crop up in RISC-V after a few tests seems to be 8 bytes (so a minimum lower bound being 8 bytes).

I still think due to the link between SVE/SME and Load/Store bandwidth, even though most users of SimEng _should_ know, the current error message helps. I think the current logic should stay, and the lower bound should be raised.

For RISC V, as above, this is 8 bytes. For aarch64 this is 16 bytes as SVE is always enabled at a minimum of 128 bits.

What are your thoughts in this case of keeping the logic, raising the `setValueBounds` to 8 bytes minimum, and a comment in the docs to explain why this minimum is needed?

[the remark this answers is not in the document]
JosephMoore25: After a discussion in person with @jj16791, this was the chosen way forward and has been implemented in the latest commit.

ABenC377 (dismissed):

jj16791 (changes requested):

src/lib/config/ModelConfig.cc
```diff
@@ -485,13 +485,29 @@ void ModelConfig::setExpectations(bool isDefault) {
 
   expectations_["LSQ-L1-Interface"].addChild(
       ExpectationNode::createExpectation<uint16_t>(32, "Load-Bandwidth"));
-  expectations_["LSQ-L1-Interface"]["Load-Bandwidth"].setValueBounds<uint16_t>(
-      1, UINT16_MAX);
+  // AArch64 requires a vector length of at least 128, requiring a minimum of 16
```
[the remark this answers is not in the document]
JosephMoore25: Good point, fixed this

jj16791 (approved):

FinnWilkinson (approved):

PR #373, opened 2024-01-18
Memory interface type check

### LINKED ISSUE #372: Interface-Type restriction not applied
Both `L1-Data-Memory` and `L1-Instruction-Memory` `Interface-Type` option has restrictions possibly based on the `Simulation-Mode` in use. Namely, only a `Flat` value can be used for `L1-Instruction-Memory` and only a non-`Flat` `L1-Data-Memory` value can be used if the `Simulation-Mode` is set to be `outoforder`.

Additionally, the outoforder tests in the regression test suite should use a `Fixed` `L1-Data-Memory` for a more realistic testing scenario.

jj16791:
This PR applies the restrictions for memory interface types possibly based on the Simulation-Mode in use. It also updates the regression test suite to use a fixed L1D for outoforder tests.

closes #372

### DIFF
--- a/src/lib/config/ModelConfig.cc
+++ b/src/lib/config/ModelConfig.cc
@@ -886,6 +886,29 @@ void ModelConfig::postValidation() {
   // Record any unlinked port names
   for (const auto& prt : portnames)
     invalid_ << "\t- " << prt << " has no associated reservation station\n";
+
+  // Ensure the L1-[Data|Instruction]-Memory:Interface-Type restrictions are
+  // enforced
+  // Currently, only outoforder core types can use non-Flat L1-Data-Memory
+  // interfaces
+  if (configTree_["Core"]["Simulation-Mode"].as<std::string>() !=
+      "outoforder") {
+    std::string l1dType =
+        configTree_["L1-Data-Memory"]["Interface-Type"].as<std::string>();
+    if (l1dType != "Flat")
+      invalid_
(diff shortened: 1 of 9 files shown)


src/lib/config/ModelConfig.cc
```diff
@@ -886,6 +886,29 @@ void ModelConfig::postValidation() {
...
+  // Currently, only outoforder core types can use non-Flat L1-Data-Memory
+  // interfaces
+  if (configTree_["Core"]["Simulation-Mode"].as<std::string>() !=
+      "outoforder") {
+    std::string l1dType =
+        configTree_["L1-Data-Memory"]["Interface-Type"].as<std::string>();
+    if (l1dType != "Flat")
+      invalid_
+          << "\t- Only a 'Flat' L1-Data-Memory Interface-Type may be used "
+             "for a non-outoforder Simulation-Mode. Interface-Type used is "
```
FinnWilkinson, line 901: Rather than "a non-outoforder", have the message print out the simulation mode chosen.

FinnWilkinson (approved):
All looks good - one minor suggestion but won't block on this

src/lib/config/ModelConfig.cc
jj16791: Good idea

FinnWilkinson (approved):

dANW34V3R (approved):
Looks good

JosephMoore25 (approved):
Looks all good, nice work

ABenC377 (approved):

PR #374, opened 2024-01-19
Perceptron predictor

ABenC377:
Introducing a new PerceptronPredictor class as an alternative to the GenericPredictor currently used in all the config files.  Shows an improvement in branch prediction rate for all the benchmarks except for STREAM, which was already being well predicted by the Generic predictor.  There is more work done by this predictor per prediction so it is marginally slower per prediction than the GenericPredictor, but it compensates for this by its improved accuracy meaning that SimEng was faster with it for all but 15 of the benchmarks (72).  These 15 benchmarks are generally already very quick and the regression I observed was at most 3.6% compared to improvements of up to 35% for the other benchmarks.  Average improvements of -5% runtime (percentage) and -8.5% mispredictions (raw).

Benchmark | DEV time | DEV mispredict | PERCEPTRON time | PERCEPTRON mispredict | Performance change (percentage) | Mispredict change (raw)
-- | -- | -- | -- | -- | -- | --
CloverLeaf serial gcc8.3.0 armv8.4 | 251655ms | 35.7% | 161811ms | 25% | **-35%** &#x2705; | **-10.7%** &#x2705;
CloverLeaf serial gcc9.3.0 armv8.4 | 171576ms | 34.5% | 158353ms | 23% | **-7.7%** &#x2705; | **-11.5%** &#x2705;
CloverLeaf serial gcc10.3.0 armv8.4 | 173001ms | 36.7% | 163963ms | 24.8% | **-5.2%** &#x2705; | **-11.9%** &#x2705;
CloverLeaf serial armclang20 armv8.4 | 142201ms | 31.6% | 134777ms | 22.5% | **-5.2%** &#x2705; | **-9.1%** &#x2705;
CloverLeaf openmp gcc8.3.0 armv8.4 | 227772ms | 34.6% | 205901ms | 24.1% | **-9.6%** &#x2705; | **-10.5%** &#x2705;
CloverLeaf openmp gcc9.3.0 armv8.4 | 225624ms | 33.5% | 199719ms | 22.6% | **-11.5%** &#x2705; | **-10.9%** &#x2705;
CloverLeaf openmp gcc10.3.0 armv8.4 | 226401ms | 34.5% | 200722ms | 23.5% | **-11.3%** &#x2705; | **-11%** &#x2705;
CloverLeaf openmp armclang20 armv8.4 | 193129ms | 31.6% | 170551ms | 20.5%| **-11.7%** &#x2705; | **-11.1%** &#x2705;
miniBUDE openmp gcc8.3.0 armv8.4 | 201411ms | 9.96% | 203321ms | 8.64% | **+0.9%** &#x274C; | **-1.3%** &#x2705;
miniBUDE openmp gcc9.3.0 armv8.4 | 201666ms | 9.93% | 202440ms | 8.59% | **+0.4%** &#x274C; | **-1.3%** &#x2705;
miniBUDE openmp gcc10.3.0 armv8.4 | 201324ms | 10% | 202448ms | 8.61% | **+0.6%** &#x274C; | **-1.4%** &#x2705;
miniBUDE openmp armclang20 armv8.4 | 183828ms | 11.6% | 185367ms | 11.4% | **+0.8%** &#x274C; | **-0.2%** &#x2705;
STREAM serial gcc8.3.0 armv8.4 | 74849ms | 0.619% | 77580ms | 0.601% | **+3.6%** &#x274C; | **-0.1%** &#x2705;
STREAM serial gcc9.3.0 armv8.4 | 76025ms | 0.942% | 78402ms | 0.774% | **+3.1%** &#x274C; | **-0.2%** &#x2705;
STREAM serial gcc10.3.0 armv8.4 | 76237ms | 0.654% | 78152ms | 0.838% | **+2.5%** &#x274C; | **+0.2%** &#x274C;
STREAM serial armclang20 armv8.4 | 84461ms | 1.16% | 87023ms | 1.24% | **+3.0%** &#x274C; | **+0.1%** &#x274C;
STREAM openmp gcc8.3.0 armv8.4 | 129391ms | 11.5% | 113201ms | 2.76% | **-12.5%** &#x2705; | **-8.7%** &#x2705;
STREAM openmp gcc9.3.0 armv8.4 | 127899ms | 11% | 115083ms | 2.4% | **-10%** &#x2705; | **-8.6%** &#x2705;
STREAM openmp gcc10.3.0 armv8.4 | 126097ms | 11.4% | 112675ms | 2.75% | **-10.6%** &#x2705; | **-8.6%** &#x2705;
STREAM openmp armclang20 armv8.4 | 131196ms | 14.4% | 123741ms | 5.07% | **-5.7%** &#x2705; | **-9.3%** &#x2705;
TeaLeaf 2D serial gcc8.3.0 armv8.4 | 128641ms | 24.9% | 126449ms | 20.7% | **-1.7%** &#x2705; | **-4.2%** &#x2705;
TeaLeaf 2D serial gcc9.3.0 armv8.4 | 127964ms | 25% | 126731ms | 21.6% | **-1.0%** &#x2705; | **-3.6%** &#x2705;
TeaLeaf 2D serial gcc10.3.0 armv8.4 | 129052ms | 24.9% | 128624ms | 20.8% | **-0.3%** &#x2705; | **-4.1%** &#x2705;
TeaLeaf 2D serial armclang20 armv8.4 | 233988ms | 13.4% | 236103ms | 10.9% | **+0.9%** &#x274C; | **-2.5%** &#x2705;
TeaLeaf 2D openmp gcc8.3.0 armv8.4 | 217511ms | 27.6% | 182287ms | 11.2% | **-16.2%** &#x2705; | **-16.4%** &#x2705;
TeaLeaf 2D openmp gcc9.3.0 armv8.4 | 214477ms | 25.4% | 183910ms | 11.7% | **-14.3%** &#x2705; | **-11.1%** &#x2705;
TeaLeaf 2D openmp gcc10.3.0 armv8.4 | 216781ms | 27.6% | 183236ms | 11.5% | **-15.5%** &#x2705; | **-16.1%** &#x2705;
TeaLeaf 2D openmp armclang20 armv8.4 | 598907ms | 16.5% | 585522ms | 9.29% | **-2.2%** &#x2705; | **-7.2%** &#x2705;
TeaLeaf 3D serial gcc8.3.0 armv8.4 | 153938ms | 17.3% | 146189ms | 11.2% | **-5.0%** &#x2705; | **-6.1%** &#x2705;
TeaLeaf 3D serial gcc9.3.0 armv8.4 | 158193ms | 18.6% | 152422ms | 12.5% | **-3.6%** &#x2705; | **-6.1%** &#x2705;
TeaLeaf 3D serial gcc10.3.0 armv8.4 | 156685ms | 17.9% | 152368ms | 12.9% | **-2.8%** &#x2705; | **-5.0%** &#x2705;
TeaLeaf 3D serial armclang20 armv8.4 | 220240ms | 29.2% | 216672ms | 22.4% | **-1.6%** &#x2705; | **-6.8%** &#x2705;
TeaLeaf 3D openmp gcc8.3.0 armv8.4 | 297986ms | 27.9% | 240845ms | 9.66% | **-19.2%** &#x2705; | **-18.2%** &#x2705;
TeaLeaf 3D openmp gcc9.3.0 armv8.4 | 307134ms | 28.5% | 248632ms | 11.1%| **-19.0%** &#x2705; | **-17.4%** &#x2705;
TeaLeaf 3D openmp gcc10.3.0 armv8.4 | 300711ms | 28% | 242736ms | 10.5% | **-19.3%** &#x2705; | **-17.5%** &#x2705;
TeaLeaf 3D openmp armclang20 armv8.4 | 489231ms | 28.8% | 449169ms | 17.3% | **-8.2%** &#x2705; | **-11.5%** &#x2705;
CloverLeaf serial gcc8.3.0 armv8.4+sve | 169930ms | 35.2% | 150743ms | 24.3% | **-11.3%** &#x2705; | **-10.9%** &#x2705;
CloverLeaf serial gcc9.3.0 armv8.4+sve | 166786ms | 34.2% | 149215ms | 22.9% | **-10.5%** &#x2705; | **-11.3%** &#x2705;
CloverLeaf serial gcc10.3.0 armv8.4+sve | 171339ms | 36.5% | 149683ms | 25.2% | **-12.6%** &#x2705; | **-11.3%** &#x2705;
CloverLeaf serial armclang20 armv8.4+sve | 159721ms | 33% | 137543ms | 21.7% | **-13.9%** &#x2705; | **-11.3%** &#x2705;
CloverLeaf openmp gcc8.3.0 armv8.4+sve | 225588ms | 34.3% | 198457ms | 23.6% | **-12.0%** &#x2705; | **-10.7%** &#x2705;
CloverLeaf openmp gcc9.3.0 armv8.4+sve | 221890ms | 33.4% | 194197ms | 22.3% | **-12.5%** &#x2705; | **-11.1%** &#x2705;
CloverLeaf openmp gcc10.3.0 armv8.4+sve | 223602ms | 34.1% | 195536ms | 23.4% | **-12.6%** &#x2705; | **-10.7%** &#x2705;
CloverLeaf openmp armclang20 armv8.4+sve | 206803ms | 32.5% | 179017ms | 21.6% | **-13.4%** &#x2705; | **-10.9%** &#x2705;
miniBUDE openmp gcc8.3.0 armv8.4+sve | 84149ms | 23.6% | 80624ms | 16.5% | **-4.2%** &#x2705; | **-7.1%** &#x2705;
miniBUDE openmp gcc9.3.0 armv8.4+sve | 81961ms | 25.1% | 77745ms | 16.7% | **-5.1%** &#x2705; | **-8.4%** &#x2705;
miniBUDE openmp gcc10.3.0 armv8.4+sve | 80809ms | 24% | 77152ms | 16.5% | **-4.5%** &#x2705; | **-7.5%** &#x2705;
miniBUDE openmp armclang20 armv8.4+sve | 80084ms | 22.9% | 80877ms | 22.6% | **+1.0%** &#x274C; | **-0.3%** &#x2705;
STREAM serial gcc8.3.0 armv8.4+sve | 40075ms | 1.68% | 40545ms | 1.9% | **+1.2%** &#x274C; | **+0.2%** &#x274C;
STREAM serial gcc9.3.0 armv8.4+sve | 40034ms | 1.84% | 41154ms | 2.1% | **+2.8%** &#x274C; | **+0.3%** &#x274C;
STREAM serial gcc10.3.0 armv8.4+sve | 39628ms | 2.04% | 40867ms | 2.02% | **+3.1%** &#x274C; | **-0.0%** &#x2705;
STREAM serial armclang20 armv8.4+sve | 24085ms | 2.13% | 24842ms | 1.88% | **+3.1%** &#x274C; | **-0.2%** &#x2705;
STREAM openmp gcc8.3.0 armv8.4+sve | 91458ms | 19.5% | 77394ms | 5.01% | **-15.4%** &#x2705; | **-14.4%** &#x2705;
STREAM openmp gcc9.3.0 armv8.4+sve | 91222ms | 19.7% | 78769ms | 6.66% | **-13.7%** &#x2705; | **-13%** &#x2705;
STREAM openmp gcc10.3.0 armv8.4+sve | 89951ms | 19.5% | 77005ms | 5.41% | **-14.4%** &#x2705; | **-15.1%** &#x2705;
STREAM openmp armclang20 armv8.4+sve | 76133ms | 18.3% | 65194ms | 5.69% | **-14.4%** &#x2705; | **-12.6%** &#x2705;
TeaLeaf 2D serial gcc8.3.0 armv8.4+sve | 130373ms | 25% | 128608ms | 21.3% | **-1.4%** &#x2705; | **-3.7%** &#x2705;
TeaLeaf 2D serial gcc9.3.0 armv8.4+sve | 129521ms | 24.9% | 127291ms | 21.7% | **-1.7%** &#x2705; | **-3.2%** &#x2705;
TeaLeaf 2D serial gcc10.3.0 armv8.4+sve | 131156ms | 25% | 128176ms | 20.7% | **-2.3%** &#x2705; | **-4.3%** &#x2705;
TeaLeaf 2D serial armclang20 armv8.4+sve | 99518ms | 19.5% | 95118ms | 14.2% | **-4.4%** &#x2705; | **-5.3%** &#x2705;
TeaLeaf 2D openmp gcc8.3.0 armv8.4+sve | 217552ms | 27.4% | 185782ms | 11.6% | **-14.6%** &#x2705; | **-15.6%** &#x2705;
TeaLeaf 2D openmp gcc9.3.0 armv8.4+sve | 214994ms | 26.5% | 192862ms | 11.7% | **-10.3%** &#x2705; | **-14.8%** &#x2705;
TeaLeaf 2D openmp gcc10.3.0 armv8.4+sve | 213683ms | 27.5% | 180638ms | 11.9% | **-15.5%** &#x2705; | **-15.7%** &#x2705;
TeaLeaf 2D openmp armclang20 armv8.4+sve | 822471ms | 16.4% | 576071ms | 8.79% | **-30.0%** &#x2705; | **-7.6%** &#x2705;
TeaLeaf 3D serial gcc8.3.0 armv8.4+sve | 130973ms | 26.4% | 130850ms | 20.6% | **-0.1%** &#x2705; | **-5.8%** &#x2705;
TeaLeaf 3D serial gcc9.3.0 armv8.4+sve | 131887ms | 26.4% | 131509ms | 20.8% | **-0.3%** &#x2705; | **-5.6%** &#x2705;
TeaLeaf 3D serial gcc10.3.0 armv8.4+sve | 132279ms | 26.1% | 130915ms | 21.7% | **-1.0%** &#x2705; | **-4.4%** &#x2705;
TeaLeaf 3D serial armclang20 armv8.4+sve | 219685ms | 29.9% | 212249ms | 23.1%| **-3.4%** &#x2705; | **-6.8%** &#x2705;
TeaLeaf 3D openmp gcc8.3.0 armv8.4+sve | 269587ms | 31.3% | 228886ms | 14.1% | **-15.1%** &#x2705; | **-16.2%** &#x2705;
TeaLeaf 3D openmp gcc9.3.0 armv8.4+sve | 273317ms | 30.7% | 234772ms | 14.3% | **-14.1%** &#x2705; | **-16.4%** &#x2705;
TeaLeaf 3D openmp gcc10.3.0 armv8.4+sve | 273946ms | 32.1% | 226325ms | 15.5% | **-17.4%** &#x2705; | **-16.6%** &#x2705;
TeaLeaf 3D openmp armclang20 armv8.4+sve | 530223ms | 30.8% | 481883ms | 17.4% | **-9.1%** &#x2705; | **-13.4%** &#x2705;


src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,153 @@
...
+PerceptronPredictor::~PerceptronPredictor() {
+  ras_.clear();
+  rasHistory_.clear();
+}
+
+BranchPrediction PerceptronPredictor::predict(uint64_t address, BranchType type,
+                                              int64_t knownOffset) {
+  // Get index via an XOR hash between the global history and the lower btbBits_
+  // bits of the instruction address
+  uint64_t hashedIndex = ((address >> 2) ^ globalHistory_) & ((1 << btbBits_) - 1);
```
dANW34V3R, line 32: This comment and subsequent line has confused me a bit. Not sure if "XOR hash" is a phrase (may well be) could replace with just "XOR". "btbBits_" could be quoted e.g. 'btbBits_', to make it clear that it is the _number_ of lower bits. If I have then interpreted the comment correctly I don't think it aligns with the code which is first XOR the address with the history then take the lower bits as opposed to what the comment suggests which would be (address & lower bits) ^ globalHistory_. These are not logically equivalent, but you probably get away with it as it is deterministic. 

With a little bit more thinking I reckon your code is correct but the comment isn't. You want the hash from the xor between the history and the address but you don't want to use this as an index as it could be out of range of the btb_ so you take the lower bits to ensure it is within the bounds. Is that the idea here?

src/include/simeng/PerceptronPredictor.hh
```diff
@@ -0,0 +1,79 @@
...
+  void flush(uint64_t address) override;
+
+ private:
+  /** The bitlength of the BTB index; BTB will have 2^bits entries. */
+  uint64_t btbBits_;
+
+  /** A 2^bits length vector of pairs containing a (globalHistoryLength_ + 1)
+   * length perceptron and a branch target. */
+  std::vector<std::pair<std::vector<int8_t>, uint64_t>> btb_;
+
```
dANW34V3R, line 52: Is branch target buffer the best name for this structure? "Table of perceptrons" therefor "top_" from the paper may be more descriptive. 

This also relates to the config options - other than the "Type" option, it isn't obvious how the other options are effecting anything to do with perceptrons e.g. "BTB-Tag-Bits" is setting the size of the "top_" and therefor the number of perceptrons in use. In the case of DEMO_RISCV that is 2^11=2048 perceptrons which isn't obvious when just looking at the config.

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,153 @@
...
+
+  // Retrieve the perceptron from the BTB
+  std::vector<int8_t> perceptron = btb_[hashedIndex].first;
+
+  // Determine direction prediction from perceptron by taking its dot product
+  // with the global history
+  // Starting with the bias weight
+  int64_t Pout = perceptron[globalHistoryLength_];
+  for (int i = 0; i < globalHistoryLength_; i++) {
+    bool historyTaken =
```
dANW34V3R, line 47: Short comment explaining this line e.g. get branch direction for ith entry in the global history

src/include/simeng/PerceptronPredictor.hh
dANW34V3R: Maybe this should be separated out into a top_ and a btb_. btb_[n].second more aligns with the idea of a branch target buffer

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,153 @@
...
+  }
+  bool direction = (Pout >= 0);
+  // Retrieve target prediction
+
+  uint64_t target =
+      (knownOffset != 0) ? address + knownOffset : btb_[hashedIndex].second;
+
+  BranchPrediction prediction = {direction, target};
+
+  // Ammend prediction based on branch type
```
dANW34V3R, line 59: Typo "amend"

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,153 @@
...
+  int64_t Pout = perceptron[globalHistoryLength_];
+  for (int i = 0; i < globalHistoryLength_; i++) {
+    bool historyTaken =
+        ((prevGlobalHistory & (1 << ((globalHistoryLength_ - 1) - i))) != 0);
+    Pout += historyTaken ? perceptron[i] : (0 - perceptron[i]);
+  }
+  bool directionPrediction = (Pout >= 0);
+
+  // Determine the magnitude of the dot product for training
+  uint64_t magnitude = (Pout < 0) ? (0 - Pout) : Pout;
```
dANW34V3R, line 105: `abs(Pout)` would be more clear here

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,153 @@
...
+
+  // update the perceptron if the prediction was wrong, or the dot product's magnitude
```
dANW34V3R, line 107: Capital "U"

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,153 @@
...
+  // was not greater than the training threshold
+  if ((directionPrediction != taken) || (magnitude < trainingThreshold_)) {
+    int8_t t = (taken) ? 1 : -1;
+
+    for (int i = 0; i < globalHistoryLength_; i++) {
+      int8_t xi =
+          ((prevGlobalHistory & (1 << ((globalHistoryLength_ - 1) - i))) == 0) ? -1 : 1;
+      int8_t product_xi_t = xi * t;
+      // Make sure no overflow
+      perceptron[i] += (perceptron[i] != 127 && perceptron[i] != -127) ? product_xi_t : 0;
```
dANW34V3R, line 117: I'm not sure this check is fully correct. If `perceptron[i]` ever gets to the bounds; -127 or 127, then it will be stuck there, unable to update. You want it to be able to get to, say, -127 but then also be able to +1 to move away from that

dANW34V3R (changes requested):
The implementation on the whole looks great. Just some comments on clarity and naming, again for clarity

configs/DEMO_RISCV.yaml
```diff
@@ -26,11 +26,10 @@ Queue-Sizes:
...
   Store: 36
 Branch-Predictor:
+  Type: "Perceptron"
   BTB-Tag-Bits: 11
-  Saturating-Count-Bits: 2
-  Global-History-Length: 10
-  RAS-entries: 5
-  Fallback-Static-Predictor: "Always-Taken"
+  Global-History-Length: 19
+  RAS-entries: 1
```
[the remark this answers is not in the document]
FinnWilkinson: I think this should be a point of discussion - if this predictor improves branch prediction rates, sometimes considerably, then why would we want to use a worse predictor?

src/lib/config/ModelConfig.cc
```diff
@@ -424,13 +424,19 @@ void ModelConfig::setExpectations(bool isDefault) {
...
       ExpectationNode::createExpectation<uint8_t>(8, "BTB-Tag-Bits"));
   expectations_["Branch-Predictor"]["BTB-Tag-Bits"].setValueBounds<uint8_t>(1,
                                                                             64);
-
   expectations_["Branch-Predictor"].addChild(
-      ExpectationNode::createExpectation<uint8_t>(2, "Saturating-Count-Bits"));
+      ExpectationNode::createExpectation<uint8_t>(2, "Saturating-Count-Bits",
+                                                  true));
   expectations_["Branch-Predictor"]["Saturating-Count-Bits"]
       .setValueBounds<uint8_t>(1, 64);
```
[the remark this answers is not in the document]
FinnWilkinson: I'm not sure how I feel about the first comment - given all config options have a default value now it doesn't matter if extra are set / if they're not set. Although it could be useful to determine exactly what should and shouldn't be set.... (thinking out loud on this one)

I disagree with the second point - this is what documentation is for and if they are redunant then they won't be used in simulation anyway. Perhaps we need to update the current documentation to make it clearer regarding what an emulation core does / will use from the config

FinnWilkinson:
Please can you update the documentation with the new config options and a small section about how the new predictor works

src/include/simeng/PerceptronPredictor.hh
ABenC377: This language is carried over from the GenericPredictor, which also uses "btb_" to describe a table containing the targets and the prediction mechanism (in that case a saturating counter).  I think separating it out into two different tables feels clunky as they would both be indexed with the same hash and the pair of data currently in each entry is symmetrical to the pair of data provided in the prediction object that is returned by predict() (i.e. both contain direction information and target information).  However, very happy to switch up the naming -- maybe something like predictorTable_?

src/lib/PerceptronPredictor.cc
ABenC377: Amazing, didn't know that existed!

src/lib/PerceptronPredictor.cc
ABenC377: Agreed, good catch

configs/DEMO_RISCV.yaml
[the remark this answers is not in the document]
dANW34V3R: I guess whatever is most accurate to the model

rahahahat:
I was just wondering if it would be possible to add the sources (research papers/ websites) you've used to carry out the implementation. 

This could help reviewers and is a good piece of documentation for future changes.

src/lib/config/ModelConfig.cc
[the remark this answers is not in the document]
ABenC377: Happy with either option here.  Subject to further discussion, I've added to the documentation for the BP config options that Saturating-Count-Bits and Fallback-Static-Predictor are only needed for the GenericPredictor.

src/include/simeng/PerceptronPredictor.hh
dANW34V3R: The reason _for_ separating into 2 different structures is that it would more accurately represent the hardware as a BTB is simply a cache for the target addresses and nothing else. This would also align the implementation more to the literature making it easier to understand. 

The reason _against_ separating is a simpler, "less clunky" and potentially slightly quicker SimEng implementation (although would need to verify that) so I can absolutely see why it is done that way. 

We don't necessarily need to align with hardware at this level so I am fine with the way it is implemented (e.g. 1 vector containing prediction mechanism and target), however I think the name is confusing when the structure is doing more than what a BTB (in the literature) is supposed to do. In that regard I am also not fully comfortable with the name in the generic predictor either. Interested in @jj16791 's opinion as he did this implementation.

"predictorTable_" could be a better name which could also be used in the generic predictor.

JosephMoore25:
I suspect these tests were run in debug mode. It's great that you have lots of results, but it may be worth doing a couple comparative spot checks in release mode to ensure that the speedup is consistent between the two (misprediction rate should remain the same), as this is the mode that performance will matter the most. No need to rerun all the tests unless differences are noticed, or if this is already in release.

src/lib/config/ModelConfig.cc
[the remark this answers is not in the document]
jj16791: @dANW34V3R The first point could go either way. There's no programming or performance advantage in having it be more conditional but it could help with code clarity and is technically more "correct". The second point, outputting the redundant variables on each run wouldn't be a viable solution as it's just too messy in the STDOUT. A solution similar to Finn's comment of moving it over to the docs would be a better solution currently.

@JosephMoore25 I like the idea of a matrix. It would have to be a 4-column, n-row (with n being the number of config options) matrix to be consistently readable. The inclusion of ranges in the docs would be advantageous. Also, the matrix idea helps with giving the user knowledge of what is needed before running SimEng (similar to your comment on the limits)

src/include/simeng/PerceptronPredictor.hh
FinnWilkinson: Having read the docs Alex wrote, I think BTB is fine. I get your point on this @dANW34V3R, but given our usage of the BTB still "caches" targets (plus some more information) it lives up to its name.

I'm not fussed either way if its just a name change (I don't feel as though its required, but I'm not against it), but I don't agree with splitting this into 2 structures

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,159 @@
...
+  // Initialise perceptron values with 0 for the global history weights, and 1 for the bias weight;
+  // and intialise the target with 0 (i.e., unknown)
+  for (int i = 0; i < btbSize; i++) {
+    btb_[i].first.assign(globalHistoryLength_, 0);
+    btb_[i].first.push_back(1);
+    btb_[i].second = 0;
+  }
+
+  // Set up training threshold according to empirically determined formula
+  trainingThreshold_ = (int)((1.93 * globalHistoryLength_) + 14);
```
FinnWilkinson, line 22: if `trainingThreshold_` is a uint64_t, why does the RHS of this need to be interpreted as an int? Should `trainingThreshold_` be of type int, or can we change `(int)` to `(uint64_t)`?

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,159 @@
...
+PerceptronPredictor::~PerceptronPredictor() {
+  ras_.clear();
+  rasHistory_.clear();
+}
+
+BranchPrediction PerceptronPredictor::predict(uint64_t address, BranchType type,
+                                              int64_t knownOffset) {
+  // Get the hashed index for the prediction table.  XOR the global history with the
+  // non-zero bits of the address, and then keep only the btbBits_ bits of the output
+  // keep it in bounds of the prediction table.
```
FinnWilkinson, line 34: "and then keep only the btbBits_ bits of the output keep it in bounds of the prediction table." doesn't flow well. I think its missing a word or punctuation?

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,159 @@
...
+  uint64_t hashedIndex =
+      ((address >> 2) ^ prevGlobalHistory) & ((1 << btbBits_) - 1);
+
+  std::vector<int8_t> perceptron = btb_[hashedIndex].first;
+
+  // Work out the most recent prediction
+  int64_t Pout = perceptron[globalHistoryLength_];
+  for (int i = 0; i < globalHistoryLength_; i++) {
+    bool historyTaken =
+        ((prevGlobalHistory & (1 << ((globalHistoryLength_ - 1) - i))) != 0);
```
FinnWilkinson, line 103: Similar comment needed here as in predict()

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,159 @@
...
+    Pout += historyTaken ? perceptron[i] : (0 - perceptron[i]);
+  }
+  bool directionPrediction = (Pout >= 0);
```
FinnWilkinson, line 106: This section of code is identical to that in predict(). You could create a new private function to reduce repeated code

test/unit/PerceptronPredictorTest.cc
```diff
@@ -0,0 +1,148 @@
...
+class PerceptronPredictorTest : public testing::Test {
+ public:
+  PerceptronPredictorTest() : uop(new MockInstruction), uopPtr(uop) {
+    uop->setInstructionAddress(0);
+  }
+
+ protected:
+  MockInstruction* uop;
+  std::shared_ptr<Instruction> uopPtr;
+};
```
FinnWilkinson, line 16: Please add a unit test for the flush() function. An example for the GenericPredictor can be found in the unit test PR

FinnWilkinson (changes requested):

configs/DEMO_RISCV.yaml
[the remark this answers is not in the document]
dANW34V3R: As long as it is closer to hardware results (not just better mispredict rates) then I'm happy for it to stay. Does @jj16791 have a table of hardware results for these codes?

src/include/simeng/PerceptronPredictor.hh
jj16791: So my opinions on both matters raised are the following.

I agree with Alex regarding not splitting into two different tables and the supporting point Dan made about not having to align with hardware is also spot on regarding SimEng's overall view of things like this.

As for the naming, both structures in our two Branch Predictors do represent Branch Target Buffers but with different predicting logic. BTB's are a very common and accessible structure regarding the understanding of its functionality so it's a good naming convention to use here. The generic predictor's saturating counter will also be well-understood so I don't think much needs to change there. The perception usage here is a bit less familiar at large in the domain we're in so perhaps some more verbose commenting above btb_ would be beneficial. We do link in the resources to fully understand the Branch Predictor at a detailed level so any additional information should just give a high-level overview of terminology/algorithms in use rather than in-depth detailings.

configs/DEMO_RISCV.yaml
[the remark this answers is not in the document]
jj16791: Sent offline

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,159 @@
...
+  // Update the perceptron if the prediction was wrong, or the dot product's magnitude
+  // was not greater than the training threshold
+  if ((directionPrediction != taken) || (magnitude < trainingThreshold_)) {
+    int8_t t = (taken) ? 1 : -1;
+
+    for (int i = 0; i < globalHistoryLength_; i++) {
+      int8_t xi =
+          ((prevGlobalHistory & (1 << ((globalHistoryLength_ - 1) - i))) == 0) ? -1 : 1;
+      int8_t product_xi_t = xi * t;
+      // Make sure no overflow
```
JosephMoore25, line 120: Perhaps change the comment to "Make sure no overflow, and cap out perceptron weight to +- 127". Took me a second to think through the logic of what was happening if we do say reach 127 with a +1 movement, and this would clear it up.

JosephMoore25 (changes requested):
Looks good overall, nice work.

A couple small comments, although others' reviews mostly encapsulate changes that need to happen before approval.

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,153 @@
...
+
+PerceptronPredictor::PerceptronPredictor(ryml::ConstNodeRef config)
+    : btbBits_(config["Branch-Predictor"]["BTB-Tag-Bits"].as<uint64_t>()),
+      globalHistoryLength_(
+          config["Branch-Predictor"]["Global-History-Length"].as<uint64_t>()),
+      rasSize_(config["Branch-Predictor"]["RAS-entries"].as<uint64_t>()) {
+  // Build BTB based on config options
+  uint32_t btbSize = (1 << btbBits_);
+  btb_.resize(btbSize);
+  for (int i = 0; i < btbSize; i++) {
```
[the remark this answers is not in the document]
dANW34V3R: Typo "initialise"

src/lib/PerceptronPredictor.cc
```diff
@@ -0,0 +1,153 @@
...
+    btb_[i].first.assign(globalHistoryLength_, 0);
+    btb_[i].first.push_back(1);
+    btb_[i].second = 0;
+  }
+
+  // Set up training threshold according to empirically determined formula
+  trainingThreshold_ = (int)((1.93 * (globalHistoryLength_)) + 14);
```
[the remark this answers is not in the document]
dANW34V3R: Inner most brackets are still not necessary as * has a higher precedence than +, but resolving as legible now

src/lib/PerceptronPredictor.cc
dANW34V3R: "output **to** keep"?

configs/DEMO_RISCV.yaml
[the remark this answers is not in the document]
dANW34V3R: Hardware results all below 5%; we are still well above in all cases. Happy to leave change as is. Once we do get to the similar misprediction rates, it may be important to use model accurate predictors to get the correct behaviour

src/lib/config/ModelConfig.cc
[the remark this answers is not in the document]
dANW34V3R: My main (first) point here was about correctness and clarity rather than anything else. It would function yes, but a condition would make it clear to a dev.

Ranges/ valid options should be a given in the docs I think. I can create an issue if we all agree so that we don't forget to do it. 

Another idea to help this could be to add a skeleton config with all possible options and ranges either to the docs or the configs folder. Potentially this can be generated from the expectations so no upkeep needed and should make it very easy for a new user to create a new config and get started straight away.

A matrix could be nice, potentially might be quite large but would need to just try it and see. We can discuss the exact format in the meeting tomorrow.

src/include/simeng/PerceptronPredictor.hh
dANW34V3R: I'm still in favour of a name change e.g. predictorTable_ as Alex suggests as getting the prediction mechanism from a buffer containing branch targets doesn't make sense. Can discuss fully in the meeting also to save time here?

src/lib/config/ModelConfig.cc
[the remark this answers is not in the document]
dANW34V3R: Offline discussion lead to decision to include `if`s in model config. I will resolve this comment once done

No offline discussion but it seems there is agreement to add ranges and potentially a matrix to the docs. This shouldn't be part of this PR. I will create an issue summarising this

src/include/simeng/PerceptronPredictor.hh
dANW34V3R: Offline discussion agreement to keep the name but use paper accurate language to the comments and docs. Will resolve once complete

src/lib/PerceptronPredictor.cc
ABenC377: I prefer the private function -- in reality I don't think that the dot product is stored between prediction and update (to reduce hardware cost)

src/include/simeng/PerceptronPredictor.hh
```diff
@@ -43,11 +44,27 @@ class PerceptronPredictor : public BranchPredictor {
   void flush(uint64_t address) override;
 
  private:
-  /** The bitlength of the BTB index; BTB will have 2^bits entries. */
+  /** Returns the dot product of a perceptron and a history vector.  Used to determine
+   * a direction prediction */
+  int64_t getDotProduct(std::vector<int8_t> perceptron, uint64_t history) {
```
FinnWilkinson, line 49: its probably worth passing the `perceptron` vector as a const reference given we only read from it. It won't effect performance too much either way,  but still worth changing in my opinion

Also, for codebase consistency, could you only have the function declaration in the header file and then define the function in the `.cc`

test/regression/RegressionTest.cc
```diff
@@ -92,7 +93,15 @@ void RegressionTest::run(const char* source, const char* triple,
       createPortAllocator();
 
   // Create a branch predictor for a pipelined core
-  simeng::GenericPredictor predictor = simeng::GenericPredictor();
+  simeng::PerceptronPredictor predictor = simeng::PerceptronPredictor();
```
FinnWilkinson, line 96: This line of code seems redundant

FinnWilkinson (changes requested):

test/regression/RegressionTest.cc
ABenC377: Yep, very much so.  This was a remnant of the process of fixing the YAML config issue I had

src/lib/config/ModelConfig.cc
```diff
@@ -424,15 +424,24 @@ void ModelConfig::setExpectations(bool isDefault) {
...
       ExpectationNode::createExpectation<uint8_t>(8, "BTB-Tag-Bits"));
   expectations_["Branch-Predictor"]["BTB-Tag-Bits"].setValueBounds<uint8_t>(1,
                                                                             64);
-
-  expectations_["Branch-Predictor"].addChild(
-      ExpectationNode::createExpectation<uint8_t>(2, "Saturating-Count-Bits"));
-  expectations_["Branch-Predictor"]["Saturating-Count-Bits"]
-      .setValueBounds<uint8_t>(1, 64);
+  // Saturating counter bits are relevant to the GenericPredictor only
+  if (!isDefault &&
```
jj16791, line 437: We could combine the two Generic-only config options under one conditional to reduce code duplication

jj16791 (approved):

src/include/simeng/PerceptronPredictor.hh
```diff
@@ -0,0 +1,87 @@
...
+  void update(uint64_t address, bool taken, uint64_t targetAddress,
+              BranchType type) override;
+
+  /** Provides RAS rewinding behaviour. */
+  void flush(uint64_t address) override;
+
+ private:
+  /** Returns the dot product of a perceptron and a history vector.  Used to
+   * determine a direction prediction */
+  int64_t getDotProduct(const std::vector<int8_t> perceptron, uint64_t history);
```
FinnWilkinson, line 49: Could `perceptron` be passed as `const std::vector<int8_t>& perceptron`

FinnWilkinson (changes requested):

src/include/simeng/PerceptronPredictor.hh
ABenC377: yes, it could.  Changed.

FinnWilkinson (dismissed):
All looks great

jj16791 (approved):

docs/sphinx/developer/components/branchPred.rst
```diff
@@ -35,3 +35,20 @@ Return Address Stack (RAS)
...
+
+Branch Target Buffer (BTB)
+    For each entry, the BTB stores the most recent target along with a perceptron for an associated direction. The indexing of this structure uses the lower, non-zero bits of an instruction address XOR'ed with the current global branch history value.
+
+    The direction prediction is obtained from the perceptron by taking its dot-product with the global history.  The prediction is not taken if this is negative, or taken otherwise.  The perceptron is updated when its prediction is wrong or when the magnitude of the dot-product is below a pre-determined threshold (i.e., the confidence of the prediction is low).  To update, each ith weight of the perceptron is incremented if the actual outcome of the branch is the same as the ith bit of ``globalHistory_``, and decremented otherwise.
+
+    If the supplied branch type is ``Unconditional``, then the predicted direction is overridden to be taken. If the supplied branch type is ``Conditional`` and the predicted direction is not taken, then the predicted target is overridden to be the next sequential instruction.
```
dANW34V3R, line 51: I think "the predicted target is _overridden_ to be the next sequential instruction" is an odd wording. Would something along the lines of "set" be more appropriate as this is precisely the intended behaviour of a BP.

docs/sphinx/user/configuring_simeng.rst
```diff
@@ -145,20 +145,23 @@ The Branch-Prediction section contains those options to parameterise the branch
 
 The current options include:
 
+Type
+    The type of branch predictor that is used, the options are ``Generic``, and ``Perceptron``.  Both types of predictor use a branch target buffer with each entry containing a direction prediction mechanism and a target address.  The direction predictor used in ``Generic`` is a saturating counter, and in ``Perceptron`` it is a perceptron.
+
 BTB-Tag-Bits
     The number of bits used to denote an entry in the Branch Target Buffer (BTB). For example, a ``bits`` value of 12 could denote 4096 entries with the calculation 1 << ``bits``.
```
dANW34V3R, line 152: I spoke with Jack about this section as I think it is a bit confusing. It would make sense to put it in this PR after agreement on wording. My suggested change is: "The number of bits used to represent a tag/index for the BTB. This therefore limits the size of the BTB to 2^bits entries. SimEng will use a BTB with the maximum number of entries based on this value. For example, a value of 12 means a BTB with 4096 entries will be used."

@jj16791 @FinnWilkinson @JosephMoore25 @ABenC377

src/include/simeng/PerceptronPredictor.hh
```diff
@@ -0,0 +1,88 @@
...
+                        uint64_t history);
+
+  /** The length in bits of the BTB index; BTB will have 2^bits entries. */
+  uint64_t btbBits_;
+
+  /** A 2^bits length vector of pairs containing a perceptron with
+   * globalHistoryLength_ + 1 inputs, and a branch target.
+   * The perceptrons are used to provide a branch direction prediction by
+   * taking a dot product with the global history, as described
+   * in Jiminez and Lin */
```
dANW34V3R, line 59: Add "This structure is analogous the the table of perceptrons as described in Jiminez and Lin"

docs/sphinx/developer/components/branchPred.rst
```diff
@@ -35,3 +35,20 @@ Return Address Stack (RAS)
...
+
+Return Address Stack (RAS)
```
dANW34V3R, line 53: Replace "a proceeding" with "the proceeding" or "the corresponding"

dANW34V3R (changes requested):

docs/sphinx/developer/components/branchPred.rst
ABenC377: I've used the verbatim language from the Generic Predictor documentation.  While I agree that this doesn't mean it is correct, per se, I think that in this case it is good.  Overridden emphasises that not only is the target set, but it replaces the target value that was already there.

src/include/simeng/PerceptronPredictor.hh
ABenC377: I don't think your proposed addition would be more correct.  The table of perceptrons in Jiminez and Lin doesn't include target prediction information, and so the btb isn't analogous to the table of perceptrons.  Rather, I was careful to limit the analogy to the Jiminez and Lin paper to the feature being lifted from it (the perceptrons).

jj16791 (dismissed):

FinnWilkinson (dismissed):

JosephMoore25 (dismissed):
Looks all good now. Nice work, and nice performance improvements 👍

dANW34V3R (approved):

jj16791 (approved):

JosephMoore25 (approved):

FinnWilkinson (approved):

PR #377, opened 2024-01-29
Name changes

dANW34V3R:
This PR updates the names of many variables for code clarity and consistency which are listed below. Some of the updates still need to permeate the docs, but this will happen once the proposed names are approved.

- `operands `-> `sourceValues_` to prevent conflict with metadata.operands which contains both source and destination registers
-`operandsPending` -> `sourceOperandsPending_`
- `getStackPointer` -> `getInitialProcessStackPointer` this name was chosen as to not conflict with Linux.getInitialStackPointer for human readability
- Underscores added throughout architecture specific header files


ABenC377 (dismissed):

t
JosephMoore25, line 1: Guessing this is an accident? I have no clue what this is, and even if it were correct, why `t`?

JosephMoore25 (changes requested):
Simple PR, everything looks in order with what we agreed on.

Looks like the rogue file `t` which is a buggy copy of some regression tests made it in though, so this needs to be removed before approval.

JosephMoore25 (approved):

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -341,24 +341,25 @@ class Instruction : public simeng::Instruction {
   const Architecture& architecture_;
 
   /** A reference to the decoding metadata for this instruction. */
-  const InstructionMetadata& metadata;
+  const InstructionMetadata& metadata_;
 
   /** A vector of source registers. */
-  std::vector<Register> sourceRegisters;
+  std::vector<Register> sourceRegisters_;
+  /** The number of source registers this instruction reads from. */
```
jj16791, line 348: Re-remove sourceRegisterCount_ that's been accidentally added back in

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -341,24 +341,25 @@ class Instruction : public simeng::Instruction {
...
+  uint16_t sourceRegisterCount_ = 0;
 
   /** A vector of destination registers. */
-  std::vector<Register> destinationRegisters;
-
```
jj16791, line 351: Add newline back in

src/include/simeng/arch/riscv/Instruction.hh
```diff
@@ -169,6 +169,7 @@ class Instruction : public simeng::Instruction {
   /** Retrieve the instruction's associated architecture. */
   const Architecture& getArchitecture() const;
 
+ private:
```
jj16791, line 172: Seem like Private member region already defined above unless GitHub is being odd

test/regression/riscv/RISCVRegressionTest.hh
```diff
@@ -4,7 +4,7 @@
 #include "simeng/arch/riscv/Architecture.hh"
 #include "simeng/arch/riscv/Instruction.hh"
 
-static const char* RISCV_ADDITIONAL_CONFIG = R"YAML(
+static constexpr const char* RISCV_ADDITIONAL_CONFIG = R"YAML(
```
jj16791, line 7: Motivation for `constexpr` use? Not sure what compile-time advantage we can get here. Also, do you need the `const` specifier with a `constexpr` specifier?

jj16791 (changes requested):

test/regression/riscv/RISCVRegressionTest.hh
dANW34V3R: Was playing around with quashing warnings with @tom91136 yesterday. Shouldn't have been added to this PR so will remove

src/include/simeng/arch/riscv/Instruction.hh
dANW34V3R: Good spot, removed

jj16791 (approved):

src/include/simeng/kernel/LinuxProcess.hh
```diff
@@ -95,7 +95,7 @@ class LinuxProcess {
   uint64_t getEntryPoint() const;
 
   /** Get the initial stack pointer address. */
-  uint64_t getStackPointer() const;
```
FinnWilkinson, line 98: Couldn't this be `getInitialStackPointer()`?

FinnWilkinson (changes requested):

src/include/simeng/kernel/LinuxProcess.hh
FinnWilkinson: Given they have to be called either by `linux_.getInitianStackPointer` or `process_.getInitialStackPointer` I think it will be evident what it refers to

dANW34V3R: Will the referring to the same value change with multithreading?

FinnWilkinson: They should do yes, but if this changes then said change should be made in the relevant multi-threading PR

FinnWilkinson (approved):

jj16791 (approved):

FinnWilkinson (approved):

JosephMoore25 (approved):

ABenC377 (approved):

PR #381, opened 2024-02-02
Added fix for returning a consistent nan value from a divide by zero

### LINKED ISSUE #380: Fix for NaN value from vector divide by 0 in AArch64 namespace
At a bit-level the value of a NaN can differ between host systems (e.g. with the sign bit set and unset) so a fix is needed to consistently return the same NaN value across varying host systems

jj16791:
This PR aims to fix the issues of differing bit-level values for a NaN value return from a vector divide by 0.

closes #380

### DIFF
--- a/src/lib/arch/aarch64/Instruction_execute.cc
+++ b/src/lib/arch/aarch64/Instruction_execute.cc
@@ -1653,14 +1653,22 @@ void Instruction::execute() {
       }
       case Opcode::AArch64_FDIVR_ZPmZ_D: {  // fdivr zdn.d, pg/m, zdn.d, zm.d
         results[0] = sveLogicOpPredicated_3vecs<double>(
-            operands, VL_bits,
-            [](double x, double y) -> double { return (y / x); });
+            operands, VL_bits, [](double x, double y) -> double {
+              if (x == 0)
+                return std::nan(0);
+              else
+                return (y / x);
+            });
         break;
       }
       case Opcode::AArch64_FDIVR_ZPmZ_S: {  // fdivr zdn.s, pg/m, zdn.s, zm.s
         results[0] = sveLogicOpPredicated_3vecs<float>(
-            operands, VL_bits,
-            [](float x, float y) -> float { return (y / x); });
+            operands, VL_bits, [](float x, float y) -> float {
+              if (x == 0)
+                return std::nanf(0);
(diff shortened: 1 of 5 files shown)


src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -1653,14 +1653,22 @@ void Instruction::execute() {
...
       case Opcode::AArch64_FDIVR_ZPmZ_D: {  // fdivr zdn.d, pg/m, zdn.d, zm.d
         results[0] = sveLogicOpPredicated_3vecs<double>(
-            operands, VL_bits,
-            [](double x, double y) -> double { return (y / x); });
+            operands, VL_bits, [](double x, double y) -> double {
+              if (x == 0)
+                return std::nan(0);
+              else
+                return (y / x);
+            });
```
FinnWilkinson, line 1661: I'm not a huge fan of the passed function being this complex. Is it worth us creating helper functions for fdiv and fdivr? We can use `enable_if` to ensure the given types are floating point and then check the size of the typename to decide if we should use `nan` or `nanf`

FinnWilkinson (changes requested):

src/lib/arch/aarch64/Instruction_execute.cc
jj16791: In my opinion, adding a conditional and a return of a nan doesn't add much complexity so I don't think that point stands. Saying that, I'm by no means tied to this so happy to add a fdiv helper function. Given that fdivr is just fidv reversed, we could probably get away with one helper function and reverse the input operands.

jj16791: Or on the one helper function point, we could pass a bool in the template to denote whether its reversed?

FinnWilkinson: Complex was probably the wrong word --- it just looks a bit bloated perhaps?

I think implementing an fdiv helper would aid consistency with the rest of the execute switch case. Passing a bool to denote when operands are reversed is good - we do something similar for other helper functions

FinnWilkinson (approved):

JosephMoore25 (approved):

dANW34V3R (approved):

PR #385, opened 2024-02-07
Architecture Class cleanup

FinnWilkinson:
This PR moves common variables into the base Architecture class, ensures functions in `.cc` and `.hh` are in the same order, and generally makes the classes consistent with the codebase.

Closes Issue #357

### DIFF
--- a/src/include/simeng/arch/aarch64/Architecture.hh
+++ b/src/include/simeng/arch/aarch64/Architecture.hh
@@ -29,12 +29,6 @@ class Architecture : public arch::Architecture {
                     uint64_t instructionAddress,
                     MacroOp& output) const override;
 
-  /** Retrieve an ExecutionInfo object for the requested instruction. If a
-   * opcode-based override has been defined for the latency and/or
-   * port information, return that instead of the group-defined execution
-   * information. */
-  virtual ExecutionInfo getExecutionInfo(const Instruction& insn) const;
-
   /** Returns a zero-indexed register tag for a system register encoding.
    * Returns -1 in the case that the system register has no mapping. */
   int32_t getSystemRegisterTag(uint16_t reg) const override;
@@ -56,6 +50,12 @@ class Architecture : public arch::Architecture {
   /** Updates System registers of any system-based timers. */
   void updateSystemTimerRegisters(RegisterFileSet* regFile,
                                   const uint64_t iterations) const override;
+
+  /** Retrieve an ExecutionInfo object for the requested instruction. If a
+   * opcode-based override has been defined for the latency and/or
+   * port information, return that instead of the group-defined execution
+   * information. */
+  virtual ExecutionInfo getExecutionInfo(const Instruction& insn) const;
 
   /** Returns the current vector length set by the provided configuration. */
   uint64_t getVectorLength() const;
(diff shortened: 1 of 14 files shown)


ABenC377 (dismissed):

test/unit/pipeline/FetchUnitTest.cc
```diff
@@ -6,6 +6,7 @@
 #include "gtest/gtest.h"
 #include "simeng/Instruction.hh"
 #include "simeng/arch/Architecture.hh"
+#include "simeng/kernel/Linux.hh"
```
jj16791, line 9: Given that `#include "simeng/arch/Architecture.hh"` already includes `#include "simeng/kernel/Linux.hh"`, this include should be redundant.

jj16791 (changes requested):

test/unit/pipeline/FetchUnitTest.cc
FinnWilkinson: Removed

jj16791 (approved):

ABenC377 (dismissed):

JosephMoore25 (dismissed):

src/include/simeng/arch/aarch64/Architecture.hh
```diff
@@ -73,31 +74,12 @@ class Architecture : public arch::Architecture {
   /** A decoding cache, mapping an instruction word to a previously decoded
    * instruction. Instructions are added to the cache as they're decoded, to
    * reduce the overhead of future decoding. */
-  static std::unordered_map<uint32_t, Instruction> decodeCache;
+  mutable std::unordered_map<uint32_t, Instruction> decodeCache_;
```
jj16791, line 77: Why are these marked as `mutable`?

jj16791 (changes requested):

dANW34V3R (commented):
1 Query

src/lib/arch/aarch64/Architecture.cc
```diff
@@ -126,23 +122,17 @@ Architecture::Architecture(kernel::Linux& kernel, ryml::ConstNodeRef config)
...
 }
-Architecture::~Architecture() {
-  cs_close(&capstoneHandle);
-  decodeCache.clear();
-  metadataCache.clear();
-  groupExecutionInfo_.clear();
-  SVCRval_ = 0;
-}
+
+Architecture::~Architecture() { cs_close(&capstoneHandle_); }
```
dANW34V3R, line 135: Why do we not need to clear the caches and groupExecutionInfo on deconstruction now?

src/include/simeng/arch/aarch64/Architecture.hh
jj16791: Sounds good to me

jj16791 (dismissed):

src/lib/arch/aarch64/Architecture.cc
FinnWilkinson: As they are no longer static, they will be destructed between instantiations of the Architecture class

dANW34V3R (dismissed):

jj16791 (approved):

dANW34V3R (approved):

JosephMoore25 (approved):

ABenC377 (approved):

PR #387, opened 2024-02-08
Core class cleanup

### LINKED ISSUE #356: Clean Up Core Class
Due to more growth in some core models than others (i.e. Out of Order) the abstract core class needs tidying up to ensure all functions and member variables are in the appropriate place

### LINKED ISSUE #386: `SimEng/src/lib/Elf.cc` missing cstdio
At least on Fedora 39 with GCC 13, `<cstdio>` is no longer brought in by other system headers.
A explicit `#include <cstdio>` is required for compilation, otherwise we get:

```
In file included from /home/tom/SimEng/src/lib/Elf.cc:1:
/home/tom/SimEng/src/include/simeng/Elf.hh:30:3: error: ‘uint32_t’ does not name a type
   30 |   uint32_t p_type;
      |   ^~~~~~~~
/home/tom/SimEng/src/include/simeng/Elf.hh:7:1: note: ‘uint32_t’ is defined in header ‘<cstdint>’; did you forget to ‘#include <cstdint>’?
    6 | #include "simeng/span.hh"
  +++ |+#include <cstdint>
    7 | 
```

FinnWilkinson:
Cleans up the Core classes by moving common member variables into the base class and ensuring the code aligns with the rest of the project.

Closes #356 
Fixes #386

### DIFF
--- a/src/include/simeng/Core.hh
+++ b/src/include/simeng/Core.hh
@@ -27,8 +27,7 @@ class Core {
         registerFileSet_(regFileStructure),
         clockFrequency_(
             config::SimInfo::getConfig()["Core"]["Clock-Frequency-GHz"]
-                .as<float>() *
-            1e9) {}
+                .as<float>()) {}
 
   virtual ~Core() {}
 
@@ -45,11 +44,14 @@ class Core {
   /** Retrieve the number of instructions retired. */
   virtual uint64_t getInstructionsRetiredCount() const = 0;
 
-  /** Retrieve the simulated nanoseconds elapsed since the core started. */
-  virtual uint64_t getSystemTimer() const = 0;
-
   /** Retrieve a map of statistics to report. */
   virtual std::map<std::string, std::string> getStats() const = 0;
+
+  /** Retrieve the simulated nanoseconds elapsed since the core started. */
+  uint64_t getSystemTimer() const {
+    // TODO: This will need to be changed if we start supporting DVFS.
+    return (ticks_ / clockFrequency_);
+  }
 
  protected:
   /** Apply changes to the process state. */
@@ -111,8 +113,8 @@ class Core {
   /** Whether or not the core has halted. */
   bool hasHalted_ = false;
 
-  /** Clock frequency of core in Hz */
-  uint64_t clockFrequency_ = 0;
+  /** Clock frequency of core in GHz */
+  float clockFrequency_ = 0.0f;
 };
 
 }  // namespace simeng
--- a/src/include/simeng/Core.hh
+++ b/src/include/simeng/Core.hh
@@ -5,9 +5,9 @@
 #include <string>
 
 #include "simeng/ArchitecturalRegisterFileSet.hh"
-#include "simeng/MemoryInterface.hh"
 #include "simeng/arch/ProcessStateChange.hh"
 #include "simeng/config/SimInfo.hh"
+#include "simeng/memory/MemoryInterface.hh"
 
 namespace simeng {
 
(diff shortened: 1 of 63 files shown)


src/include/simeng/arch/ProcessStateChange.hh
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
jj16791, line 1: How come this file needs no includes?

FinnWilkinson: That is a great question.... perhaps when it is linked with the files that do use it, it has the necessary headers available?

It also seems to be ok in the Jenkin's pipeline, but I can add the headers it should need if preferable?

FinnWilkinson: Headers added along with some general code "de-tangling"

jj16791 (dismissed):

ABenC377 (changes requested):
Only comment relates to the units used for clock frequency.  So far as I can tell, there isn't a good reason for using GHz in some places and Hz in others.  In which case, it would seem better to stick to one unit throughout the code base.  However, @FinnWilkinson @jj16791 @dANW34V3R @JosephMoore25, please let me know if I'm missing something

jj16791:
> Only comment relates to the units used for clock frequency. So far as I can tell, there isn't a good reason for using GHz in some places and Hz in others. In which case, it would seem better to stick to one unit throughout the code base. However, @FinnWilkinson @jj16791 @dANW34V3R @JosephMoore25, please let me know if I'm missing something

You'd be correct in the use cases primarily converting the input GHz value into Hz. A significant reason behind the input being in GHz is that that's the natural unit of measure for modern HPC CPU cores. It's very unnatural to think about the frequency in terms of Hz and we'd need to add comments and notes to ensure that the user diverts from the natural assumptions. We kept it as GHz and got the code to convert it to Hz to make it "easier"/more convenient for the user.

FinnWilkinson:
> > Only comment relates to the units used for clock frequency. So far as I can tell, there isn't a good reason for using GHz in some places and Hz in others. In which case, it would seem better to stick to one unit throughout the code base. However, @FinnWilkinson @jj16791 @dANW34V3R @JosephMoore25, please let me know if I'm missing something
> 
> You'd be correct in the use cases primarily converting the input GHz value into Hz. A significant reason behind the input being in GHz is that that's the natural unit of measure for modern HPC CPU cores. It's very unnatural to think about the frequency in terms of Hz and we'd need to add comments and notes to ensure that the user diverts from the natural assumptions. We kept it as GHz and got the code to convert it to Hz to make it "easier"/more convenient for the user.

Is there a reason for using Hz in the code at all though? We could store `clockFrequency_` as a float and in terms of GHz given the usage inside each core is to just divide it by 1e9 again (i.e. put it back from Hz to GHz)

src/include/simeng/Core.hh
```diff
@@ -33,6 +50,69 @@ class Core {
...
+  /** The active exception handler. */
+  std::shared_ptr<arch::ExceptionHandler> exceptionHandler_;
+
+  /** The number of times this core has been ticked. */
+  uint64_t ticks_ = 0;
+
+  /** Whether or not the core has halted. */
+  bool hasHalted_ = false;
+
+  /** Clock frequency of core in Hz */
```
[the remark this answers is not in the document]
FinnWilkinson: Good point - I've updated this so that the clock frequency is stored as GHz inside the core

jj16791 (dismissed):

ABenC377 (dismissed):

JosephMoore25 (dismissed):

dANW34V3R (dismissed):

JosephMoore25 (approved):

jj16791 (approved):

dANW34V3R (dismissed):

ABenC377 (approved):

jj16791 (approved):

JosephMoore25 (approved):

ABenC377 (approved):

PR #414, opened 2024-05-24
Speculated assertions

dANW34V3R:
This PR updates logic used to decode compressed instructions. Previously, assertions were used to check for invalid usages of instructions but unfortunately these could be falsely triggered by speculated instructions which did not retire. Logic has been updated to cause an exception if one of these instructions reaches the commit stage. RISC-V instruction objects now carry a small string to provide extra information when the exception is printed.

Instructions with parameters that encode HINTS are now allowed to pass freely through the pipeline as they will cause no architectural state change. Quote from the spec: "HINTs are encoded as integer computational instructions with rd=x0. Hence, like the NOP instruction, HINTs do not change any architecturally visible state, except for advancing the pc and any applicable performance counters. Implementations are always allowed to ignore the encoded hints. ... [They] are usually used to communicate performance hints to the microarchitecture"

### DIFF
--- a/src/include/simeng/arch/riscv/Instruction.hh
+++ b/src/include/simeng/arch/riscv/Instruction.hh
@@ -185,9 +185,6 @@ class Instruction : public simeng::Instruction {
    * processing this instruction. */
   InstructionException getException() const;
 
-  /** Retrieve exception string */
-  std::string getExceptionString() const;
-
  private:
   /** Process the instruction's metadata to determine source/destination
    * registers. */
@@ -246,9 +243,6 @@ class Instruction : public simeng::Instruction {
 
   /** The current exception state of this instruction. */
   InstructionException exception_ = InstructionException::None;
-
-  /** Extra information about the exception */
-  std::string exceptionString_ = "";
 
   /** The number of source operands that have not yet had values supplied. Used
    * to determine execution readiness. */
--- a/src/include/simeng/arch/riscv/Instruction.hh
+++ b/src/include/simeng/arch/riscv/Instruction.hh
@@ -185,6 +185,9 @@ class Instruction : public simeng::Instruction {
    * processing this instruction. */
   InstructionException getException() const;
 
+  /** Retrieve exception string */
+  std::string getExceptionString() const;
+
  private:
   /** Process the instruction's metadata to determine source/destination
    * registers. */
@@ -243,6 +246,9 @@ class Instruction : public simeng::Instruction {
 
   /** The current exception state of this instruction. */
   InstructionException exception_ = InstructionException::None;
+
+  /** Extra information about the exception */
+  std::string exceptionString_ = "";
 
   /** The number of source operands that have not yet had values supplied. Used
    * to determine execution readiness. */
--- a/src/lib/arch/aarch64/ExceptionHandler.cc
+++ b/src/lib/arch/aarch64/ExceptionHandler.cc
@@ -942,6 +942,12 @@ void ExceptionHandler::printException(const Instruction& insn) const {
   std::cout << std::endl;
   std::cout << "[SimEng:ExceptionHandler]       opcode ID: " << metadata.opcode
             << std::endl;
+
+  std::string extraInformation = insn.getExceptionString();
+  if (!extraInformation.empty()) {
+    std::cout << "[SimEng:ExceptionHandler]     Extra information: "
+              << extraInformation << std::endl;
+  }
 }
 
 bool ExceptionHandler::fatal() {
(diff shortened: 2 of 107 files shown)


ABenC377 (dismissed):

FinnWilkinson (changes requested):
I really like this concept of being able to add extra information to an instruction when an exception occurs. Could you also add this to the aarch64 instruction class and ExceptionHandler?

src/lib/arch/riscv/InstructionMetadata.cc
```diff
@@ -549,9 +554,6 @@ void InstructionMetadata::convertCompressedInstruction(const cs_insn& insn) {
       // C.LI rd, imm, _ -> addi rd, zero, imm
 
       // rd = zero encodes hints
```
jj16791, line 556: I assume this possible decoding is no longer possible?

jj16791: Rather than write this comment multiple times, there are several cases in this file where asserts are removed but don't seem to be replaced with an `illegalAlias` call. Just double checking they're all intended

dANW34V3R: This is intentional. While these instructions encode hints which we do not support, the instructions themselves aren't illegal and should pass through the pipeline. Looking closely, all instructions encoding hints have Rd = X0 i.e. the destination register is the zero register. The result of any operation that these instructions perform is discarded so they can safely flow through the pipeline and have no architectural effect

src/include/simeng/arch/riscv/Instruction.hh
```diff
@@ -185,6 +185,9 @@ class Instruction : public simeng::Instruction {
    * processing this instruction. */
   InstructionException getException() const;
 
+  /** Retrieve exception string */
+  std::string getExceptionString() const;
+
```
FinnWilkinson: Re-looking through this, I think it can be simplified. Both the Instruction and Metadata classes contain a `exceptionString_` variable, and a `getExceptionString` function, but I don't think this is necessary. 

Given `getExceptionString` is only called in the exceptionHandler (where we get the instruction's metadata anyway) we could remove both the variable and function from the Instruction classes, and simply access `exceptionString_` via the metadata function directly.

dANW34V3R:
#rerun tests

jj16791 (commented):
Holding off on approval until https://github.com/UoB-HPC/SimEng/pull/414#discussion_r1658223370 has been resolved

src/include/simeng/arch/riscv/Instruction.hh
dANW34V3R: I agree, de-duplication would be cleaner. The proposed solution would mean we can only set the exception string from within the InstructionMetadata class as it is private e.g. we only get extra information from metadata. This would remove some potentially useful but currently unused functionality of setting the exception string from elsewhere e.g. instruction execute, pipeline units etc. This could be useful when debugging or provide more useful NYI information. There are a couple of avenues for this implementation (setter in `Instruction` to set the value in metadata or keeping the duplication) but these can be done when we want this functionality. Will de-duplicate for now unless you think differently.

FinnWilkinson: I think (and as you said) implementing extended functionality as we see fit and de-duplicating now is a sensible route. If it becomes apparent that this is very useful for debugging, or is required to be set elsewhere in the pipeline then it should be trivial to extend the functionality at that time. However, at the moment it is unknown if it will be used beyond the current case at all, so a tidier solution seems ideal

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -942,6 +942,12 @@ void ExceptionHandler::printException(const Instruction& insn) const {
   std::cout << std::endl;
   std::cout << "[SimEng:ExceptionHandler]       opcode ID: " << metadata.opcode
             << std::endl;
+
+  std::string extraInformation = insn.getMetadata().getExceptionString();
```
FinnWilkinson: no need to re-get metadata, this is already done above via `auto& metadata = ...`

src/lib/arch/riscv/ExceptionHandler.cc
```diff
@@ -923,6 +923,12 @@ void ExceptionHandler::printException(const Instruction& insn) const {
   std::cout << std::endl;
   std::cout << "[SimEng:ExceptionHandler]      opcode ID: " << metadata.opcode;
   std::cout << std::endl;
+
+  std::string extraInformation = insn.getMetadata().getExceptionString();
```
FinnWilkinson: Same as AArch64

FinnWilkinson (changes requested):

FinnWilkinson (approved):

ABenC377 (approved):

jj16791 (approved):

JosephMoore25 (approved):

PR #415, opened 2024-06-12
Full SME(1) instruction support and STREAMING Groups

FinnWilkinson:
This PR implements all available SME (version 1) instructions that are contained within LLVM 14.0.5. Specifically, this is [Version 2021-06 of the Armv9-A A64 ISA](https://developer.arm.com/documentation/ddi0602/2021-06/SME-Instructions?lang=en). 

No FP16 or BF16 instructions have been supported due to lacking C++17 types. All Quad-Word instruction variants have been emulated using 64-bit data-types.

In addition to this, new STREAMING_SVE and STREAMING_PREDICATE groups have been introduced (along with corresponding decode logic) to allow for a different pipeline / latency configuration for these instructions when SVE Streaming Mode (the context mode which SME instructions are executed in) is enabled. This can allow for a co-processor style implementation of SME to be implemented within SimEng; with additional latency / reduced throughput being configured to mimic an offload penalty, and different execution or LD/STR hardware being modelled for said co-processor compared to the main core.

- [x] Add STREAMING Group support
- [x] Add execution logic and regression tests for all missing SME instructions


FinnWilkinson:
#rerun tests

src/include/simeng/arch/aarch64/Architecture.hh
```diff
@@ -70,6 +70,12 @@ class Architecture : public arch::Architecture {
   /** Returns the current value of SVCRval_. */
   uint64_t getSVCRval() const;
 
+  /** Returns if SVE Streaming Mode is enabled. */
+  bool isStreamingModeEnabled() const;
+
+  /** Returns if the SME ZA Register is enabled. */
+  bool isZA_RegisterEnabled() const;
```
ABenC377: Is the underscore in ZA_Register needed/canonical?  Looks a bit weird jumping between camel and snake cases

src/lib/arch/aarch64/Architecture.cc
```diff
@@ -281,6 +295,12 @@ void Architecture::setSVCRval(const uint64_t newVal) const {
   SVCRval_ = newVal;
 }
 
+// 0th bit of SVCR register determines if streaming-mode is enabled.
+bool Architecture::isStreamingModeEnabled() const { return SVCRval_ & 1; }
+
+// 1st bit of SVCR register determines if ZA register is enabled.
+bool Architecture::isZA_RegisterEnabled() const { return SVCRval_ & 2; }
```
ABenC377: Same comment about underscore

src/include/simeng/arch/aarch64/InstructionGroups.hh
```diff
@@ -72,37 +72,53 @@ const uint16_t LOAD_SVE = 62;
...
-const uint16_t SME_SIMPLE_CVT = 79;
-const uint16_t SME_MUL = 80;
-const uint16_t SME_DIV_OR_SQRT = 81;
-const uint16_t LOAD_SME = 82;
-const uint16_t STORE_ADDRESS_SME = 83;
-const uint16_t STORE_DATA_SME = 84;
-const uint16_t STORE_SME = 85;
-const uint16_t ALL = 86;
-const uint16_t NONE = 87;
+const uint16_t STREAMING_SVE = 66;
```
ABenC377: When I looked at this file in isolation, I assumed that the ordering of these was arbitrary.  However, it seems to be important for how instruction.cc works.  So I'd be in favour of adding a comment here to explain that the ordering of these instructions is important and can't be changed without looking at instruction.cc.

src/lib/arch/aarch64/Instruction_address.cc
```diff
@@ -91,6 +91,19 @@ span<const memory::MemoryAccessTarget> Instruction::generateAddresses() {
         setMemoryAddresses({{sourceValues_[2].get<uint64_t>(), 8}});
         break;
       }
+      case Opcode::AArch64_LD1_MXIPXX_V_B:    // ld1b {zatv.b[ws, #imm]}, pg/z,
```
ABenC377: Should we be using "[[fallthrough]];" here as well?  Apply same to other non-attributed fallthroughs below

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -108,6 +108,144 @@ void Instruction::execute() {
...
+            sourceValues_[rowCount + 1].getAsVector<uint64_t>();
+        const uint64_t* zn =
+            sourceValues_[rowCount + 2].getAsVector<uint64_t>();
+
+        for (uint16_t row = 0; row < rowCount; row++) {
+          const uint64_t* zaRow = sourceValues_[row].getAsVector<uint64_t>();
+          uint64_t out[32] = {0};
+          std::memcpy(out, zaRow, rowCount * sizeof(uint64_t));
+          // Slice element is active IFF:
+          //  - Element in 1st source pred corresponding to horiz. slice is TRUE
```
ABenC377: Add an AND between these two statements

test/regression/aarch64/instructions/sme.cc
```diff
@@ -8,110 +8,269 @@ namespace {
...
+                  fillNeon<uint32_t>(inter32, (SVL / 8)));
+    if (i % 2 == 0) {
+      // Even cols, all elements
+      CHECK_MAT_COL(ARM64_REG_ZAS2, i, uint32_t,
+                    fillNeon<uint32_t>(index32, (SVL / 8)));
+      // Even cols, even elements
+      CHECK_MAT_COL(ARM64_REG_ZAS3, i, uint32_t,
+                    fillNeon<uint32_t>(inter32, (SVL / 8)));
+    } else {
+      // Even cols, all elements
```
ABenC377, line 194: Should this be Odd cols?

ABenC377 (changes requested):

CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 # Create variable to enable additional compiler warnings for SimEng targets only
-set(SIMENG_COMPILE_OPTIONS -Wall -pedantic -Werror) #-Wextra
+set(SIMENG_COMPILE_OPTIONS -Wall -pedantic) #-Wextra
```
dANW34V3R: What was causing the issue such that this needed to be removed. Should be fixed and Werror reinstated before we merge

src/include/simeng/arch/aarch64/Architecture.hh
dANW34V3R: +1

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -646,7 +641,8 @@ void Instruction::decode() {
       (2506 <= metadata_.opcode && metadata_.opcode <= 2508) ||
       (3682 <= metadata_.opcode && metadata_.opcode <= 3685) ||
       (2405 <= metadata_.opcode &&
-       metadata_.opcode <= 2408) ||  // all SME FMOPS & FMOPA variants
+       metadata_.opcode <= 2408) ||  // all SME MOPS & MOPA variants
+      (2405 <= metadata_.opcode && metadata_.opcode <= 2408) ||
       (4337 <= metadata_.opcode && metadata_.opcode <= 4340) ||
       (5391 <= metadata_.opcode && metadata_.opcode <= 5394) ||
```
dANW34V3R: I know there is a heck of a lot of them now but shouldn't these all be the opcode names as found in AArchGenInstrInfo.inc similar to how RISC-V does it

src/include/simeng/Register.hh
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include <cstdint>
```
dANW34V3R, line 2: Is this include needed? Either way I don't think iostream is needed

configs/a64fx_SME.yaml
```diff
@@ -62,7 +62,12 @@ Ports:
     Instruction-Group-Support:
     - INT_SIMPLE
     - INT_MUL
-    - STORE_DATA
+    - STORE_DATA_INT
+    - STORE_DATA_SCALAR
+    - STORE_DATA_VECTOR
+    - STORE_DATA_SVE
+    - STORE_DATA_STREAMING_SVE
+    - STORE_DATA_SME
```
dANW34V3R: I haven't properly looked through all the groups but can this block be changed simply to `STORE_DATA`. Same below

dANW34V3R: Can this port not do `STORE_FP`?

CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 # Create variable to enable additional compiler warnings for SimEng targets only
-set(SIMENG_COMPILE_OPTIONS -Wall -pedantic -Werror) #-Wextra
```
jj16791: Why was this removed?

configs/a64fx_SME.yaml
```diff
@@ -62,7 +62,12 @@ Ports:
     Instruction-Group-Support:
     - INT_SIMPLE
     - INT_MUL
-    - STORE_DATA
+    - STORE_DATA_INT
```
jj16791: Aren't these expansions redundant due to the inheritance mappings defined?

docs/sphinx/assets/instruction_groups_AArch64.png
jj16791, line 1: LOAD_FP/STORE_ADDRESS_FP/STORE_DATA_FP don't exist. Also, this may be GitHub not showing it correctly, but the background should just be the alpha channel.

src/lib/arch/aarch64/Instruction_address.cc
jj16791: Functionally I think it's fine but it provides a better structure and possibly improved readability so I'd say we should use fallthroughs

src/lib/arch/aarch64/Instruction.cc
```diff
@@ -192,6 +164,33 @@ const Architecture& Instruction::getArchitecture() const {
...
+  // Only instruction groups that depend on SVE Streaming Mode are SVE and
+  // PREDICATE
+  const uint16_t currentGroup = instructionGroup_;
+  const bool smEnabled = architecture_.isStreamingModeEnabled();
+  if (isInstruction(InsnType::isPredicate)) {
+    // Decide on predicate group based on whether SVE Streaming Mode is enabled.
+    instructionGroup_ = smEnabled ? InstructionGroups::STREAMING_PREDICATE
+                                  : InstructionGroups::PREDICATE;
+  } else if (isInstruction(InsnType::isSVEData)) {
+    assert(((instructionGroup_ >= InstructionGroups::SVE &&
```
jj16791: Was there a particular reason why there's an assert for this conditional body but not the one above? Is there a possible edge case or something?

src/lib/arch/aarch64/Architecture.cc
```diff
@@ -188,6 +188,20 @@ uint8_t Architecture::predecode(const uint8_t* ptr, uint16_t bytesAvailable,
     newInsn.setExecutionInfo(getExecutionInfo(newInsn));
     // Cache the instruction
     iter = decodeCache_.insert({insn, newInsn}).first;
+  } else {
```
jj16791: I have no data on this but doing this process for every AArch64 instruction may have a detrimental effect on performance. Will need to run through the new CI/CD when it's ready to determine this. Would argue that such a change shouldn't be merged until we know there's no significant performance regression.

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -695,6 +691,50 @@ void Instruction::decode() {
...
+    group += 6;
+  else if (isInstruction(InsnType::isLogical)) {
+    if (isInstruction(InsnType::isShift))
+      group += 4;
+    else
+      group += 5;
+  } else if (isInstruction(InsnType::isShift))
+    group += 2;
+  else
+    group += 3;  // Default is {Data type}_SIMPLE_ARTH
```
jj16791: Default is {Data type}_SIMPLE_ARTH_NOSHIFT

src/include/simeng/arch/aarch64/helpers/neon.hh
```diff
@@ -568,9 +568,14 @@ RegisterValue vecUMaxP(srcValContainer& sourceValues) {
   const T* n = sourceValues[0].getAsVector<T>();
   const T* m = sourceValues[1].getAsVector<T>();
 
+  // Concatenate the vectors
```
jj16791, line 571: Have you double-checked the ordering of the concatenation? Ran it on ookami and I think these may be the wrong way round but worth double checking in case I've made a mistake

src/lib/arch/aarch64/Instruction_execute.cc
jj16791: I don't think we need to short-hand horizontal.

test/regression/aarch64/instructions/sme.cc
jj16791: Several occurrences of the possible same issue throughout

jj16791 (changes requested):
Some comments and I agree with several of Alex's comments. I think it would be good to get the ARM SME/SVE loops as part of our functional verification checks to help test these new instructions. I assume it would have to be done somewhere private though (not sure if we already have that guarantee in the upcoming CI/CD pipelines)?

CMakeLists.txt
FinnWilkinson: This was a temp fix for the ACM summer school as some were getting weird errors, related to #428 . Will revert

dANW34V3R (commented):
Haven't finished the review but posting comments to prevent overlaps

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -370,6 +370,12 @@ class Instruction : public simeng::Instruction {
    * processing this instruction. */
   InstructionException getException() const;
 
+  /** Checks whether the current SVE Streaming Mode status is different to when
+   * this instruction was first decoded, and updates the instruction group
+   * accordingly if required.
+   * Returns TRUE if the group was updated, FALSE otherwise. */
+  bool checkStreamingGroup();
```
dANW34V3R: To me the word "check" implies that there won't be state change but there is. Something along the lines of `checkStreamingGroupAndUpdate` may be more descriptive

src/lib/arch/riscv/Instruction_decode.cc
```diff
@@ -310,6 +310,33 @@ void Instruction::decode() {
...
+  else if (isInstruction(InsnType::isShift) ||
+           isInstruction(InsnType::isConvert))
+    group += 5;
+  else if (isInstruction(InsnType::isLogical))
+    group += 4;
+  else if (isInstruction(InsnType::isCompare))
+    group += 3;
+  else
+    group += 2;  // Default return is {Data type}_SIMPLE_ARTH
+
```
dANW34V3R: Would a switch statement on `instructionIdentifier` not be viable? Would be slightly nicer to read

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -5625,6 +7267,158 @@ void Instruction::execute() {
...
+              const uint64_t shifted_active_zn = 1ull << (znIndex % 64);
+              const uint64_t shifted_active_zm = 1ull << (zmIndex % 64);
+              if ((pn[znIndex / 64] & shifted_active_zn) &&
+                  (pm[zmIndex / 64] & shifted_active_zm))
+                sum -= (static_cast<int32_t>(zn[znIndex]) *
+                        static_cast<int32_t>(zm[zmIndex]));
+            }
+            outRow[col] = sum;
+          }
+          results_[row] = {outRow, 256};
```
dANW34V3R, line 7448: I may be misremembering but I think we said we would stop using this implicit registerValue initialisation in favour of an explicit one. If that is the case this should be updated throughout

dANW34V3R: e.g. `results_[row] = RegisterValue(outRow, 256)`

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -4528,26 +5851,26 @@ void Instruction::execute() {
...
-            mdata[md_size] = sourceValues_[x].getAsVector<uint32_t>()[sliceNum];
-            md_size++;
-          } else if (md_size) {
+            memData.push_back(
+                sourceValues_[x].getAsVector<uint32_t>()[sliceNum]);
+          } else if (memData.size() > 0) {
             memoryData_[index] =
-                RegisterValue((char*)mdata.data(), md_size * 4);
-            md_size = 0;
+                RegisterValue((char*)memData.data(), memData.size() * 4);
```
dANW34V3R, line 5888: `sizeof(uint32_t)` might be better than `4` but not that important. Applied throughout

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -4453,65 +5737,104 @@ void Instruction::execute() {
         const uint32_t sliceNum =
             (ws + metadata_.operands[0].sme_index.disp) % partition_num;
 
-        const uint64_t* tileSlice =
-            sourceValues_[sliceNum].getAsVector<uint64_t>();
-        memoryData_ = sve_merge_store_data<uint64_t>(tileSlice, pg, VL_bits);
+        std::vector<uint64_t> memData;
```
dANW34V3R, line 5764: Have you profiled the use of std::vector throughout. It might not make too much difference but "the internet" is saying they are always initialised (which might not be desirable for performance). Do you NEED 0 initialised memory? Do you need dynamically sized arrays? Or could this be done with a std::array or normal array? CPP reference also says reallocations (when resizing) could be costly, `reserve()` could be useful it if is a problem

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -2620,46 +3547,72 @@ void Instruction::execute() {
...
-          if (pg[i / 16] & shifted_active) {
-            out[i] = data[i];
+          uint16_t* row =
+              const_cast<uint16_t*>(sourceValues_[i].getAsVector<uint16_t>());
+          uint64_t shifted_active = 1ull << ((i % 32) * 2);
+          if (pg[i / 32] & shifted_active) {
+            row[sliceNum] = data[i];
           } else {
-            out[i] = 0;
+            row[sliceNum] = 0;
```
dANW34V3R: This assignment through the pointer to the underlying memory of the registerValue technically breaks our rule (as per the comment) that a registerValue is immutable. This will probably need to be refactored when the reinterpret_cast issue is being resolved with the solution being similar to the implementations below e.g. populate a std::vector and creating in new registerValue

dANW34V3R (commented):
LOOTS of new instructions, well done for grinding through them. Bring on SAIL

CMakeLists.txt
FinnWilkinson: see https://github.com/UoB-HPC/SimEng/pull/415#discussion_r1818888420

src/include/simeng/arch/aarch64/helpers/neon.hh
FinnWilkinson: As per the [spec](https://developer.arm.com/documentation/ddi0602/2024-09/SIMD-FP-Instructions/UMAXP--Unsigned-maximum-pairwise-?lang=en):

> This instruction creates a vector by concatenating the vector elements of the **first source** SIMD&FP register **after** the vector elements of the **second source** SIMD&FP register...

i.e. N is concatonated onto the end of M (M:N)

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: This logic is changed in PR #429 and uses regex on the opcode name, and wasn't in the scope of this PR

src/include/simeng/Register.hh
FinnWilkinson: Yes, see issue #426

configs/a64fx_SME.yaml
FinnWilkinson: I think this was an ACM summer school change to make it clearer for new users. Will revert / make simpler

src/lib/arch/aarch64/Instruction_execute.cc
FinnWilkinson: Thanks for noticing, will update this to keep RegVals immutable

test/regression/aarch64/instructions/sme.cc
FinnWilkinson: No, throughout the test file some SME tests use the same predicate patterns. Two predicates are always used: `p0` and `p1`.

 - `p0` is always set to all true using `ptrue p0.d` (for example)
 - `p1` is always set to the pattern {ON, OFF, ON, OFF...} using `zip1 p1.s, p0.s, p1.s` (for example)
 
Hence, when using `p1` as the predicate the rows, columns, or individual elements per row/col updated will always be even ones. This should also be reflected in the test initialisation with `inter32[i] = (i % 2 == 0) ? i : 65;` where only the even vector elements are set to increasing values, and odd values to the test-default of 66.

configs/a64fx_SME.yaml
FinnWilkinson: They are - this was done to make the example SME yaml more easily interpreted for ACM students. Will revert

docs/sphinx/assets/instruction_groups_AArch64.png
FinnWilkinson: Good spot - thanks

src/lib/arch/aarch64/Instruction.cc
FinnWilkinson: An assert was added to this case as the instruction group is changes by + or - other instruction groups (values) from each other. Hence, if an incorrect group was assigned then an incorrect group could be set.

For the above condition (i.e. for a predicate), one of two groups is set based on the value of `smEnabled` so it doesn't actually matter what the current group of the instruction is; hence no assert

src/lib/arch/aarch64/Architecture.cc
FinnWilkinson: For non Predicate / SVE instructions the overhead is 3 if statements and a function call which _should_ be minor. But yes, a performance regression test for this would be good. 

Not sure on what an alternative solution could be though if the performance impact is significant...

configs/a64fx_SME.yaml
FinnWilkinson: Also STORE_FP doesn't exist

src/lib/arch/riscv/Instruction_decode.cc
FinnWilkinson: I don't this a switch would work here, or would decrease readability if it did. `instructionIdentifier` is just an `uint32_t` and so the associated value used in the switch has no meaning. With the current implementation the `if` clause contains what is being checked against (i.e. `isMultiply`)

src/include/simeng/Register.hh
FinnWilkinson: iostream removed though

src/lib/arch/aarch64/Instruction_execute.cc
FinnWilkinson: There are 158 instances of using `{...}` to initialise a RegisterValue. Possibly more if they contain a line break due to a long-names helper function being used. 

If this is something we want to move away from then I think it should be a seperate PR, and the `{...}` constructor should be prohibited for this class (if possible)

src/lib/arch/aarch64/Instruction_execute.cc
FinnWilkinson: Again, if this is something we want to enforce then it should be its own PR as this occurs extremely often in this file. An instruction's opcode defines the data type used so we know what the multiplicand should be

src/lib/arch/aarch64/Instruction_execute.cc
FinnWilkinson: Given the vector length is unknown before simulation, and this instruction is predicated, to me the sensible and reasonable choice is a dynamic data structure such as a vector.

test/regression/aarch64/instructions/sme.cc
jj16791: So what does `i` represent here?

src/lib/arch/aarch64/Architecture.cc
jj16791: I'll leave this comment unresolved so we remember to look into this

src/include/simeng/arch/aarch64/helpers/neon.hh
jj16791: I think with "what the spec says" vs "observed values", the latter should probably be taken as the truth. So it's worth someone else double-checking that the values I've observed do go against what the spec says

test/regression/aarch64/instructions/sme.cc
FinnWilkinson: `i` is the row or column element index (i.e. from something like `int i = 0; i < sliceElements; i++` with `sliceElements` sometimes fixed to a constant representing the number of elements for the max SVL of 2048-bits) depending on if the Horizontal or Vertical instruction variant is being used.

So in the above, if `i` is even (index 0, 2, 4 etc) then the row/column element is set to `i` to reflect the use of `index z1.s, #0, #1` in the test. If `i` is odd then the value remains unchanged; for this test a value of `65`.

src/include/simeng/arch/aarch64/helpers/neon.hh
FinnWilkinson: This is very odd and confusing... I've also checked on Ookami and Isambard-AI with the following asm programme:
```asm
        movi v0.16b, #0
	movi v1.16b, #1
	movi v2.16b, #2

	umaxp v0.16b, v1.16b, v2.16b

	mov w12, v0.s[0]
	mov w13, v0.s[3]
```
Which after executing yields the following:
 - `v0.b` = `{1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2}`
 - `v0.s` = `{16843009, 16843009, 33686018, 33686018}`

Which means the concatonation is `v1:v2`, NOT `v2:v1`. 
I double checked that gdb doesn't display vector registers "in reverse" (i.e. left-hand most element is in fact `v0[0]` and not `v0[15]`) using the final two instructions. Their results were:
 - `w12` = 16843009
 - `w13` = 33686018

So yes, on hardware the concatonation is seemingly `vn:vm`.

However, the spec and its pseudo code for `UMAXP` doesn't align with this... From [this](https://developer.arm.com/documentation/ddi0602/2024-09/SIMD-FP-Instructions/UMAXP--Unsigned-maximum-pairwise-?lang=en) page, the pseudo code is as follows:
```C
CheckFPAdvSIMDEnabled64();
constant bits(datasize) operand1 = V[n, datasize];
constant bits(datasize) operand2 = V[m, datasize];
bits(datasize) result;
constant bits(2*datasize) concat = operand2:operand1;
integer element1;
integer element2;
integer max;

for e = 0 to elements-1
    element1 = UInt(Elem[concat, 2*e, esize]);
    element2 = UInt(Elem[concat, (2*e)+1, esize]);
    max = Max(element1, element2);
    Elem[result, e, esize] = max<esize-1:0>;

V[d, datasize] = result;
```
 Where it is clear that the concatonation according to this is `vm:vn`....

In this instance, we should probably go with hardware. But it is quite annoying that the spec doesn't align with hardware on this, and that updating our code in-line with the spec still fixed the issue that was occuring!

src/lib/arch/aarch64/Instruction_execute.cc
dANW34V3R: Separate PR is sensible. Something to discuss as a team when we next meet

ABenC377 (dismissed):

test/regression/aarch64/instructions/sme.cc
ABenC377: I'm happy, but leaving for @jj16791 to resolve

src/lib/arch/aarch64/Instruction_execute.cc
FinnWilkinson: To add to this, there will also be 10s-100s of extra uses inside helper functions and possibly RV64 also

src/lib/arch/aarch64/Instruction_execute.cc
FinnWilkinson: Similar also occurs in instruction_address

src/lib/arch/aarch64/Instruction_execute.cc
dANW34V3R: This isn't a problem in RV as I was requested not to use full explicit initialisation. Although it was in a worse state https://github.com/UoB-HPC/SimEng/pull/234/commits/5a1652fbaaabcb2f39bf6dea30bee332f568825f

This isn't a big deal, would just be nice to have a consistent style

FinnWilkinson:
**Now outdated as STREAMING groups logic removed which was the only cause for slowdown.**
See below for this PR's performance compared to `dev` (times averaged over 5 runs):


|Benchmark|`dev` Time (ms)|`dev` StdDev|    |This PR Time (ms)|% diff to `dev`|This PR StdDev|
|---------|---------------|-----------|----|-----------------|---------------|--------------|
|CloverLeaf serial gcc8.3.0 armv8.4|13194.4|60.1| |13557.0|2.71%|132.52|
|CloverLeaf serial gcc9.3.0 armv8.4|13050.6|102.7| |13580.2|3.98%|84.94|
|CloverLeaf serial gcc10.3.0 armv8.4|13290.4|47.9| |13623.0|2.47%|44.06|
|CloverLeaf serial armclang20 armv8.4|11804.4|39.1| |12343.2|4.46%|77.05|
|CloverLeaf openmp gcc8.3.0 armv8.4|17509.4|161.5| |17889.8|2.15%|65.83|
|CloverLeaf openmp gcc9.3.0 armv8.4|17584.4|182.0| |17995.2|2.31%|152.27|
|CloverLeaf openmp gcc10.3.0 armv8.4|17119.8|61.3| |17651.4|3.06%|79.05|
|CloverLeaf openmp armclang20 armv8.4|15820.8|95.4| |16211.0|2.44%|83.58|
|miniBUDE openmp gcc8.3.0 armv8.4|24691.2|52.3| |24505.6|-0.75%|276.93|
|miniBUDE openmp gcc9.3.0 armv8.4|24500.0|175.6| |24412.8|-0.36%|155.77|
|miniBUDE openmp gcc10.3.0 armv8.4|24438.0|146.7| |24260.6|-0.73%|77.47|
|miniBUDE openmp armclang20 armv8.4|22725.2|150.0| |22343.4|-1.69%|67.39|
|STREAM serial gcc8.3.0 armv8.4|7378.0|40.3| |7769.8|5.17%|29.84|
|STREAM serial gcc9.3.0 armv8.4|7380.4|48.6| |7722.6|4.53%|68.62|
|STREAM serial gcc10.3.0 armv8.4|7530.6|71.7| |7632.6|1.35%|39.53|
|STREAM serial armclang20 armv8.4|8948.0|70.6| |8317.4|-7.30%|36.88|
|STREAM openmp gcc8.3.0 armv8.4|11552.6|139.5| |12020.4|3.97%|111.61|
|STREAM openmp gcc9.3.0 armv8.4|11737.0|133.1| |11855.8|1.01%|48.96|
|STREAM openmp gcc10.3.0 armv8.4|11357.4|36.4| |11768.0|3.55%|95.17|
|STREAM openmp armclang20 armv8.4|12701.0|227.5| |12309.0|-3.13%|87.32|
|TeaLeaf 2D serial gcc8.3.0 armv8.4|13964.4|41.8| |13605.8|-2.60%|42.13|
|TeaLeaf 2D serial gcc9.3.0 armv8.4|13976.2|40.8| |13553.6|-3.07%|88.90|
|TeaLeaf 2D serial gcc10.3.0 armv8.4|14231.0|92.2| |13961.2|-1.91%|109.20|
|TeaLeaf 2D serial armclang20 armv8.4|25691.8|86.2| |24628.8|-4.22%|199.33|
|TeaLeaf 2D openmp gcc8.3.0 armv8.4|20085.2|88.6| |20070.4|-0.07%|110.76|
|TeaLeaf 2D openmp gcc9.3.0 armv8.4|19980.2|79.3| |20492.8|2.53%|146.48|
|TeaLeaf 2D openmp gcc10.3.0 armv8.4|19684.8|88.1| |19522.4|-0.83%|100.20|
|TeaLeaf 2D openmp armclang20 armv8.4|58068.6|251.6| |61880.2|6.36%|284.36|
|TeaLeaf 3D serial gcc8.3.0 armv8.4|15853.0|128.6| |15818.2|-0.22%|57.76|
|TeaLeaf 3D serial gcc9.3.0 armv8.4|16483.8|58.3| |16334.6|-0.91%|87.93|
|TeaLeaf 3D serial gcc10.3.0 armv8.4|16839.8|86.0| |16521.4|-1.91%|28.94|
|TeaLeaf 3D serial armclang20 armv8.4|23052.2|157.0| |22959.8|-0.40%|134.67|
|TeaLeaf 3D openmp gcc8.3.0 armv8.4|26103.0|145.5| |26294.8|0.73%|190.12|
|TeaLeaf 3D openmp gcc9.3.0 armv8.4|26203.6|103.0| |27278.8|4.02%|239.28|
|TeaLeaf 3D openmp gcc10.3.0 armv8.4|26068.2|278.0| |26129.6|0.24%|112.81|
|TeaLeaf 3D openmp armclang20 armv8.4|45312.4|179.0| |48379.4|6.55%|136.36|
|CloverLeaf serial gcc8.3.0 armv8.4+sve|12763.0|89.1| |13372.0|4.66%|59.14|
|CloverLeaf serial gcc9.3.0 armv8.4+sve|12675.4|52.4| |13300.4|4.81%|134.66|
|CloverLeaf serial gcc10.3.0 armv8.4+sve|12665.4|88.7| |13086.4|3.27%|63.11|
|CloverLeaf serial armclang20 armv8.4+sve|12512.8|79.5| |12963.4|3.54%|71.92|
|CloverLeaf openmp gcc8.3.0 armv8.4+sve|16973.8|119.5| |17630.2|3.79%|197.66|
|CloverLeaf openmp gcc9.3.0 armv8.4+sve|17076.6|132.9| |17460.8|2.22%|53.09|
|CloverLeaf openmp gcc10.3.0 armv8.4+sve|16814.4|96.4| |17264.4|2.64%|76.24|
|CloverLeaf openmp armclang20 armv8.4+sve|16436.8|82.2| |16844.2|2.45%|98.85|
|miniBUDE openmp gcc8.3.0 armv8.4+sve|9745.6|125.8| |10291.4|5.45%|90.47|
|miniBUDE openmp gcc9.3.0 armv8.4+sve|9172.0|41.3| |10081.6|9.45%|64.37|
|miniBUDE openmp gcc10.3.0 armv8.4+sve|9180.0|36.6| |10054.0|9.09%|61.30|
|miniBUDE openmp armclang20 armv8.4+sve|9746.6|63.0| |10098.8|3.55%|85.55|
|STREAM serial gcc8.3.0 armv8.4+sve|3915.0|18.9| |4139.4|5.57%|15.92|
|STREAM serial gcc9.3.0 armv8.4+sve|3919.4|16.7| |4139.2|5.46%|18.14|
|STREAM serial gcc10.3.0 armv8.4+sve|3862.0|29.9| |4086.2|5.64%|23.04|
|STREAM serial armclang20 armv8.4+sve|2550.2|3.7| |2593.4|1.68%|17.33|
|STREAM openmp gcc8.3.0 armv8.4+sve|7977.4|32.4| |8196.2|2.71%|38.70|
|STREAM openmp gcc9.3.0 armv8.4+sve|7987.4|87.9| |8265.6|3.42%|12.76|
|STREAM openmp gcc10.3.0 armv8.4+sve|7999.2|69.2| |8051.0|0.65%|34.07|
|STREAM openmp armclang20 armv8.4+sve|6836.0|10.0| |6990.8|2.24%|35.39|
|TeaLeaf 2D serial gcc8.3.0 armv8.4+sve|14022.8|99.5| |13579.0|-3.22%|59.23|
|TeaLeaf 2D serial gcc9.3.0 armv8.4+sve|13996.4|63.8| |13610.4|-2.80%|64.07|
|TeaLeaf 2D serial gcc10.3.0 armv8.4+sve|14362.6|59.8| |13831.0|-3.77%|65.83|
|TeaLeaf 2D serial armclang20 armv8.4+sve|9835.2|75.5| |9782.0|-0.54%|113.49|
|TeaLeaf 2D openmp gcc8.3.0 armv8.4+sve|19885.8|62.1| |20026.2|0.70%|69.48|
|TeaLeaf 2D openmp gcc9.3.0 armv8.4+sve|20028.2|143.4| |20322.0|1.46%|111.62|
|TeaLeaf 2D openmp gcc10.3.0 armv8.4+sve|19695.6|83.6| |19575.4|-0.61%|38.66|
|TeaLeaf 2D openmp armclang20 armv8.4+sve|57176.4|405.7| |59327.4|3.69%|324.77|
|TeaLeaf 3D serial gcc8.3.0 armv8.4+sve|13828.8|50.1| |14023.6|1.40%|51.09|
|TeaLeaf 3D serial gcc9.3.0 armv8.4+sve|13901.6|36.4| |14065.6|1.17%|35.20|
|TeaLeaf 3D serial gcc10.3.0 armv8.4+sve|14043.8|58.0| |14203.2|1.13%|103.32|
|TeaLeaf 3D serial armclang20 armv8.4+sve|22478.8|138.4| |22850.6|1.64%|51.21|
|TeaLeaf 3D openmp gcc8.3.0 armv8.4+sve|23927.6|73.3| |24201.0|1.14%|94.22|
|TeaLeaf 3D openmp gcc9.3.0 armv8.4+sve|23638.8|119.3| |24663.4|4.24%|138.94|
|TeaLeaf 3D openmp gcc10.3.0 armv8.4+sve|23550.4|130.2| |24060.4|2.14%|31.31|
|TeaLeaf 3D openmp armclang20 armv8.4+sve|48104.8|253.4| |50293.2|4.45%|319.78|

src/lib/arch/aarch64/Instruction_execute.cc
dANW34V3R: I mainly say this as the `sizeof` syntax is used in other places throughout this file. It prevents the use of magic numbers so is preferable but would also be consistent. I think going forward this syntax would be beneficial, if we want to change it throughout then its own PR would be sensible

dANW34V3R: No action needed, resolve the conversation when you see it

src/lib/arch/aarch64/Instruction_execute.cc
dANW34V3R: Resolving as no immediate action needed

src/lib/arch/riscv/Instruction_decode.cc
dANW34V3R: I hadn't considered the case where an instruction has multiple types so I think this is probably the simpler solution (although not pretty). No action required

dANW34V3R (dismissed):

ABenC377 (dismissed):

src/include/simeng/arch/aarch64/helpers/neon.hh
jj16791: What was the issue that was occuring?

test/regression/aarch64/instructions/sme.cc
jj16791: Okay, it makes sense with all this ^^^. I would say add some comments to make it clear but I don't think it can be done succinctly. Motion to just resolve this comment (leaving so @FinnWilkinson can see this)

jj16791 (commented):
Looks good, just need to finish off resolve existing conversations for the sake of clarity

src/include/simeng/arch/aarch64/helpers/neon.hh
FinnWilkinson: The weird error with printf where when the binary was run through SimEng, it would print `"my string %d %d %d"` rather than `"my string 1 2 3"`

This instruction is used to find the end of the string, but with the previous implementation it could not and so it printed incorrectly. I think in the instance it is used, HW way or Spec way works due to luck...

jj16791: Well I motion to match HW and assume that the Spec has human-error

src/include/simeng/arch/aarch64/helpers/neon.hh
```diff
@@ -585,9 +590,14 @@ RegisterValue vecUMinP(srcValContainer& sourceValues) {
...
+  T temp[2 * I];
+  memcpy(temp, m, sizeof(T) * I);
```
jj16791: Shouldn't `m` and `n` be switched here?

jj16791 (changes requested):

src/include/simeng/arch/aarch64/helpers/neon.hh
FinnWilkinson: I agree - code is inline with HW

src/include/simeng/arch/aarch64/helpers/neon.hh
FinnWilkinson: Good spot, only updated maxP to be in-line with HW...

ABenC377 (dismissed):

dANW34V3R (dismissed):

ABenC377 (approved):

jj16791 (commented):
I assume that it has been checked, but a reminder of the SME loops comment in https://github.com/UoB-HPC/SimEng/pull/415#pullrequestreview-2397083736

FinnWilkinson:
> I assume that it has been checked, but a reminder of the SME loops comment in [#415 (review)](https://github.com/UoB-HPC/SimEng/pull/415#pullrequestreview-2397083736)

This PR doesn't add support for many of those loops (most LD/STR instructions still missing as need newer ISA version than this PR targets). Without access to physical hardware (i.e. an Apple M4) where we can generate small tests and verify the output against our current regression tests for these instructions, there currently aren't any binaries that we could run to ensure validity of most instructions added in this PR.

For PR #439, the loops would be able to be used (as that PR targets them) but still need to be done privately.

jj16791 (approved):

dANW34V3R (approved):

PR #417, opened 2024-06-19
Ensure Emulation core executes each micro-op to completion in 1 cycle

FinnWilkinson:
The previous emulation core implementation did not ensure that load instructions completed in a single cycle. Given we restrict emulation cores to only use Flat emmory interfaces (i.e. no cycle delay to update or access memory) there is no need for each load instruction to take 2 cycles.

This PR also simplifies some of the logic in the `tick()` function of the emulation core.

### DIFF
--- a/.jenkins/build_test_run.sh
+++ b/.jenkins/build_test_run.sh
@@ -63,7 +63,7 @@ run () {
     cat run
     echo ""
     compare_outputs "$(grep "retired:" run | rev | cut -d ' ' -f1 | rev)" "6708" "retired instructions"
-    compare_outputs "$(grep "cycles:" run | rev | cut -d ' ' -f1 | rev)" "7955" "simulated cycles"
+    compare_outputs "$(grep "cycles:" run | rev | cut -d ' ' -f1 | rev)" "6736" "simulated cycles"
     echo ""
 
     ./bin/simeng "$SIMENG_TOP"/configs/tx2.yaml > run
--- a/.jenkins/build_test_run.sh
+++ b/.jenkins/build_test_run.sh
@@ -63,7 +63,7 @@ run () {
     cat run
     echo ""
     compare_outputs "$(grep "retired:" run | rev | cut -d ' ' -f1 | rev)" "6721" "retired instructions"
-    compare_outputs "$(grep "cycles:" run | rev | cut -d ' ' -f1 | rev)" "6736" "simulated cycles"
+    compare_outputs "$(grep "cycles:" run | rev | cut -d ' ' -f1 | rev)" "6721" "simulated cycles"
     echo ""
 
     ./bin/simeng "$SIMENG_TOP"/configs/tx2.yaml > run
--- a/.jenkins/build_test_run.sh
+++ b/.jenkins/build_test_run.sh
@@ -62,7 +62,7 @@ run () {
     echo "Simulation without configuration file argument:"
     cat run
     echo ""
-    compare_outputs "$(grep "retired:" run | rev | cut -d ' ' -f1 | rev)" "6708" "retired instructions"
+    compare_outputs "$(grep "retired:" run | rev | cut -d ' ' -f1 | rev)" "6721" "retired instructions"
     compare_outputs "$(grep "cycles:" run | rev | cut -d ' ' -f1 | rev)" "6736" "simulated cycles"
     echo ""
 
--- a/src/lib/models/emulation/Core.cc
+++ b/src/lib/models/emulation/Core.cc
@@ -83,7 +83,7 @@ void Core::tick() {
       }
     }
 
-    // Execute
+    // Execute & Write-back
     if (uop->isLoad()) {
       auto addresses = uop->generateAddresses();
       previousAddresses_.clear();
@@ -139,6 +139,7 @@ void Core::tick() {
     execute(uop);
     macroOp_.erase(macroOp_.begin());
   }
+  // Commit
   instructionsExecuted_++;
   // Fetch memory for next cycle
   instructionMemory_.requestRead({pc_, FETCH_SIZE});
--- a/src/lib/models/emulation/Core.cc
+++ b/src/lib/models/emulation/Core.cc
@@ -172,18 +172,11 @@ void Core::execute(std::shared_ptr<Instruction>& uop) {
   }
 
   // Writeback
-  auto results = uop->getResults();
-  auto destinations = uop->getDestinationRegisters();
-  if (uop->isStoreData()) {
-    for (size_t i = 0; i < results.size(); i++) {
-      auto reg = destinations[i];
-      registerFileSet_.set(reg, results[i]);
-    }
-  } else {
-    for (size_t i = 0; i < results.size(); i++) {
-      auto reg = destinations[i];
-      registerFileSet_.set(reg, results[i]);
-    }
+  const auto& results = uop->getResults();
+  const auto& destinations = uop->getDestinationRegisters();
+  for (size_t i = 0; i < results.size(); i++) {
+    auto reg = destinations[i];
+    registerFileSet_.set(reg, results[i]);
   }
 }
 
--- a/src/lib/models/emulation/Core.cc
+++ b/src/lib/models/emulation/Core.cc
@@ -97,7 +97,7 @@ void Core::tick() {
       return;
     }
     if (addresses.size() > 0) {
-      // Memory reads required; request them.
+      // Memory reads required; request them
       for (auto const& target : addresses) {
         dataMemory_.requestRead(target);
         // Save addresses for use by instructions that perform a LD and STR
@@ -105,7 +105,7 @@ void Core::tick() {
         previousAddresses_.push_back(target);
       }
       // Emulation core can only be used with a Flat memory interface, so data
-      // is ready immediately.
+      // is ready immediately
       const auto& completedReads = dataMemory_.getCompletedReads();
       assert(completedReads.size() == addresses.size() &&
              "Number of completed reads does not match the number of requested "
--- a/src/lib/models/emulation/Core.cc
+++ b/src/lib/models/emulation/Core.cc
@@ -106,6 +106,9 @@ void Core::tick() {
       }
       // Emulation core can only be used with a Flat memory interface, so data
       // is ready immediately
+      assert((config::SimInfo::getConfig()["L1-Data-Memory"]["Interface-Type"]
+                  .as<std::string>() == "Flat") &&
+             "Emulation core is only compatable with a Flat Memory Interface.");
       const auto& completedReads = dataMemory_.getCompletedReads();
       assert(completedReads.size() == addresses.size() &&
              "Number of completed reads does not match the number of requested "
@@ -226,6 +229,11 @@ void Core::processExceptionHandler() {
 
   // Clear the handler
   exceptionHandler_ = nullptr;
+
+  // For an exception to reach this part of the code, it must be something akin
+  // to a system call, which itself should be counted as an instruction
+  // finishing execution.
+  instructionsExecuted_++;
 
   // Fetch memory for next cycle
   instructionMemory_.requestRead({pc_, FETCH_SIZE});
--- a/src/lib/models/emulation/Core.cc
+++ b/src/lib/models/emulation/Core.cc
@@ -130,7 +130,6 @@ void Core::tick() {
     execute(uop);
     macroOp_.erase(macroOp_.begin());
   }
-  // Commit
   instructionsExecuted_++;
   // Fetch memory for next cycle
   instructionMemory_.requestRead({pc_, FETCH_SIZE});
@@ -189,12 +188,8 @@ void Core::processExceptionHandler() {
   assert(exceptionHandler_ != nullptr &&
          "Attempted to process an exception handler that wasn't present");
 
-  while (true) {
-    bool success = exceptionHandler_->tick();
-    if (success) {
-      // No more ticks needed to complete exception
-      break;
-    }
+  // Tick until true is returned, signifying completion
+  while (exceptionHandler_->tick() == false) {
   }
 
   const auto& result = exceptionHandler_->getResult();
--- a/src/lib/models/emulation/Core.cc
+++ b/src/lib/models/emulation/Core.cc
@@ -27,6 +27,7 @@ Core::Core(memory::MemoryInterface& instructionMemory,
 
 void Core::tick() {
   ticks_++;
+  isa_.updateSystemTimerRegisters(&registerFileSet_, ticks_);
 
   if (hasHalted_) return;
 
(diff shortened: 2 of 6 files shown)


src/lib/models/emulation/Core.cc
```diff
@@ -118,19 +97,23 @@ void Core::tick() {
       return;
     }
     if (addresses.size() > 0) {
-      // Memory reads are required; request them, set `pendingReads_`
-      // accordingly, and end the cycle early
+      // Memory reads required; request them.
```
ABenC377: We don't use periods at the end of comments elsewhere, so I'd delete here

src/lib/models/emulation/Core.cc
```diff
@@ -118,19 +97,23 @@ void Core::tick() {
...
         previousAddresses_.push_back(target);
       }
-      pendingReads_ = addresses.size();
-      return;
-    } else {
-      // Early execution due to lacking addresses
-      execute(uop);
-      return;
+      // Emulation core can only be used with a Flat memory interface, so data
+      // is ready immediately.
```
ABenC377: as above re period

ABenC377 (changes requested):
Looks good.  Only minor comment

.jenkins/build_test_run.sh
```diff
@@ -63,7 +63,7 @@ run () {
     cat run
     echo ""
     compare_outputs "$(grep "retired:" run | rev | cut -d ' ' -f1 | rev)" "6708" "retired instructions"
-    compare_outputs "$(grep "cycles:" run | rev | cut -d ' ' -f1 | rev)" "7955" "simulated cycles"
+    compare_outputs "$(grep "cycles:" run | rev | cut -d ' ' -f1 | rev)" "6736" "simulated cycles"
```
dANW34V3R: Shouldn't these be the same after this change? What is the cause of the current discrepancy? Exceptions/syscalls maybe?

src/lib/models/emulation/Core.cc
```diff
@@ -118,19 +97,23 @@ void Core::tick() {
...
         previousAddresses_.push_back(target);
       }
-      pendingReads_ = addresses.size();
-      return;
-    } else {
-      // Early execution due to lacking addresses
-      execute(uop);
-      return;
+      // Emulation core can only be used with a Flat memory interface, so data
+      // is ready immediately
```
dANW34V3R: Is there a way that this can be enforced through an assert or different typing? I think currently the only line of defence is the checks in modelConfig:1040

dANW34V3R (commented):
Minor comments. Looking good

.jenkins/build_test_run.sh
FinnWilkinson: Load instructions now take 1 cycle rather than 2, so there will be a decrease in cycles for the emulation core

dANW34V3R: I am referring to the difference between retired instructions and cycles here. If every instruction takes 1 cycle then these should be equal

FinnWilkinson: Having looked into this, the disparity comes from an exception that occurs after an instruction is executed (i.e. a system call). I can move where the instruction executed counter is incremented to match cycles to retired instructions

FinnWilkinson: The exception handler is also multi-cycle on all core models, so even accounting for syscalls there is a natural disparity between instructions retired and cycles

dANW34V3R: That makes sense. I don't really mind either way as long as the behaviour is clear and obvious to the user. I guess the question is: Conceptually, should an exception that is raised by instruction "n" be handled, to completion, on the same cycle as "n" is executed in the emulation core? If the answer is yes, then I think the way that we handle an exception should be changed to reflect this rather than early returning on each tick and repeatedly ticking the core until the exception completes. Open to other thoughts. @jj16791 

As the initial change came about because of the GDB support, does this (the fact that exceptions occur on subsiquent cycle) have any effect on that as well?

FinnWilkinson: Specifically for the defult program, this is the `readlinkat` syscall which performs `readStringThen` which is multi-cycle as it reads one character from memory per cycle

FinnWilkinson: For the main point, it will be hard to make exceptions single cycle for just the emulation core without having a completely seperate exception handler per model, as the exception handlers are model agnostic.

As for GDB, I don't think stepping is reliant on the tick function of the core, so this shouldn't be an issue

dANW34V3R: One, not very well thought out, solution would be to have a while loop in Core::handleException (which is model specific) that loops until the `exceptionHandler_` becomes a `nullptr`. This would have the effect of handling the exception to completion on one call to `handleException`.

Another option is to put a similar while loop at the end of tick. Currently we are just repeatedly calling `Core::tick` until the exception handler completes, and on the next cycle carrying on with the next instruction. Haven't thought too hard so there could be something I am missing.

Good news for GDB.

src/lib/models/emulation/Core.cc
```diff
@@ -227,6 +230,11 @@ void Core::processExceptionHandler() {
   // Clear the handler
   exceptionHandler_ = nullptr;
 
+  // For an exception to reach this part of the code, it must be something akin
+  // to a system call, which itself should be counted as an instruction
+  // finishing execution.
+  instructionsExecuted_++;
```
dANW34V3R: Not convinced by this. Just marking so I don't forget. Waiting on the conclusion to the above comments

.jenkins/build_test_run.sh
FinnWilkinson: @jj16791 would be good to get your opinion on the above solutions to emulation core single cycle exceptions. I think the while loop idea could work well

src/lib/models/emulation/Core.cc
FinnWilkinson: This change will be uneffected by any change made to make exceptions single cycle. 

This part of the function can only be reached if the exception is triggered by something like a system call. Any exceptions due to a fault will return before this point. Given when an exception is reached in `Core::execute()` or `Core::tick()` we return before the usual `instructionsExecuted_++;` is not reached, this additional increment is required.

.jenkins/build_test_run.sh
jj16791: As discussed offline, the ticking of the handler in a while loop is a good way forward and reflects prior solutions made for the original GDB implementation

src/lib/models/emulation/Core.cc
```diff
@@ -40,29 +41,7 @@ void Core::tick() {
...
-    }
-
-    // More data pending, end cycle early
-    return;
-  }
-
   // Fetch
-
   // Determine if new uops are needed to be fetched
   if (!microOps_.size()) {
```
jj16791: Whilst this provides security, kinda, it does imply that we could tick an em core with microOps from the previous cycle's macroOp still in the buffer. The em core should see an entire instruction's lifecycle with an instruction being a macroOp not a microOp

src/lib/models/emulation/Core.cc
```diff
@@ -40,29 +41,7 @@ void Core::tick() {
...
     // Find fetched memory that matches the current PC
```
jj16791: I can't comment on the region of code which requires this change, but below we are only dealing with one microOp at a time.  Line `auto& uop = microOps_.front();` and most lines below it should be wrapped in a while loop with the condition of something equivalent to `!microOps_.empty()`

src/lib/models/emulation/Core.cc
```diff
@@ -118,19 +97,26 @@ void Core::tick() {
...
       }
-      pendingReads_ = addresses.size();
-      return;
-    } else {
-      // Early execution due to lacking addresses
-      execute(uop);
-      return;
+      // Emulation core can only be used with a Flat memory interface, so data
+      // is ready immediately
+      assert((config::SimInfo::getConfig()["L1-Data-Memory"]["Interface-Type"]
```
jj16791: We'd want to place this assert in the constructor rather than on each load. Also, as we did with the original GDB implementation, we could simply remove this limitation with code akin to 
```
        pendingReads_ = addresses.size();
        // Tick data memory until all data is recieved
        while (pendingReads) {
```
and in the while loop, continually check the memory interface for complete reads deduct from the number of pendingReads on each response. Could move pendingReads to a local variable rather than re-instate as a member variable.

A motivation for this would be a simulation setup where we want to do some sort of memory DSE and simplify the compute side of the core. The em core could act as a traffic generator in essence.

src/lib/models/emulation/Core.cc
```diff
@@ -139,23 +125,18 @@ void Core::tick() {
...
-    // Store addresses for use by next store data operation
+    // Store addresses for use by next store data operation in `execute()`
     for (auto const& target : addresses) {
       previousAddresses_.push_back(target);
     }
-    if (uop->isStoreData()) {
-      execute(uop);
-    } else {
-      // Fetch memory for next cycle
+    if (!uop->isStoreData()) {
```
jj16791: We could unify the 
```
instructionMemory_.requestRead({pc_, FETCH_SIZE});
microOps_.pop();
```
code snippet into one area rather than repeat this code multiple times. Will depend on any changes made from my recommendation so leaving this comment here to be revisited once we've resolved my other comments.

src/lib/models/emulation/Core.cc
```diff
@@ -249,6 +230,11 @@ void Core::processExceptionHandler() {
   // Clear the handler
   exceptionHandler_ = nullptr;
 
+  // For an exception to reach this part of the code, it must be something akin
```
jj16791: Think we could remove all code from here down if we just added a `if (hasHalted_)` condition in front of the return in the body of `if (uop->exceptionEncountered()) {` in `execute`

jj16791 (changes requested):
A few comments to do with more future-proofing and changed the emulation core to work on a per-macroOp basis rather than a per-microOp basis

src/lib/models/emulation/Core.cc
FinnWilkinson: I see two routes - either ensure all micro ops are dealt with during a tick, or disabling micro-opping for the emulation core. In my opinion, the latter is cleaner and easier, and I don't see a huge loss in functionality

FinnWilkinson: After offline discussion, each macro-op will be completed in one cycle, with micro-opping allowed

src/lib/models/emulation/Core.cc
FinnWilkinson: Comment likely irrelevant after most recent changes - instructionExecuted is now done in `tick()` and nowhere else

src/lib/models/emulation/Core.cc
FinnWilkinson: Have done the first part of this comment. As disccussed offline, the second part (i.e. enabling latent memory) is not being done as part of this PR, and instead will be combined with future in-order core changes to allow both of the mentioned core models to support latent memory

src/lib/models/emulation/Core.cc
```diff
@@ -26,136 +37,112 @@ Core::Core(memory::MemoryInterface& instructionMemory,
...
-      microOps_.pop();
+      if (!uop->isStoreData()) {
+        // No further action needed, move onto next micro-op
+        macroOp_.erase(macroOp_.begin());
+        continue;
+      }
     }
-
-    return;
+    execute(uop);
```
jj16791, line 130: Can't comment on code line but in `execute()`, we should be able to remove the if statement below `// Writeback` as the bodies of both conditions are the same.

src/lib/models/emulation/Core.cc
```diff
@@ -26,136 +37,112 @@ Core::Core(memory::MemoryInterface& instructionMemory,
...
-      if (fetched[fetchIndex].target.address == pc_) {
-        break;
-      }
-    }
-    if (fetchIndex == fetched.size()) {
-      // Need to wait for fetched instructions
-      return;
-    }
+  ticks_++;
+  isa_.updateSystemTimerRegisters(&registerFileSet_, ticks_);
```
jj16791, line 48: For consistency between core types, I think we'd either want all of them to call this at the start or end of `tick()`. As far as I can tell, there's no functional harm with either option. Maybe having it at the start of `tick()` is better as we're denoting the tick we're on rather than the tick we just performed.

jj16791 (changes requested):

src/lib/models/emulation/Core.cc
FinnWilkinson: Have moved `ticks_++` and system register update below the hasHalted check for all core types

src/lib/models/emulation/Core.cc
FinnWilkinson: Removed

jj16791 (dismissed):

ABenC377 (dismissed):

JosephMoore25 (dismissed):

src/lib/models/emulation/Core.cc
```diff
@@ -204,16 +188,13 @@ void Core::handleException(const std::shared_ptr<Instruction>& instruction) {
...
-  if (!success) {
-    // Handler needs further ticks to complete
-    return;
+  while (true) {
+    bool success = exceptionHandler_->tick();
+    if (success) {
+      // No more ticks needed to complete exception
+      break;
+    }
   }
```
dANW34V3R: This while loop is pretty horrible. The condition statement (i.e. `true`) doesn't convey the function of the loop. Other options include 

`while (!exceptionHandler_.tick()) {}`
or 
```
bool success = false;
while (!success) {
    success = exceptionHandler_.tick();
}
```

src/lib/models/emulation/Core.cc
```diff
@@ -26,117 +37,103 @@ Core::Core(memory::MemoryInterface& instructionMemory,
...
-      // No further action needed, fetch memory for next cycle and return early
-      instructionMemory_.requestRead({pc_, FETCH_SIZE});
-      microOps_.pop();
-      return;
     }
+    execute(uop);
+    macroOp_.erase(macroOp_.begin());
   }
-  execute(uop);
+  // Commit
```
dANW34V3R: Labelling this line as `Commit` seems a bit odd as nothing is functionally happening

dANW34V3R (commented):
Stylistic comments

src/lib/models/emulation/Core.cc
dANW34V3R: This line has been removed

src/lib/models/emulation/Core.cc
FinnWilkinson: Updated

src/lib/models/emulation/Core.cc
FinnWilkinson: Removed

dANW34V3R (approved):

jj16791 (approved):

ABenC377 (approved):

FinnWilkinson:
#rerun tests

PR #419, opened 2024-07-08
Group Tests

dANW34V3R:
Add infrastructure for testing the groups assigned to each micro-op of a macro-op, along with 3 example tests but no extensive usage. Expected micro-op groups should be provided in the order which micro-ops are held in the `MacroOp` structure.

Code in `RegressionTest.cc` has been refactored to reduce repetition. This is a little hard to decipher purely from the diff provided by GitHub. Similar reductions in repetition have also been performed in the child classes.

I plan to provide more extensive testing in a future PR.

### DIFF
--- a/test/regression/RegressionTest.cc
+++ b/test/regression/RegressionTest.cc
@@ -21,79 +21,23 @@ void RegressionTest::TearDown() {
   }
 }
 
-void RegressionTest::createArchitecture(const char* source, const char* triple,
-                                        const char* extensions) {
-  // Zero-out process memory from any prior runs
-  if (processMemory_ != nullptr)
-    std::memset(processMemory_, '\0', processMemorySize_);
-
-  // Assemble the source to a flat binary
-  assemble(source, triple, extensions);
-  if (HasFatalFailure()) return;
-
-  // Generate the predefined model config
-  generateConfig();
-
-  // Due to SimInfo being static, we need to ensure the config values/options
-  // stored are up-to-date with the latest generated config file
-  simeng::config::SimInfo::reBuild();
-
-  // Create a linux process from the assembled code block.
-  // Memory allocation for process images also takes place
-  // during linux process creation. The Elf binary is parsed
-  // and relevant sections are copied to the process image.
-  // The process image is finalised by the createStack method
-  // which creates and populates the initial process stack.
-  // The created process image can be accessed via a shared_ptr
-  // returned by the getProcessImage method.
-  process_ = std::make_unique<simeng::kernel::LinuxProcess>(
-      simeng::span(reinterpret_cast<const uint8_t*>(code_), codeSize_));
-
-  ASSERT_TRUE(process_->isValid());
-  uint64_t entryPoint = process_->getEntryPoint();
-  processMemorySize_ = process_->getProcessImageSize();
-  // This instance of procImgPtr pointer needs to be shared because
-  // getMemoryValue in RegressionTest.hh uses reference to the class
-  // member processMemory_.
-  std::shared_ptr<char> procImgPtr = process_->getProcessImage();
-  processMemory_ = procImgPtr.get();
-
+void RegressionTest::instantiateMemoryInterfaces() {
   // Create memory interfaces for instruction and data access.
   // For each memory interface, a dereferenced shared_ptr to the
   // processImage is passed as argument.
-  simeng::memory::FlatMemoryInterface instructionMemory(processMemory_,
-                                                        processMemorySize_);
+  instructionMemory_ = std::make_unique<simeng::memory::FlatMemoryInterface>(
+      processMemory_, processMemorySize_);
 
-  std::unique_ptr<simeng::memory::FlatMemoryInterface> flatDataMemory =
-      std::make_unique<simeng::memory::FlatMemoryInterface>(processMemory_,
-                                                            processMemorySize_);
+  flatDataMemory_ = std::make_unique<simeng::memory::FlatMemoryInterface>(
+      processMemory_, processMemorySize_);
 
-  std::unique_ptr<simeng::memory::FixedLatencyMemoryInterface>
-      fixedLatencyDataMemory =
-          std::make_unique<simeng::memory::FixedLatencyMemoryInterface>(
-              processMemory_, processMemorySize_, 4);
-  std::unique_ptr<simeng::memory::MemoryInterface> dataMemory;
-
-  // Create the OS kernel and the process
-  simeng::kernel::Linux kernel(
-      simeng::config::SimInfo::getConfig()["CPU-Info"]["Special-File-Dir-Path"]
-          .as<std::string>());
-  kernel.createProcess(*process_);
-
-  // Populate the heap with initial data (specified by the test being run).
-  ASSERT_LT(process_->getHeapStart() + initialHeapData_.size(),
-            process_->getInitialStackPointer());
-  std::copy(initialHeapData_.begin(), initialHeapData_.end(),
-            processMemory_ + process_->getHeapStart());
-
-  // Create the architecture
-  architecture_ = createArchitecture(kernel);
+  fixedLatencyDataMemory_ =
+      std::make_unique<simeng::memory::FixedLatencyMemoryInterface>(
+          processMemory_, processMemorySize_, 4);
 }
 
-void RegressionTest::run(const char* source, const char* triple,
-                         const char* extensions) {
-  testing::internal::CaptureStdout();
-
+void RegressionTest::createProcess(const char* source, const char* triple,
+                                   const char* extensions) {
   // Zero-out process memory from any prior runs
   if (processMemory_ != nullptr)
     std::memset(processMemory_, '\0', processMemorySize_);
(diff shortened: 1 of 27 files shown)


test/regression/riscv/RISCVRegressionTest.cc
```diff
@@ -3,15 +3,24 @@
 #include "simeng/arch/riscv/Architecture.hh"
 #include "simeng/pipeline/BalancedPortAllocator.hh"
 
+using MacroOp = std::vector<std::shared_ptr<simeng::Instruction>>;
```
dANW34V3R: Now unused. Remove

test/regression/aarch64/MicroOperation.cc
```diff
@@ -10,6 +10,7 @@
 namespace {
 
 using MicroOp = AArch64RegressionTest;
+using namespace simeng::arch::aarch64;
```
dANW34V3R, line 13: Could add `InstructionGroups` to namespace so this doesn't need to be repeatedly qualified in the tests. Thoughts @UoB-HPC/simeng-developers ?

test/regression/aarch64/AArch64RegressionTest.cc
```diff
@@ -3,22 +3,23 @@
 #include "simeng/arch/aarch64/Architecture.hh"
 #include "simeng/pipeline/BalancedPortAllocator.hh"
 
+using MacroOp = std::vector<std::shared_ptr<simeng::Instruction>>;
```
dANW34V3R: No unused, remove

test/regression/aarch64/MicroOperation.cc
dANW34V3R: Doing this makes development much easier so done in #420

test/regression/RegressionTest.cc
```diff
@@ -52,51 +50,71 @@ void RegressionTest::run(const char* source, const char* triple,
...
+  architecture_ = instantiateArchitecture(*kernel_);
+}
 
-  // Create a port allocator for an out-of-order core
-  std::unique_ptr<simeng::pipeline::PortAllocator> portAllocator =
-      createPortAllocator();
+void RegressionTest::instantiateMemoryInterfaces() {
+  // Create memory interfaces for instruction and data access.
+  // For each memory interface, a dereferenced shared_ptr to the
+  // processImage is passed as argument.
```
ABenC377: argument is missing an article ('an' or 'the').

test/regression/RegressionTest.cc
```diff
@@ -52,51 +50,71 @@ void RegressionTest::run(const char* source, const char* triple,
...
   // Create the architecture
-  architecture_ = createArchitecture(kernel);
+  architecture_ = instantiateArchitecture(*kernel_);
+}
 
-  // Create a port allocator for an out-of-order core
-  std::unique_ptr<simeng::pipeline::PortAllocator> portAllocator =
-      createPortAllocator();
+void RegressionTest::instantiateMemoryInterfaces() {
+  // Create memory interfaces for instruction and data access.
```
ABenC377, line 95: There aren't full stops at the end of other comments in this file.  I think we should be consistent one way or the other.

test/regression/RegressionTest.hh
```diff
@@ -71,19 +71,29 @@ class RegressionTest
   /** Generate a default YAML-formatted configuration. */
   virtual void generateConfig() const = 0;
 
+  /** Instantiate the architecture object using the kernel */
```
ABenC377: I think its confusing to say that the createArchitecture function Instantiates the architecture, when there is another function later on called instantiateArchitecture.  Could you Change comment to 'Create the ...'

Also, the other comments in this file have a full stop

test/regression/RegressionTest.hh
```diff
@@ -71,19 +71,29 @@ class RegressionTest
...
+  void createArchitecture(const char* source, const char* triple,
+                          const char* extensions);
+
   /** Run the assembly in `source`, building it for the target `triple` and ISA
    * extensions. */
   void run(const char* source, const char* triple, const char* extensions);
 
   /** Create an ISA instance from a kernel. */
```
ABenC377: As above, I'd change the verb here to 'Instantiate'

test/regression/RegressionTest.hh
```diff
@@ -71,19 +71,29 @@ class RegressionTest
...
-  virtual std::unique_ptr<simeng::arch::Architecture> createArchitecture(
+  virtual std::unique_ptr<simeng::arch::Architecture> instantiateArchitecture(
       simeng::kernel::Linux& kernel) const = 0;
 
   /** Create a port allocator for an out-of-order core model. */
   virtual std::unique_ptr<simeng::pipeline::PortAllocator> createPortAllocator(
       ryml::ConstNodeRef config =
           simeng::config::SimInfo::getConfig()) const = 0;
 
+  /** Predecode the first instruction in source and check the assigned group
```
ABenC377, line 92: full stop

test/regression/RegressionTest.hh
```diff
@@ -108,41 +118,81 @@ class RegressionTest
...
-  /** The process memory. */
-  char* processMemory_ = nullptr;
+ private:
+  /** Assemble test source to a flat binary for the given triple and ISA
+   * extensions. */
+  void assemble(const char* source, const char* triple, const char* extensions);
 
-  /** The size of the process memory in bytes. */
-  size_t processMemorySize_ = 0;
+  /** Instantiate the process from the source bytes for the given architecture.
```
ABenC377: As we are using instantiate and create to mean different things previously, can you change comment to use Create

test/regression/RegressionTest.hh
```diff
@@ -108,41 +118,81 @@ class RegressionTest
...
+   */
+  void createProcess(const char* source, const char* triple,
+                     const char* extensions);
 
-  /** The process that was executed. */
-  std::unique_ptr<simeng::kernel::LinuxProcess> process_;
+  /** Instantiate the kernel object using the process. */
```
ABenC377: As above, change comment to use Create

test/regression/riscv/RISCVRegressionTest.hh
```diff
@@ -53,49 +53,102 @@ inline std::string paramToString(
...
 
-/** The test fixture for all RISCV regression tests. */
+/** A helper macro to predecode the first instruction in a snippet of RISC-V
+ * assembly code and check the assigned group(s) for each micro-op matches the
+ * expected group(s). Returns from the calling function if a fatal error occurs.
+ * Four bytes containing zeros are appended to the source to ensure that the
+ * program will terminate with an unallocated instruction encoding exception
+ * instead of running into the heap.
+ */
+#define EXPECT_GROUP(source, ...)                                   \
```
ABenC377, line 91: Is this the same as the macro defined in the aarch64RegressionTest.hh file?  If so, could it be moved to the generic RegressionTest.hh file to avoid duplication?

test/regression/riscv/RISCVRegressionTest.hh
```diff
@@ -53,49 +53,102 @@ inline std::string paramToString(
...
+
+  /** Run the first instruction in source through predecode and check the
+   * groups. */
+  void checkGroup(const char* source, const std::vector<int> expectedGroups,
+                  bool compressed);
 
   /** Generate a default YAML-formatted configuration. */
   void generateConfig() const override;
 
   /** Create an ISA instance from a kernel. */
```
ABenC377: Again, create versus instantiate

ABenC377 (changes requested):
Looks good!  Mainly minor comments about the use of the words Create and Instantiate throughout.

test/regression/RegressionTest.cc
dANW34V3R: There are a few, mostly from old comments. Unsure of the best course of action as we are very inconsistent throughout the project. @jj16791 @FinnWilkinson

test/regression/riscv/RISCVRegressionTest.hh
dANW34V3R: These are not the same. They call their respective checkGroup function which is ISA specific. Both initialise LLVM for their respective ISA and the RISC-V version takes an extra parameter for compressed instructions

test/regression/RegressionTest.hh
dANW34V3R: Updated

test/regression/RegressionTest.cc
ABenC377: It's not all that important.  Would be nice to have a house preference though

ABenC377 (dismissed):
👌🏻

test/regression/RegressionTest.cc
```diff
@@ -52,51 +50,71 @@ void RegressionTest::run(const char* source, const char* triple,
...
   // member processMemory_.
   std::shared_ptr<char> procImgPtr = process_->getProcessImage();
   processMemory_ = procImgPtr.get();
 
-  // Create memory interfaces for instruction and data access.
-  // For each memory interface, a dereferenced shared_ptr to the
-  // processImage is passed as argument.
-  simeng::memory::FlatMemoryInterface instructionMemory(processMemory_,
-                                                        processMemorySize_);
+  // TODO ensure performing this before kernel_.createProcess is ok - Think it's
```
jj16791: This is valid

test/regression/RegressionTest.cc
```diff
@@ -21,10 +21,8 @@ void RegressionTest::TearDown() {
   }
 }
 
-void RegressionTest::run(const char* source, const char* triple,
-                         const char* extensions) {
-  testing::internal::CaptureStdout();
-
+void RegressionTest::createProcess(const char* source, const char* triple,
```
jj16791: What was the motivation for splitting the creation of simulation into multiple functions? It all seems a bit messy and jumbled to me

test/regression/RegressionTest.hh
```diff
@@ -108,41 +119,82 @@ class RegressionTest
...
+                    const char* extensions);
+
+  /** Instantiate the memory interfaces. */
+  void instantiateMemoryInterfaces();
+
+  /** Instantiating all of the objects used to setup and run the simulation. */
+  void instantiateSimulationObjects(const char* source, const char* triple,
+                                    const char* extensions);
+
+  /* Instantiation of the kernel. */
```
jj16791: "Pointer to the instantiated ...". Same should apply to the below

test/regression/RegressionTest.hh
```diff
@@ -108,41 +119,82 @@ class RegressionTest
   /** The initial data to populate the heap with. */
   std::vector<uint8_t> initialHeapData_;
 
-  /** The maximum number of ticks to run before aborting the test. */
-  uint64_t maxTicks_ = UINT64_MAX;
+  /** The process that was executed. */
```
jj16791: The process to be executed

jj16791 (changes requested):

test/regression/RegressionTest.cc
jj16791: Probably needs a formal discussion but I'd say full stop for only the DOxygen style comments

test/regression/RegressionTest.cc
dANW34V3R: To check what group an instruction is assigned `architecture->predecode` is called. "RegressionTest::checkGroup::184". To do this the architecture needs to be instantiated.

Previously I just had a new function with all the code up to the architecture instantiation repeated from the old `run` function. The current code is an attempt to reduce the repetition with the side effect of also breaking up the old massive function. 

I think the chained calling is a little odd but the hope was that you wouldn't have to orchestrate the instantiations yourself as they need to happen in a specific order e.g. you need `kernel_` before you can instantiate the architecture.

Looking again it does seem a little messy and not consistent. I can move the BP and core instantiations into their own functions as well which helps a little (should have just pushed). Other thoughts on cleanup?

tldr, both `run` and `checkGroup` need to instantiate some simulation objects and this reduces repeated code

jj16791 (dismissed):

test/regression/RegressionTest.cc
jj16791: After offline discussion, changes made look a lot better

test/regression/RegressionTest.cc
```diff
@@ -143,6 +156,24 @@ void RegressionTest::run(const char* source, const char* triple,
   programFinished_ = true;
 }
 
+void RegressionTest::checkGroup(const char* source, const char* triple,
+                                const char* extensions,
+                                const std::vector<int> expectedGroups) {
```
FinnWilkinson: In both AArch64 and RV Instruction groups are defined as `uint16_t` so the vector should reflect this too

test/regression/riscv/instructions/atomic.cc
FinnWilkinson, line 1: These all seem to have the group of only `LOAD_INT`. Are there no atomic instructions which perform a LOAD and STORE? i.e. a swap or a load-action-store?

FinnWilkinson (changes requested):
Looks pretty good, just a few minor points

JosephMoore25 (dismissed):
Clean PR and implementation. Makes it easy to add group tests, and the background logic all looks sound. Nice work

test/regression/riscv/instructions/atomic.cc
dANW34V3R: Comment added explaining current functionality but more discussion needed about how to handle this long term

JosephMoore25 (approved):

FinnWilkinson (approved):
Looks good to me!

jj16791 (approved):

ABenC377 (approved):

jj16791 (approved):

PR #425, opened 2024-08-30
ACFL23 Instruction Support

JosephMoore25:
Merging work done towards enabling support for a few codes for ACFL 23, namely STREAM, Minibude, Cloverleaf, Tealeaf, and Minisweep.

This PR is mostly made up of added instruction support. 58 instructions have been added, with 24 unique instructions with the remainder being variants. Most instructions are SVE, with some NEON added.

An additional feature of "infinite loop checking" has been added. This adds a counter in the ROB which throws an error if the same address has been at the head of the ROB for a **very** long time. This catches a few errors previously found where an erroneous config or broken logic can cause SimEng to get caught in a loop and sometimes eventually hit OOM.

This also fixes an OpenMP bug that has previously popped up for ACFL 23 support, work that Jack had done in a separate branch.

Tests are still being added, and the new group tests need to be added for all instructions. The PR will leave draft stage once all tests have been added.

Here are a list of instructions added:

| Opcode                           	| Inst Format                               	| General Test added? 	| Group Test added? 	|
|----------------------------------	|-------------------------------------------	|---------------------	|-------------------	|
| Opcode::AArch64_UADDLVv8i8v: {   	| // uaddlv hd, vn.8b                       	|         Yes         	|         Yes         	|
| Opcode::AArch64_FTSMUL_ZZZ_S: {  	| // ftsmul zd.s, zn.s, zm.s                		|         Yes         	|         Yes         	|
| Opcode::AArch64_FTSMUL_ZZZ_D: {  	| // ftsmul zd.d, zn.d, zm.d                		|         Yes         	|         Yes         	|
| Opcode::AArch64_FTSSEL_ZZZ_S: {  	| // ftssel zd.s, zn.s, zm.s                	|         Yes         	|         Yes         	|
| Opcode::AArch64_FTSSEL_ZZZ_D: {  	| // ftssel zd.d, zn.d, zm.d                	|         Yes         	|         Yes         	|| Opcode::AArch64_FTMAD_ZZI_S: {   	| // ftmad zd.s, zn.s, zm.s, #imm           	|         Yes         	|         Yes         	|
| Opcode::AArch64_FTMAD_ZZI_D: {   	| // ftmad zd.s, zn.s, zm.s, #imm           	|         Yes         	|         Yes         	|
| Opcode::AArch64_CMEQv2i32rz: {   	| // cmeq vd.2s, vn.2s, #0                  	|         Yes         	|         Yes         	|
| Opcode::AArch64_CMHIv2i32: {     	| // cmhi vd.2s, vn.2s, vm.2s               	|         Yes         	|         Yes         	|
| Opcode::AArch64_CMPHS_PPzZZ_B: { 	| // cmphs pd.b, pg/z, zn.b, zm.b           	|         Yes         	|         Yes         	|
| Opcode::AArch64_CMPHS_PPzZZ_D: { 	| // cmphs pd.d, pg/z, zn.d, zm.d           	|         Yes         	|         Yes         	|
| Opcode::AArch64_CMPHS_PPzZZ_H: { 	| // cmphs pd.h, pg/z, zn.h, zm.h           	|         Yes         	|         Yes         	|
| Opcode::AArch64_CMPHS_PPzZZ_S: { 	| // cmphs pd.s, pg/z, zn.s, zm.s           	|         Yes         	|         Yes         	|
| Opcode::AArch64_CPY_ZPmV_B: {    	| // cpy zd.b, pg/m, vn.b                   	|         Yes         	|         Yes         	|
| Opcode::AArch64_CPY_ZPmV_D: {    	| // cpy zd.d, pg/m, vn.d                   	|         Yes         	|         Yes         	|
| Opcode::AArch64_CPY_ZPmV_H: {    	| // cpy zd.h, pg/m, vn.h                   	|         Yes         	|         Yes         	|
| Opcode::AArch64_CPY_ZPmV_S: {    	| // cpy zd.s, pg/m, vn.s                   	|         Yes         	|         Yes         	|
| Opcode::AArch64_FDIVv4f32: {     	| // fdiv vd.4s, vn.4s, vm.4s               	|         Yes         	|         Yes         	|
| Opcode::AArch64_LASTB_VPZ_D: {   	| // lastb dd, pg, zn.d                     	|         Yes         	|         Yes         	|
| Opcode::AArch64_LASTB_VPZ_S: {   	| // lastb sd, pg, zn.s                     	|         Yes         	|         Yes         	|
| Opcode::AArch64_LASTB_VPZ_H: {   	| // lastb hd, pg, zn.h                     	|         Yes         	|         Yes         	|
| Opcode::AArch64_LASTB_VPZ_B: {   	| // lastb bd, pg, zn.b                     	|         Yes         	|         Yes         	|
| Opcode::AArch64_CLASTB_VPZ_D: {  	| // clastb dd, pg, dn, zn.d                	|         Yes         	|         Yes         	|
| Opcode::AArch64_CLASTB_VPZ_S: {  	| // clastb sd, pg, sn, zn.s                	|         Yes         	|         Yes         	|
| Opcode::AArch64_CLASTB_VPZ_H: {  	| // clastb hd, pg, hn, zn.h                	|         Yes         	|         Yes         	|
| Opcode::AArch64_CLASTB_VPZ_B: {  	| // clastb bd, pg, bn, zn.b                	|         Yes         	|         Yes         	|
| Opcode::AArch64_LDAXRB: {        	| // ldaxrb wt, [xn]                        	|         Yes         	|         Yes         	|
| Opcode::AArch64_LDRSWroW: {      	| // ldrsw xt, [xn, wm, {extend {#amount}}] 	|         Yes         	|         Yes         	|
| Opcode::AArch64_ORNv8i8: {       	| // orn vd.8b, vn.8b, vn.8b                	|         Yes         	|         Yes         	|
| Opcode::AArch64_PFIRST_B: {      	| // pfirst pdn.b, pg, pdn.b                		|         Yes         	|         Yes         	|
| Opcode::AArch64_PNEXT_B: {       	| // pnext pdn.b, pv, pdn.b                 	|         Yes         	|         Yes         	|
| Opcode::AArch64_PNEXT_H: {       	| // pnext pdn.h, pv, pdn.h                 	|         Yes         	|         Yes         	|
| Opcode::AArch64_PNEXT_S: {       	| // pnext pdn.s, pv, pdn.s                 	|         Yes         	|         Yes         	|
| Opcode::AArch64_PNEXT_D: {       	| // pnext pdn.d, pv, pdn.d                 	|         Yes         	|         Yes         	|
| Opcode::AArch64_SMAX_ZI_D: {     	| // smax zdn.d, zdn.d, #imm                	|         Yes         	|         Yes         	|
| Opcode::AArch64_SMAX_ZI_H: {     	| // smax zdn.h, zdn.h, #imm                	|         Yes         	|         Yes         	|
| Opcode::AArch64_SMAX_ZI_B: {     	| // smax zdn.b, zdn.b, #imm                	|         Yes         	|         Yes         	|
| Opcode::AArch64_SMAX_ZPmZ_D: {   	| // smax zd.d, pg/m, zn.d, zm.d            	|         Yes         	|         Yes         	|
| Opcode::AArch64_SMAX_ZPmZ_H: {   	| // smax zd.h, pg/m, zn.h, zm.h            	|         Yes         	|         Yes         	|
| Opcode::AArch64_SMAX_ZPmZ_B: {   	| // smax zd.b, pg/m, zn.b, zm.b            	|         Yes         	|         Yes         	|
| Opcode::AArch64_SMINV_VPZ_D: {   	| // sminv sd, pg, zn.d                     	|         Yes         	|         Yes         	|
| Opcode::AArch64_SMINV_VPZ_H: {   	| // sminv sd, pg, zn.h                     	|         Yes         	|         Yes         	|
| Opcode::AArch64_SMINV_VPZ_B: {   	| // sminv sd, pg, zn.b                     |         Yes         |         Yes         	|
| Opcode::AArch64_SMIN_ZPmZ_D: {   	| // smin zd.d, pg/m, zn.d, zm.d            	|         Yes         	|         Yes         	|
| Opcode::AArch64_SMIN_ZPmZ_H: {   	| // smin zd.h, pg/m, zn.h, zm.h            	|         Yes         	|         Yes         	|
| Opcode::AArch64_SMIN_ZPmZ_B: {   	| // smin zd.b, pg/m, zn.b, zm.b            	|         Yes         	|         Yes         	|
| Opcode::AArch64_SPLICE_ZPZ_D: {  	| // splice zdn.d, pv, zdn.t, zm.d          		|         Yes         	|         Yes         	|
| Opcode::AArch64_SPLICE_ZPZ_S: {  	| // splice zdn.s, pv, zdn.t, zm.s          		|         Yes         	|         Yes         	|
| Opcode::AArch64_STLXRB:          	| // stlxrb ws, wt, [xn]                    	|         Yes         	|         Yes         	|
| Opcode::AArch64_STLXRH:          	| // stlxrh ws, wt, [xn]                    	|         Yes         	|         Yes         	|
| Opcode::AArch64_STLXR:          	| // stlxrb ws, {w,x}t, [xn]                    	|         Yes         	|         Yes         	|
| Opcode::AArch64_UMAXVv16i8v: {   	| // umaxv bd, vn.16b                       	|         Yes         	|         Yes         	|
| Opcode::AArch64_UMAXVv4i16v: {   	| // umaxv hd, vn.4h                        	|         Yes         	|         Yes         	|
| Opcode::AArch64_UMAXVv4i32v: {   	| // umaxv sd, vn.4s                        	|         Yes         	|         Yes         	|
| Opcode::AArch64_UMAXVv8i16v: {   	| // umaxv hd, vn.8h                        	|         Yes         	|         Yes         	|
| Opcode::AArch64_UMAXVv8i8v: {    	| // umaxv bd, vn.8b                        	|         Yes         	|         Yes         	|
| Opcode::AArch64_WHILELS_PXX_B: { 	| // whilels pd.b, xn, xm                   	|         Yes         	|         Yes         	|
| Opcode::AArch64_WHILELS_PXX_D: { 	| // whilels pd.d, xn, xm                   	|         Yes         	|         Yes         	|
| Opcode::AArch64_WHILELS_PXX_H: { 	| // whilels pd.h, xn, xm                   	|         Yes         	|         Yes         	|
| Opcode::AArch64_WHILELS_PXX_S: { 	| // whilels pd.s, xn, xm                   |         Yes         	|         Yes         	|


src/include/simeng/pipeline/ReorderBuffer.hh
```diff
@@ -125,6 +125,14 @@ class ReorderBuffer {
...
   uint64_t pc_;
 
+  /** The address of the last instruction at the head of the ROB to check if
+   * it's stuck */
+  uint64_t last_inst_addr = 0;
+
+  /** A counter for how many cycles the same instruction has been at the head of
+   * the ROB */
+  uint64_t inst_repeat_counter = 0;
+
```
FinnWilkinson: Class variables should use camelCase and end with `_`

src/include/simeng/pipeline/ReorderBuffer.hh
```diff
@@ -125,6 +125,14 @@ class ReorderBuffer {
    */
   uint64_t pc_;
 
+  /** The address of the last instruction at the head of the ROB to check if
+   * it's stuck */
+  uint64_t last_inst_addr = 0;
```
FinnWilkinson: Is this the instruction currently at the head of the ROB? Or the last instruction to be retired?

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -429,15 +429,16 @@ bool ExceptionHandler::init() {
                       << std::endl;
             return fatal();
           }
-          uint64_t retval = (pid == 0) ? 1 : 0;
-          stateChange = {ChangeType::REPLACEMENT, {R0}, {retval}};
-          stateChange.memoryAddresses.push_back({mask, 1});
+          uint64_t retval = static_cast<uint64_t>(bitmask);
+          stateChange = {ChangeType::REPLACEMENT, {R0}, {sizeof(retval)}};
+          stateChange.memoryAddresses.push_back({mask, 8});
```
FinnWilkinson: Based on comments from other PRs, we should probably move away from using numbers for sizes and use `sizeof` consistently throughout. i.e. update `stateChange.memoryAddresses.push_back({mask, 8});` to use appropriate sizeof

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -429,15 +429,16 @@ bool ExceptionHandler::init() {
...
           stateChange.memoryAddressValues.push_back(bitmask);
         } else {
           stateChange = {ChangeType::REPLACEMENT, {R0}, {-1ll}};
         }
         break;
       }
+
```
FinnWilkinson: Remove newline

src/lib/arch/aarch64/InstructionMetadata.cc
```diff
@@ -504,6 +556,17 @@ InstructionMetadata::InstructionMetadata(const cs_insn& insn)
...
       break;
     }
+    case Opcode::AArch64_FTMAD_ZZI_D:
+      [[fallthrough]];
+    case Opcode::AArch64_FTMAD_ZZI_S: {
+      // No defined access types
+      operands[0].access = CS_AC_READ | CS_AC_WRITE;
+      operands[1].access = CS_AC_READ;
+      operands[2].access = CS_AC_READ;
+      operands[3].access = CS_AC_READ;
```
FinnWilkinson: operands of type `ARM64_OP_IMM` don't need access type defined

src/lib/arch/aarch64/InstructionMetadata.cc
```diff
@@ -1660,6 +1731,48 @@ InstructionMetadata::InstructionMetadata(const cs_insn& insn)
...
+      [[fallthrough]];
+    case Opcode::AArch64_LASTB_VPZ_S:
+      [[fallthrough]];
+    case Opcode::AArch64_LASTB_VPZ_H:
+      [[fallthrough]];
+    case Opcode::AArch64_LASTB_VPZ_B: {
+      operands[0].access = CS_AC_WRITE;
+      operands[1].access = CS_AC_READ;
+      operands[2].access = CS_AC_READ;
+    }
```
FinnWilkinson: Case doesn't have break - is this intensional?

src/lib/arch/aarch64/InstructionMetadata.cc
```diff
@@ -2406,6 +2526,19 @@ void InstructionMetadata::revertAliasing() {
...
+          opcode == Opcode::AArch64_EXTRXrri) {
+        // ror wd, ws, #shift; alias for : extr wd, ws, ws, #shift
+        operandCount = 4;
+        operands[3] = operands[2];
+        operands[2] = operands[1];
+
+        operands[2].type = ARM64_OP_REG;
+        operands[2].access = CS_AC_READ;
+        operands[3].type = ARM64_OP_IMM;
+        operands[3].access = CS_AC_READ;
```
FinnWilkinson: operand of type `ARM64_OP_IMM` doesn't need access defined

src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -81,6 +81,20 @@ unsigned int ReorderBuffer::commit(uint64_t maxCommitSize) {
...
+      inst_repeat_counter = 0;
+    }
+    if (inst_repeat_counter > 10000000) {
+      std::cout
+          << "Infinite loop detected in rob commit at instruction address "
+          << std::hex << uop->getInstructionAddress() << std::dec << " ("
+          << uop->getMicroOpIndex() << "). Killing.\n";
+      exit(1);
+    }
+    last_inst_addr = uop->getInstructionAddress();
```
FinnWilkinson: Can this infinite loop logic go inside the ` if (!uop->canCommit())` clause? AFAIK only if the head of the ROB is stuck in a non-commitable state can an infinite loop occur

src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -81,6 +81,20 @@ unsigned int ReorderBuffer::commit(uint64_t maxCommitSize) {
...
+    if (uop->getInstructionAddress() == last_inst_addr) {
+      inst_repeat_counter++;
+    } else {
+      inst_repeat_counter = 0;
+    }
+    if (inst_repeat_counter > 10000000) {
+      std::cout
+          << "Infinite loop detected in rob commit at instruction address "
+          << std::hex << uop->getInstructionAddress() << std::dec << " ("
+          << uop->getMicroOpIndex() << "). Killing.\n";
```
FinnWilkinson: Printout needs prefix of `[SimEng:ReorderBuffer]`

src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -81,6 +81,20 @@ unsigned int ReorderBuffer::commit(uint64_t maxCommitSize) {
   unsigned int n;
   for (n = 0; n < maxCommits; n++) {
     auto& uop = buffer_[0];
+    if (uop->getInstructionAddress() == last_inst_addr) {
+      inst_repeat_counter++;
+    } else {
+      inst_repeat_counter = 0;
+    }
+    if (inst_repeat_counter > 10000000) {
```
FinnWilkinson: Wonder if we should parameterise this limit somehow? Maybe in config? @jj16791 @JosephMoore25 @dANW34V3R @ABenC377

test/regression/aarch64/instructions/load.cc
```diff
@@ -695,6 +696,70 @@ TEST_P(InstLoad, ldarb) {
...
+    add sp, sp, #32
+    ldaxrb w7, [sp]
+  )");
+
+  EXPECT_EQ(getGeneralRegister<uint32_t>(4), 16);
+  EXPECT_EQ(getGeneralRegister<uint32_t>(5), 32);
+  EXPECT_EQ(getGeneralRegister<uint32_t>(6), 48);
+  EXPECT_EQ(getGeneralRegister<uint32_t>(7), 64);
+
+  EXPECT_GROUP(R"(ldaxrb w7, [sp])", LOAD_INT);
```
FinnWilkinson: Whats the difference between these two cases? i.e. why two sets of `RUN_AARCH64`?

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -114,6 +114,33 @@ RegisterValue sveAddvPredicated(srcValContainer& sourceValues,
   return {out, 256};
 }
 
+/** Helper function for NEON instructions with the format `uaddlv Vd, Vn.T`.
+ * T represents the type of the destination register (e.g. for h0, T =
+ * uint32_t). U represents the type of the sourceValues[0] (e.g. for v0.8b, U =
+ * uint8_t) Returns correctly formatted RegisterValue. */
```
FinnWilkinson: Very nit-picky (sorry) but to help with function comment readability each templated variable and return statement should begin a new line. Exists throughout. 

For example:
```C++
/** Helper function for NEON instructions with the format `uaddlv Vd, Vn.T`.
 * T represents the type of the destination register (e.g. for h0, T =
 * uint32_t). 
 * U represents the type of the sourceValues[0] (e.g. for v0.8b, U =
 * uint8_t) 
 * Returns correctly formatted RegisterValue. */
```

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -257,6 +284,32 @@ RegisterValue sveCpy_imm(
   return {out, 256};
 }
 
+/** Helper function for SVE instructions with the format `cpy zd, pg/m, vn
```
FinnWilkinson: would suggest changing `vn` to `rn` as to me `v` suggests a vector register, whereas `r` makes it clearer its a scalar register. Applies to variable name in function also

src/include/simeng/arch/aarch64/helpers/sve.hh
FinnWilkinson: This particular comment also uses `V` rather than `z` for the register

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -114,6 +114,33 @@ RegisterValue sveAddvPredicated(srcValContainer& sourceValues,
...
+  T out = 0;
+  for (int i = 0; i < I; i++) {
+    out += n[i];
+  }
+  return {out, 256};
+}
+
+/** Helper function for NEON instructions with the format `umaxv Vd, Vn.T`.
+ * T represents the type of sourceValues (e.g. for vn.s, T = uint32_t).
+ * Returns correctly formatted RegisterValue. */
```
FinnWilkinson: Comment uses `V` rather than `z` for operand registers

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -934,6 +1110,67 @@ RegisterValue sveIndex(
   return {out, 256};
 }
 
+/** Helper function for SVE instructions with the format `lastb vd, pg, zn`.
+ * T represents the vector register type (e.g. zd.d would be uint64_t).
+ * Returns correctly formatted RegisterValue. */
+template <typename T>
+RegisterValue sveLastBScalar(srcValContainer& sourceValues,
+                             const uint16_t VL_bits) {
+  // sourceValues are wrong and the correct value is in the previous index.
```
FinnWilkinson: If the source values are wrong, then the instruction metadata is incorrect as the destination operand should be read only as per the spec

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1291,6 +1528,69 @@ std::array<uint64_t, 4> svePsel(
   return out;
 }
 
+/** Helper function for SVE instructions with the format `pfirst pdn, pg, pdn`.
+ * Returns an array of 4 uint64_t elements. */
+std::array<uint64_t, 4> svePfirst(srcValContainer& sourceValues,
+                                  const uint16_t VL_bits) {
+  const uint16_t partition_num = VL_bits / 8;
+  // sourceValues are wrong and the correct value is in the previous index.
```
FinnWilkinson: Confused by this comment?

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -934,6 +1110,67 @@ RegisterValue sveIndex(
   return {out, 256};
 }
 
+/** Helper function for SVE instructions with the format `lastb vd, pg, zn`.
```
FinnWilkinson: Change `vd` to `rd` as NEON vector register not used

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -934,6 +1110,67 @@ RegisterValue sveIndex(
...
+    }
+    // If no active lane has been found, select highest element instead
+    if (i == 0) lastElem = partition_num - 1;
+  }
+
+  out = n[lastElem];
+  return {out, 256};
+}
+
+/** Helper function for SVE instructions with the format `clastb vd, pg, vd,
```
FinnWilkinson: Change `vd` to `rd` as NEON reigster not used

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -934,6 +1110,67 @@ RegisterValue sveIndex(
...
+ * zn`. T represents the vector register type (e.g. zd.d would be uint64_t).
+ * Returns correctly formatted RegisterValue. */
+template <typename T>
+RegisterValue sveCLastBScalar(srcValContainer& sourceValues,
+                              const uint16_t VL_bits) {
+  // sourceValues are wrong and the correct value is in the previous index.
+  const uint64_t* p = sourceValues[1].getAsVector<uint64_t>();
+  const uint64_t* m = sourceValues[2].getAsVector<uint64_t>();
```
FinnWilkinson: Given destination / first (non predicate) source regiser can be W or X reg, we shouldn't be always getting as uint64_t and should be using `get<>` not `getAsVector<>`

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -934,6 +1110,67 @@ RegisterValue sveIndex(
...
+      lastElem = i;
+      break;
+    }
+  }
+
+  if (lastElem < 0) {
+    out = static_cast<uint64_t>(static_cast<T>(m[0]));
+  } else {
+    out = static_cast<uint64_t>(static_cast<T>(n[lastElem]));
+  }
```
FinnWilkinson: We should be 0 extending not casting. I think this functionality is baked into RegisterValue?

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -934,6 +1110,67 @@ RegisterValue sveIndex(
...
+  return {out, 256};
+}
+
+/** Helper function for SVE instructions with the format `clastb vd, pg, vd,
+ * zn`. T represents the vector register type (e.g. zd.d would be uint64_t).
+ * Returns correctly formatted RegisterValue. */
+template <typename T>
+RegisterValue sveCLastBScalar(srcValContainer& sourceValues,
+                              const uint16_t VL_bits) {
+  // sourceValues are wrong and the correct value is in the previous index.
```
FinnWilkinson: Confused by this comment

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1291,6 +1528,69 @@ std::array<uint64_t, 4> svePsel(
   return out;
 }
 
+/** Helper function for SVE instructions with the format `pfirst pdn, pg, pdn`.
```
FinnWilkinson: add `.b` after `pdn` for clarity that it only operates on byte predicates

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1291,6 +1528,69 @@ std::array<uint64_t, 4> svePsel(
...
+ * Returns an array of 4 uint64_t elements. */
+std::array<uint64_t, 4> svePfirst(srcValContainer& sourceValues,
```
FinnWilkinson: This instruction should also update the NZCV register

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1291,6 +1528,69 @@ std::array<uint64_t, 4> svePsel(
...
+    const uint16_t VL_bits) {
+  const uint16_t partition_num = VL_bits / (sizeof(T) * 8);
+  const uint64_t* p = sourceValues[1].getAsVector<uint64_t>();
+  const uint64_t* dn = sourceValues[2].getAsVector<uint64_t>();
+  // Set destination elements to 0
+  std::array<uint64_t, 4> out = {0, 0, 0, 0};
+
+  // Get pattern
+  const uint16_t count =
+      sveGetPattern(metadata.operandStr, sizeof(T) * 8, VL_bits);
```
FinnWilkinson: pnext doesn't ever have a pattern so this isn't necessary. Can just use `partition_num` instead

ABenC377 (commented):
Code all looks good.  Though, I'm not sure about the infinite loop checker.  If I've missed a discussion about this then please ignore me.  But if it is being added as a work around for problems encountered because of erroneous configs or broken logic, shouldn't we be fixing the causes not the symptoms?  Erroneous configs are kind of a user error, but we could update the documentation to help them avoid this, and if there is broken logic in SimEng we should be fixing it not plastering over the resulting problem.  I suppose I can see the value of this type of check in debug mode to flag a problem to the user if they are running into issues in release, but it seems like unnecessary overhead for Release.

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1291,6 +1528,69 @@ std::array<uint64_t, 4> svePsel(
...
+      out[i / 64] |= shifted_active;
+      break;
+    }
+  }
+  return out;
+}
+
+/** Helper function for SVE instructions with the format `pnext pdn, pv, pdn`.
+ * Returns an array of 4 uint64_t elements, and updates the NZCV flags. */
+template <typename T>
```
FinnWilkinson: Need to add description of what T represents to comment

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1291,6 +1528,69 @@ std::array<uint64_t, 4> svePsel(
...
+  const uint64_t* dn = sourceValues[2].getAsVector<uint64_t>();
+  // Set destination elements to 0
+  std::array<uint64_t, 4> out = {0, 0, 0, 0};
+
+  // Get pattern
+  const uint16_t count =
+      sveGetPattern(metadata.operandStr, sizeof(T) * 8, VL_bits);
+
+  // Exit early if count == 0
+  if (count == 0) return {out, getNZCVfromPred(out, VL_bits, sizeof(T))};
```
FinnWilkinson: This can never be triggered

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1291,6 +1528,69 @@ std::array<uint64_t, 4> svePsel(
...
+  // Get last active element of dn.pattern
+  int lastElem = -1;
+  for (int i = partition_num - 1; i >= 0; i--) {
+    if (i < count) {
```
FinnWilkinson: Don't think this if will ever not be true

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -4002,11 +4137,49 @@ void Instruction::execute() {
...
+      }
       case Opcode::AArch64_PFALSE: {  // pfalse pd.b
         uint64_t out[4] = {0, 0, 0, 0};
         results_[0] = out;
         break;
       }
+      case Opcode::AArch64_PFIRST_B: {  // pfirst pdn.b, pg, pdn.b
+        results_[0] = svePfirst(sourceValues_, VL_bits);
+        break;
+      }
```
FinnWilkinson: pfirst should update NZCV

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -4931,16 +5166,17 @@ void Instruction::execute() {
       }
       case Opcode::AArch64_STLRW:    // stlr wt, [xn]
       case Opcode::AArch64_STLRX: {  // stlr xt, [xn]
-        // STORE
```
FinnWilkinson: undo deletion of comment

FinnWilkinson (changes requested):
Most bits look good. Comments mainly about adhearing to the project's style and some confusion on SVE helpers

src/include/simeng/pipeline/ReorderBuffer.hh
JosephMoore25: Resolved by removing variable entirely through new implementation

src/lib/pipeline/ReorderBuffer.cc
JosephMoore25: Implemented this which ended up being a much cleaner option - thanks for the spot

src/lib/pipeline/ReorderBuffer.cc
JosephMoore25: Definitely worth a brief discussion. Perhaps worth a config option that defaults to a large value (10 or 100 million cycles) that we don't explicitly include in the provided configurations. Just in case someone wants to try out some crazy memory backends with huge latencies. If necessary, I can always experiment with the longest no. cycles an instruction in a bottlenecked CPU (slow memory, small frontend) stays at the head of the ROB, though I'd expect this is for a few thousand cycles and nowhere near 10 million.

src/lib/arch/aarch64/InstructionMetadata.cc
JosephMoore25: Fixed this in my latest commit

test/regression/aarch64/instructions/load.cc
JosephMoore25: Must have accidentally written two sets of tests. Reduced down to one

src/include/simeng/arch/aarch64/helpers/sve.hh
JosephMoore25: Fixed style issues in all helpers added in this PR

src/include/simeng/arch/aarch64/helpers/sve.hh
JosephMoore25: Noticed this separate to this comment and fixed

src/include/simeng/arch/aarch64/helpers/sve.hh
JosephMoore25: Checked RegisterValue class and this is the case. Removed any casting

src/include/simeng/arch/aarch64/helpers/sve.hh
JosephMoore25: Checked the spec + pseudocode and you're right. Removed any references to `count` and aligned with the implementation in the spec

test/regression/aarch64/Syscall.cc
```diff
@@ -1080,7 +1080,7 @@ TEST_P(Syscall, sched_getaffinity) {
     )");
   EXPECT_EQ(getGeneralRegister<int64_t>(21), -1);
   EXPECT_EQ(getGeneralRegister<int64_t>(22), -1);
-  EXPECT_EQ(getGeneralRegister<int64_t>(23), 1);
```
dANW34V3R, line 1083: What has caused this to change?

src/include/simeng/arch/aarch64/helpers/neon.hh
```diff
@@ -558,6 +558,35 @@ RegisterValue vecLogicOp_3vecs(srcValContainer& sourceValues,
   return {out, 256};
 }
 
+/** Helper function for NEON instructions with the format `uaddlv zd, zn.T`.
```
FinnWilkinson: Comment uses `z` regs rather than `r`/`v` respectively

src/include/simeng/arch/aarch64/helpers/neon.hh
```diff
@@ -558,6 +558,35 @@ RegisterValue vecLogicOp_3vecs(srcValContainer& sourceValues,
...
+ * T represents the type of the destination register (e.g. for h0, T =
+ * uint32_t).
+ * U represents the type of the sourceValues[0] (e.g. for v0.8b, U =
+ * uint8_t)
+ * Returns correctly formatted RegisterValue. */
+template <typename T, typename U, int I>
```
FinnWilkinson, line 569: Nothing in comment about what `I` corresponds to

src/include/simeng/arch/aarch64/helpers/neon.hh
```diff
@@ -558,6 +558,35 @@ RegisterValue vecLogicOp_3vecs(srcValContainer& sourceValues,
...
+    out += n[i];
+  }
+  return {out, 256};
+}
+
+/** Helper function for NEON instructions with the format `umaxv Vd, Vn.T`.
+ * T represents the type of sourceValues (e.g. for vn.s, T = uint32_t).
+ * Returns correctly formatted RegisterValue. */
+template <typename T, int I>
+RegisterValue vecUMaxV(srcValContainer& sourceValues) {
```
FinnWilkinson: Nothing in comment about what `I` corresponds to, and `vd` should probably be `rd` as it isn't a vector register but a scalar register

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -849,6 +875,130 @@ RegisterValue sveFsqrtPredicated_2vecs(srcValContainer& sourceValues,
   return {out, 256};
 }
 
+/** Helper function for SVE instructions with the format `ftsmul zd, zn, zm`.
+ * T represents the type of sourceValues (e.g. for zn.d, T = double).
+ * Returns correctly formatted RegisterValue. U represents the same precision as
+ * T, but as an integer type for the second source register. */
```
FinnWilkinson: Odd/inconsistent comment formatting

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -849,6 +875,130 @@ RegisterValue sveFsqrtPredicated_2vecs(srcValContainer& sourceValues,
...
+    out[i] = std::abs(out[i]) * sign_bit;
+  }
+
+  return {out, 256};
+}
+
+/** Helper function for SVE instructions with the format `ftssel zd, zn, zm`.
+ * T represents the type of sourceValues (e.g. for zn.d, T = double).
+ * Returns correctly formatted RegisterValue. U represents the same precision as
+ * T, but as an integer type for the second source register. */
```
FinnWilkinson: Odd/inconsistent comment formatting

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -934,6 +1084,66 @@ RegisterValue sveIndex(
...
+
+/** Helper function for SVE instructions with the format `clastb rd, pg, rd,
+ * zn`.
+ * T represents the vector register type (e.g. zd.d would be uint64_t).
+ * Returns correctly formatted RegisterValue. */
+template <typename T>
+RegisterValue sveCLastBScalar(srcValContainer& sourceValues,
+                              const uint16_t VL_bits) {
+  const uint64_t* p = sourceValues[1].getAsVector<uint64_t>();
+  const uint64_t m = sourceValues[2].get<T>();
```
FinnWilkinson: Does `m` always want to be cast from `T` to uint64_t?

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -934,6 +1084,66 @@ RegisterValue sveIndex(
...
+      break;
+    }
+  }
+
+  if (lastElem < 0) {
+    out = m;
+  } else {
+    out = n[lastElem];
+  }
+  return {out, 256};
```
FinnWilkinson, line 1145: `clastb` writes to a `X` or `W` register and so the returned value should be `{out, 8}`

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -934,6 +1084,66 @@ RegisterValue sveIndex(
...
+    uint64_t shifted_active = 1ull << ((i % (64 / sizeof(T))) * sizeof(T));
+    if (p[i / (64 / sizeof(T))] & shifted_active) {
+      lastElem = i;
+      break;
+    }
+    // If no active lane has been found, select highest element instead
+    if (i == 0) lastElem = partition_num - 1;
+  }
+
+  out = n[lastElem];
```
FinnWilkinson: `out` variable not needed

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -934,6 +1084,66 @@ RegisterValue sveIndex(
...
+  return {out, 256};
```
FinnWilkinson: `lastb` writes to a `X` or `W` register and so the returned value should be `{out, 8}`

src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -81,9 +81,21 @@ unsigned int ReorderBuffer::commit(uint64_t maxCommitSize) {
...
+
     if (!uop->canCommit()) {
+      // If an instruction has been stuck at the head of the rob for
+      // sufficiently long, assume an error in SimEng has occured.
+      robHeadRepeatCounter_++;
+      if (robHeadRepeatCounter_ > 10000000) {
+        std::cerr << "[SimEng:ReorderBuffer] Infinite loop detected in rob "
+                     "commit at instruction address "
+                  << std::hex << uop->getInstructionAddress() << std::dec
+                  << " (" << uop->getMicroOpIndex() << ")." << std::endl;
```
FinnWilkinson: Whats the rational for printing the Micro-op index?

FinnWilkinson (changes requested):
All looking pretty good - just a few changes needed and some pedantic comments on comments 😅

src/include/simeng/arch/aarch64/helpers/sve.hh
JosephMoore25: Missed this - good spot!

src/include/simeng/arch/aarch64/helpers/sve.hh
JosephMoore25: There was some confusion over the implemented instruction. I have implemented a helper for `CLASTB (SIMD&FP scalar)` which I named `sveCLastBScalar`. There is another instruction (that I hadn't seen initially) which is `CLASTB (scalar)` which has NOT been implemented yet, and this confuses the names. The implemented instruction lines up with the SIMD&FP one which does in fact write to an SVE vector register. I have renamed the function to `sveCLastBSimdScalar` to clear this up.

src/lib/pipeline/ReorderBuffer.cc
JosephMoore25: This may give additional context to the user what exactly is stuck at the head of the ROB if the instruction is uopd. I have updated the comment generally, though we should have a discussion offline on what exactly we want to print out in one of these cases.

src/include/simeng/arch/aarch64/helpers/sve.hh
JosephMoore25: See `clastb` conversation - exactly the same case happened here so function renamed accordingly.

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -429,9 +429,9 @@ bool ExceptionHandler::init() {
                       << std::endl;
             return fatal();
           }
-          uint64_t retval = (pid == 0) ? 1 : 0;
-          stateChange = {ChangeType::REPLACEMENT, {R0}, {retval}};
-          stateChange.memoryAddresses.push_back({mask, 1});
+          uint64_t retval = static_cast<uint64_t>(bitmask);
+          stateChange = {ChangeType::REPLACEMENT, {R0}, {sizeof(retval)}};
```
dANW34V3R, line 433: The man page for `sched_getaffinity` states that the function returns 0 on success and -1 on failure. This seems to be returning the size of a uint64_t which will always be 8. I think this is incorrect. 

What I think you have done is update the value being set in memory correctly on 434 (updating the size). But also updated the value returned to the program to be the size also on 433. Depending on the behaviour we want, 433 should be updated potentially in the way it was done previously i.e. set to 0 if `pid == 0` and -1 otherwise.

What was the reason for the update?

test/regression/aarch64/Syscall.cc
dANW34V3R: See below

src/lib/pipeline/ReorderBuffer.cc
JosephMoore25: Following offline discussion, we agree to keep this message in release mode, but add more detail s.t. the user is aware of why this is being triggered, what's triggering it, and what to do to resolve the issue. The Micro-op index in particular just adds additional verbosity so remains in the message.

src/lib/arch/aarch64/ExceptionHandler.cc
JosephMoore25: This is worth @jj16791 investigating as it was his find/fix so he will know more than I do on the issue.

The reason given at the time was:

> The assert you were triggering was KMP_ASSERT(__kmp_avail_proc == __kmp_topology->get_num_hw_threads());. Newer LLVM OMP runtimes require the affinity mask to be at least 8 bytes in length otherwise it will read the number of available cores out as 0 due to some casting. The affinity mask we were returning was 1 byte in length hence the assert triggered as __kmp_avail_proc was 0. Figured it out from a combination of isolating the instructions run leading up to this assert and then from GodBolt/SimEng figuring out why our mask was being converted to 0 procs available

I've been testing using a STREAM binary (with OpenMP support) compiled with ACFL23. With the current fix, this works. Removing the `sizeof` on 433 means that this fails. I do agree though that the current implementation doesn't line up with what I'd expect should work.

jj16791: Not sure what manpage you found but most should say something along the lines of "_but see "C library/kernel differences" below, which notes that the underlying sched_getaffinity() differs in its return value_". The difference is that is returns the number of bytes used to represent the mask. With a 64bit mask that's 8 bytes

jj16791: Always check for stuff like this re the non-wrapped version of the syscall does something different

src/lib/pipeline/ReorderBuffer.cc
```diff
@@ -81,9 +81,31 @@ unsigned int ReorderBuffer::commit(uint64_t maxCommitSize) {
...
+                  << "). This is unexpected behaviour for most valid core "
+                     "configurations, though may arise in designs with very "
+                     "high latencies or bottlenecks. If this is not the case, "
+                     "please try re-running. If this may be expected, you can "
+                     "increase this limit in "
+                     "`SimEng/src/include/pipeline/ReorderBuffer.hh` under the "
+                     "variable `robHeadRepeatLimit_`. Please raise "
+                     "an issue on GitHub if the problem persists."
+                  << std::endl;
+        exit(1);
```
jj16791, line 104: Anything to be said here for resetting the counter and letting the simulation continue? Just thinking about a possible scenario where this triggers after days of simulation but it wasn't actually a failing state.

src/include/simeng/arch/aarch64/helpers/neon.hh
```diff
@@ -558,6 +558,39 @@ RegisterValue vecLogicOp_3vecs(srcValContainer& sourceValues,
...
+ * uint8_t)
+ * I represents the number of elements in the output array to be updated (e.g.
+ * for vd.8b I = 8).
+ * Returns correctly formatted RegisterValue. */
+template <typename T, typename U, int I>
+RegisterValue vecAddlv(srcValContainer& sourceValues) {
+  const U* n = sourceValues[0].getAsVector<U>();
+  T out = 0;
+  for (int i = 0; i < I; i++) {
+    out += n[i];
```
jj16791, line 574: Should `n[i]` be explicitly cast here? Just aware we've had some issues with relying on implicit casting in prior years

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -849,6 +875,132 @@ RegisterValue sveFsqrtPredicated_2vecs(srcValContainer& sourceValues,
...
+template <typename T, typename U>
+RegisterValue sveFTrigSMul(srcValContainer& sourceValues,
+                           const uint16_t VL_bits) {
+  const T* n = sourceValues[0].getAsVector<T>();
+  const U* m = sourceValues[1].getAsVector<U>();
+
+  const uint16_t partition_num = VL_bits / (sizeof(T) * 8);
+  T out[256 / sizeof(T)] = {0};
+
+  U bit_0_mask = static_cast<U>(1) << (sizeof(T) * 8 - 1);
```
jj16791, line 892: We should be getting bit 0 for the sign not bit N-1

jj16791: Check if there are any other instances of this

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -936,6 +1088,63 @@ RegisterValue sveIndex(
...
+  for (int i = partition_num - 1; i >= 0; i--) {
+    uint64_t shifted_active = 1ull << ((i % (64 / sizeof(T))) * sizeof(T));
+    if (p[i / (64 / sizeof(T))] & shifted_active) {
+      lastElem = i;
+      break;
+    }
+    // If no active lane has been found, select highest element instead
+    if (i == 0) lastElem = partition_num - 1;
+  }
+  return {n[lastElem], 256};
```
jj16791, line 1113: Need to make sure we zero-extend here

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -936,6 +1088,63 @@ RegisterValue sveIndex(
...
+}
+
+/** Helper function for SVE instructions with the format `clastb zd, pg, zd,
```
jj16791, line 1116: I think the mnemonic is wrong here (reflects CLASTB (vectors)

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -936,6 +1088,63 @@ RegisterValue sveIndex(
...
+      break;
+    }
+  }
+
+  if (lastElem < 0) {
+    out = m[0];
+  } else {
+    out = n[lastElem];
+  }
+  return {out, 256};
```
jj16791, line 1145: Need to make sure we zero-extend here

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1425,6 +1692,51 @@ RegisterValue sveSminv(srcValContainer& sourceValues, const uint16_t VL_bits) {
   return {out, 256};
 }
 
+/** Helper function for SVE instructions with the format `splice zd, pg, zn,
+ * zm`.
+ * T represents the type of sourceValues (e.g. for zn.d, T = uint64_t).
+ * Returns correctly formatted RegisterValue. */
+template <typename T>
+RegisterValue sveSplice(srcValContainer& sourceValues, const uint16_t VL_bits) {
```
jj16791, line 1700: I assume this is the destructive variant of `splice`? If so, we should denote this somewhere

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -4359,6 +4588,14 @@ void Instruction::execute() {
                             sourceValues_[1].get<uint64_t>());
         break;
       }
+      case Opcode::AArch64_SPLICE_ZPZ_D: {  // splice zdn.d, pv, zdn.t, zm.d
```
jj16791, line 4591: Change `.t` in mnemonic to actual data type in use

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -4359,6 +4588,14 @@ void Instruction::execute() {
...
+        results_[0] = sveSplice<double>(sourceValues_, VL_bits);
+        break;
+      }
+      case Opcode::AArch64_SPLICE_ZPZ_S: {  // splice zdn.s, pv, zdn.t, zm.s
```
jj16791, line 4595: Change .t in mnemonic to actual data type in use

jj16791 (changes requested):

Issue #426, opened 2024-09-02
Compilation of SimEng fails on Manjaro Linux with GCC 14.2.1


jandrovins:
**System Description**
Please provide the following as a list:
 - OS: Manjaro Linux
 - Compiler: GCC 14.2.1
 - Processor: 13th Gen Intel(R) Core(TM) i7-1365U
 - Memory: 16GiB

**SimEng Version**
The tip of main branch, commit 7af3bc3fe7deb703756def31a208839bc5bdc4c2

**SimEng CMAKE Options Used**
- `-GNinja`
- `-DCMAKE_INSTALL_PREFIX=$PWD/install`
- `-DCMAKE_BUILD_TYPE=RelWithDebInfo`

**Problem Description**
Compilation fails while generating: src/lib/CMakeFiles/libsimeng.dir/arch/aarch64/ExceptionHandler.cc.o

```
(base) vincent@vanjaro at build >:/ ninja -j 1
[1/17] Building CXX object src/lib/CMakeFiles/libsimeng.dir/arch/aarch64/ExceptionHandler.cc.o
FAILED: src/lib/CMakeFiles/libsimeng.dir/arch/aarch64/ExceptionHandler.cc.o
/usr/bin/c++ -Dlibsimeng_EXPORTS -I/home/vincent/SimEng/src/include -I/home/vincent/SimEng/build/_deps/capstone-lib-build/include -I/home/vincent/SimEng/build/_deps/capstone-lib-src/include -I/home/vincent/SimEng/src/lib -I/home/vincent/SimEng/build/src/lib -O2 -g -DNDEBUG -std=c++17 -fPIC -fno-rtti -Wall -pedantic -Werror -MD -MT src/lib/CMakeFiles/libsimeng.dir/arch/aarch64/ExceptionHandler.cc.o -MF src/lib/CMakeFiles/libsimeng.dir/arch/aarch64/ExceptionHandler.cc.o.d -o src/lib/CMakeFiles/libsimeng.dir/arch/aarch64/ExceptionHandler.cc.o -c /home/vincent/SimEng/src/lib/arch/aarch64/ExceptionHandler.cc
In file included from /home/vincent/SimEng/src/include/simeng/RegisterFileSet.hh:6,
                 from /home/vincent/SimEng/src/include/simeng/ArchitecturalRegisterFileSet.hh:3,
                 from /home/vincent/SimEng/src/include/simeng/Core.hh:7,
[...]

PR #427, opened 2024-09-03
Added cstdint header to Register.hh

FinnWilkinson:
Adds the `cstdint` header to Register.hh, which has caused some compilers to not recognise the `uint8_t` etc types.
As per Issue #426

### DIFF
--- a/src/include/simeng/Register.hh
+++ b/src/include/simeng/Register.hh
@@ -1,5 +1,6 @@
 #pragma once
 #include <iostream>
+#include <cstdint>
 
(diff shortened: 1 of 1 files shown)


dANW34V3R (dismissed):
I think GitHub is complaining about the EOF but not an issue

JosephMoore25 (approved):

dANW34V3R (approved):

ABenC377 (approved):

jj16791 (approved):

PR #429, opened 2024-09-04
AArch64 Capstone Update / SME2 support

### LINKED ISSUE #349: Add SME2 support
Now that the Capstone Framework has been updated with the AArch64 auto-sync tool, SimEng should be updated to be compatable with this latest version of Capstone so that SME2 & more SVE2 instructions can be executed through SimEng.

FinnWilkinson:
This PR updates SimEng to work with a newer release of Capstone who's AArch64 engine is based on LLVM18, hence allowing SME2 support.

The reccommended version of LLVM has also been changed to 18.1.8 and the test suite has been updated to work with this version in order to dissassemble new AArch64 instructions.

Closes Issue #349 

There have been large changes to Capstone internally for AArch64, some of the main ones which effect our current development / usage include:

- A writeback from a pre/post index load/store is now included as an implicit destination. Hence, in instruction_execute the updated base address register is `resuts[0]` rather than `results[1]` (for example).
- Post index immediate operands are now included as part of the memory operand in `mem.disp`.  Post indexing with a register however is added as an additional operand.
- Aliasing can be dissabled for auto-sync architectures, and has been done for AArch64. This means expected registers for the operand are now correct and allows us to remove `revertalias()` from instruction_metadata completely! The only oddity is the `mnemonic` and `operandStr` are still that of the alias.  This means exceptions or debugging print statements may be confusing. To indicate aliases better, an `isAlias` bool has been added to the metadata and exception print out.

- [x] Get automatic LLVM download and install working for version 18.1.8
- [x] <del>Add AArch64 multi-vector operand enum decoding logic to instruction_decode</del>
   - No longer required as multi-vector operands already give each register individually rather than using the multi-register enums


CMakeLists.txt
```diff
@@ -149,14 +140,17 @@ if(SIMENG_ENABLE_TESTS)
     find_package(LLVM REQUIRED CONFIG NO_CMAKE_BUILDS_PATH)
 
       # Check LLVM version
-    if ((${LLVM_PACKAGE_VERSION} VERSION_LESS "8.0") OR (${LLVM_PACKAGE_VERSION} VERSION_GREATER_EQUAL "14.1"))
-      message(FATAL_ERROR "LLVM version must be >= 8.0 and <= 14.0")
+    if ((${LLVM_PACKAGE_VERSION} VERSION_LESS "8.0") OR (${LLVM_PACKAGE_VERSION} VERSION_GREATER_EQUAL "18.2"))
```
ABenC377, line 152: I think the if statement checks for >= 18.2, but the message says can be <= 18.2. Is 18.2 itself okay?

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -83,141 +84,160 @@ struct MicroOpInfo {
...
-  // ARM64_REG_V0 -> {end} are vector registers
-  if (op.reg >= ARM64_REG_V0) {
-    // Data size for vector registers relies on opcode, get vector access
-    // specifier
-    arm64_vas vas = op.vas;
-    assert(vas != ARM64_VAS_INVALID && "Invalid VAS type");
+inline uint8_t getDataSize(cs_aarch64_op op) {
+  // No V-register enum identifiers exist. Instead, depending on whether a full
+  // or half vector is accessed, a Q or D register is used instead.
+  // A `is_vreg` bool in `op` defines if we are using v-vecotr registers.
```
ABenC377: typo: 'vector'

src/lib/arch/aarch64/ExceptionHandler.cc
```diff
@@ -658,8 +658,17 @@ bool ExceptionHandler::init() {
     if (metadata.opcode == Opcode::AArch64_MSR) {
       newSVCR = instruction_.getSourceOperands()[0].get<uint64_t>();
     } else if (metadata.opcode == Opcode::AArch64_MSRpstatesvcrImm1) {
+      // Enusre operand metadata is as expected
```
ABenC377: typo -- 'Ensure'

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -208,186 +225,145 @@ void Instruction::decode() {
...
-          sourceRegisters_[sourceRegisterCount_] = csRegToRegister(op.reg);
-          sourceRegisterCount_++;
-          sourceOperandsPending_++;
-        }
+        // Add register reads to destinations
+        sourceRegisters_[sourceRegisterCount_] = csRegToRegister(op.reg);
+        sourceRegisterCount_++;
+        sourceOperandsPending_++;
+
         // TODO checking of the shift type is a temporary fix to help reduce the
```
ABenC377: Has this todo become actionable by the capstone update?

src/lib/config/ModelConfig.cc
```diff
@@ -370,8 +370,7 @@ void ModelConfig::setExpectations(bool isDefault) {
     expectations_["Core"].addChild(ExpectationNode::createExpectation<uint64_t>(
```
ABenC377: If we know that the maximum value for this is 2048, do we need to use unit64_t?  Elsewhere we have gone for the smallest suitable datatype.  Not sure if there is a technical advantage to using a smaller int type, but if there is then it might apply here as well.

ABenC377 (changes requested):

src/lib/config/ModelConfig.cc
FinnWilkinson: Good point / spot - changing to uint16_t

CMakeLists.txt
FinnWilkinson: Currently updating how we build LLVM in CMake so will see if this comment is still relevant then

CMakeLists.txt
```diff
@@ -14,21 +15,17 @@ macro(FetchContent_MakeAvailable_Args NAME ARGS)
...
+    cmake_policy(SET CMP0135 NEW)
+endif()
 
+# we don't use git for LLVM here as it clones the entire LLVM repo which takes too long and we only need a relatively small part of it
 FetchContent_Declare(
-  llvm
-  URL      https://github.com/llvm/llvm-project/releases/download/llvmorg-14.0.5/llvm-14.0.5.src.tar.xz
-  URL_HASH MD5=6bd202e403d950c78985048ce499a518
+        llvm
+        # URL https://github.com/llvm/llvm-project/releases/download/llvmorg-19.1.0/llvm-project-19.1.0.src.tar.xz
```
jj16791: Is there a reason why these commented-out lines are kept?

CMakeLists.txt
jj16791: Was there an outcome for this?

docs/sphinx/developer/arch/index.rst
```diff
@@ -5,7 +5,7 @@ SimEng architecture definitions are responsible for describing the features and
...
+ISA support is achieved through the use of the `Capstone <https://github.com/aquynh/capstone/>`_ disassembly framework, which disassembles a binary instruction into a C/C++ object that include operand registers, access types, and immediate values to name a few. In order to update SimEng's AArch64 support from Armv8.4-a to Armv9.2-a, we undertook a Capstone update to allow for disassembly of the Armv9.2-a ISA. The work done for this can be found `here <https://github.com/capstone-engine/capstone/pull/1907>`_. Extensive work continues to be done to Capstone by its community in order to allow for easier updating of ISA versions. This facilitated the recent update of SimEng's AArch64 ISA support to Armv9.4-a.
```
jj16791: Is it worth noting that we use a branch/fork of Capstone?

docs/sphinx/developer/arch/supported/aarch64.rst
```diff
@@ -135,7 +135,8 @@ There are several useful variables that execution behaviours have access to:
...
+  Some instructions have "implicit" destination registers: in these cases, the implicit destinations are added to the **start** of the results vector. For example, ``subs w0, w1, #1`` writes explicitly to ``w0``, but also implicitly sets the "NZCV" comparison flags. In this case, ``results[0]`` is expected to be the updated NZCV flags, while ``results[1]`` is expected to be the new value of ``w0``.
+  Some Load and Store instructions update the base memory address register (pre- or post-indexing). These registers are also classes as implicit destination registers and thus will be added to the start of the results vector. For example, ``ldr	x1, [x0], #8`` writes explicitly to ``x1`` but also writes implicitly to ``x0``. In this case, ``results[0]`` is expected to be the updated ``x0`` value (``x0 + 8``) and ``results[1]`` is expected to be the new value of ``x1``.
```
jj16791: *classed not classes

docs/sphinx/developer/arch/supported/aarch64.rst
```diff
@@ -223,9 +226,7 @@ Concerning SVE & SME loads and stores, an effort should be made to merge contigu
...
 
-As Capstone is primarily a disassembler, it will attempt to generate the correct aliases for instructions: for example, the ``cmp w0, #0`` instruction is an alias for ``subs wzr, w0, #0``. As it's the underlying instruction that is of use (in this case, the ``subs`` instruction), this implementation includes a de-aliasing component that reverses this conversion. The logic for this may be found in ``src/lib/arch/aarch64/InstructionMetadata``.
-
-If a known but unsupported alias is encountered, it will generate an invalid instruction error, and the output will identify the instruction as unknown in place of the usual textual representation. It is recommended to reference a disassembled version of the program to identify what the instruction at this address should be correctly disassembled to, and implement the necessary dealiasing logic accordingly.
+Although Capstone has been configured to produce the disassembly information for the "real" instruction rather than that of its (preferred) alias, the instruction's mnemonic and operand string will still be that of its alias. Hence, if an exception occurs the printed instruction informtion may not match the internal opcode used.
```
jj16791: *information

docs/sphinx/user/building_simeng.rst
```diff
@@ -41,7 +41,7 @@ With this configuration, the build files will be generated in a directory called
                 More information about the LLVM_DIR value can be found `here <https://llvm.org/docs/CMake.html#embedding-llvm-in-your-project>`_.
 
         .. Note::
-                LLVM versions greater than 14 or less than 8 are not supported. We'd recommend using LLVM 14.0.5 where possible as this has been verified by us to work correctly.
+                LLVM versions greater than 18 or less than 8 are not supported. We'd recommend using LLVM 18.1.8 where possible as this has been verified by us to work correctly for the most recent version of SimEng.
```
jj16791: Is this trying to say that LLVM versions 19+ and 8- are not supported? Think it's worth being more specific given the versioning convention used by LLVM (e.g. version greater than 18.2.0 rather than just 18)

docs/sphinx/user/configuring_simeng.rst
```diff
@@ -102,6 +102,9 @@ AArch64
 - Matrix-Count (Optional)
     The number of physical ``za`` Arm SME registers.
 
+- Table-Count (Optional)
```
jj16791: I think both Matrix and Table counts could benefit from `SME` somewhere in the config key. A bit too vague at the moment in my opinion

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -83,141 +84,160 @@ struct MicroOpInfo {
...
+      case AARCH64LAYOUT_VL_H:
         return 2;
-      }
-      case ARM64_VAS_1B: {
+      case AARCH64LAYOUT_VL_B:
         return 1;
-      }
-      default: {
+      default:
         assert(false && "Unknown VAS type");
```
jj16791: I'm aware this wasn't changed but I'd argue that this is a fatal error for SimEng so should be treated as such in either build type

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -83,141 +84,160 @@ struct MicroOpInfo {
   int microOpIndex = 0;
 };
 
+// TODO: Handle multi-register aarch64_reg operands
```
jj16791: Where do these show up? Do we need to consider them now before merging

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -83,141 +84,160 @@ struct MicroOpInfo {
...
+  // ARCH64_REG_PN0 -> +15 are scalable predicate registers
+  if (op.reg >= AARCH64_REG_PN0) {
+    /** TODO: Check functionality is correct when multi-vector operands + SME2
+     * has been supported. */
+    return 1;
+  }
+
+  // AARCH64_REG_P0 -> +15 are 256-bit (P) registers
+  if (op.reg >= AARCH64_REG_P0) {
     return 1;
```
jj16791: I'm aware this wasn't changed but why do we return a size of 1 for predicate registers?

src/include/simeng/arch/aarch64/Instruction.hh
jj16791: could be something like
```
std::cerr << "[SimEng] Cannot determine size of Arm V vector register with `reg` value " << op.reg << " and `vas` value of " << vas << ". Exiting..." << std::endl;
exit(1);
```
Not sure if the values are enum classes so may not be able to print as much info as I've suggested.

src/include/simeng/arch/aarch64/MicroDecoder.hh
```diff
@@ -84,17 +84,19 @@ class MicroDecoder {
 
   // Default objects
   /** Default capstone instruction structure. */
-  cs_arm64 default_info = {ARM64_CC_INVALID, false, false, 0, {}};
+  cs_aarch64 default_info = {AArch64CC_Invalid, false, false, 0, {}};
```
jj16791: don't we need another false in here for `is_doing_sme`?

jj16791: Also, not your fault but, I don't know how `{}` is being translated into `cs_aarch64_op operands[NUM_AARCH64_OPS];`...

src/include/simeng/arch/aarch64/helpers/auxiliaryFunctions.hh
```diff
@@ -125,6 +125,7 @@ bitfieldManipulate(T value, T dest, uint8_t rotateBy, uint8_t sourceBits,
 
 /** Function to check if NZCV conditions hold. */
 inline bool conditionHolds(uint8_t cond, uint8_t nzcv) {
+  // Due to Capstone enum changes, need to add 1 to cond
```
jj16791: Where is this addition done?

test/regression/aarch64/AArch64RegressionTest.hh
```diff
@@ -76,7 +76,7 @@ inline std::string paramToString(
 inline std::vector<std::tuple<CoreType, std::string>> genCoreTypeVLPairs(
     CoreType type) {
   std::vector<std::tuple<CoreType, std::string>> coreVLPairs;
-  for (uint64_t i = 128; i <= 2048; i += 128) {
+  for (uint64_t i = 128; i <= 2048; i *= 2) {
```
jj16791: I recall something about VLs of {128,256,512,1024,2048} only being supported but I don't think us removing support for VLs not in that set is good for the sake of research (unless there's logic which does not work if we include those VLS)

src/lib/config/ModelConfig.cc
```diff
@@ -367,16 +367,15 @@ void ModelConfig::setExpectations(bool isDefault) {
       std::vector{false, true});
 
   if (isa_ == ISA::AArch64) {
-    expectations_["Core"].addChild(ExpectationNode::createExpectation<uint64_t>(
+    expectations_["Core"].addChild(ExpectationNode::createExpectation<uint16_t>(
         128, "Vector-Length", true));
     expectations_["Core"]["Vector-Length"].setValueSet(
-        std::vector<uint64_t>{128, 256, 384, 512, 640, 768, 896, 1024, 1152,
-                              1280, 1408, 1536, 1664, 1792, 1920, 2048});
+        std::vector<uint16_t>{128, 256, 512, 1024, 2048});
```
jj16791: I recall something about VLs of {128,256,512,1024,2048} only being supported but I don't think us removing support for VLs not in that set is good for the sake of research (unless there's logic which does not work if we include those VLs)

src/lib/arch/aarch64/Architecture.cc
```diff
@@ -14,13 +14,14 @@ Architecture::Architecture(kernel::Linux& kernel, ryml::ConstNodeRef config)
...
                  (config["Core"]["Timer-Frequency-MHz"].as<uint32_t>() * 1e6)) {
-  if (cs_open(CS_ARCH_ARM64, CS_MODE_ARM, &capstoneHandle_) != CS_ERR_OK) {
+  if (cs_open(CS_ARCH_AARCH64, CS_MODE_ARM, &capstoneHandle_) != CS_ERR_OK) {
     std::cerr << "[SimEng:Architecture] Could not create capstone handle"
               << std::endl;
     exit(1);
   }
 
   cs_option(capstoneHandle_, CS_OPT_DETAIL, CS_OPT_ON);
+  cs_option(capstoneHandle_, CS_OPT_DETAIL, CS_OPT_DETAIL_REAL);
```
jj16791, line 26: What's this option?

src/lib/arch/aarch64/InstructionMetadata.cc
```diff
@@ -28,2577 +28,168 @@ InstructionMetadata::InstructionMetadata(const cs_insn& insn)
...
   std::memcpy(groups, insn.detail->groups, sizeof(uint8_t) * groupCount);
-  std::memcpy(operands, insn.detail->arm64.operands,
-              sizeof(cs_arm64_op) * operandCount);
+  std::memcpy(operands, insn.detail->aarch64.operands,
+              sizeof(cs_aarch64_op) * operandCount);
 
   // Fix some inaccuracies in the decoded metadata
   switch (opcode) {
-    case Opcode::AArch64_ADR_LSL_ZZZ_D_0:
-      [[fallthrough]];
```
jj16791, line 37: given we use -werror, we should keep these in to avoid generating fallthrough warnings

src/lib/arch/riscv/InstructionMetadata.cc
```diff
@@ -457,23 +457,6 @@ void InstructionMetadata::duplicateFirstOp() {
   operandCount = 3;
 }
 
-void InstructionMetadata::createMemOpPosOne() {
```
jj16791, line 460: I assume something changed in capstone which resolved this issue?

src/lib/arch/aarch64/InstructionMetadata.cc
```diff
@@ -13,13 +13,13 @@ InstructionMetadata::InstructionMetadata(const cs_insn& insn)
       implicitSourceCount(insn.detail->regs_read_count),
```
jj16791, line 13: How did you go about removing/adding/altering the entries in the switch statement? Just want to make sure we haven't missed anything regarding an unintended removal of a prior instruction fix that still needs fixing

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1044,17 +1046,16 @@ RegisterValue sveMax_vecImm(
   return {out, 256};
 }
 
-/** Helper function for SVE instructions with the format `max zdn, zdn,
- * #imm`.
+/** Helper function for SVE instructions with the format `max zdn, pg/m zdn,
```
jj16791: * pg/m, zdn

src/lib/arch/aarch64/MicroDecoder.cc
```diff
@@ -622,66 +627,91 @@ uint8_t MicroDecoder::decode(const Architecture& architecture, uint32_t word,
...
-      case arm64_op_type::ARM64_OP_CIMM:
-      case arm64_op_type::ARM64_OP_REG_MRS:
-      case arm64_op_type::ARM64_OP_REG_MSR:
-      case arm64_op_type::ARM64_OP_PSTATE:
-      case arm64_op_type::ARM64_OP_SYS:
-      case arm64_op_type::ARM64_OP_SVCR:
-      case arm64_op_type::ARM64_OP_PREFETCH:
-      case arm64_op_type::ARM64_OP_BARRIER:
-      case arm64_op_type::ARM64_OP_SME_INDEX:
+      case aarch64_op_type::AARCH64_OP_INVALID:
```
jj16791, line 655: I know they weren't there before but since the inclusion of -werror, we should have `[[fallthrough]]`'s

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -208,186 +225,140 @@ void Instruction::decode() {
...
-
-      if (metadata_.writeback) {
-        // Writeback instructions modify the base address
-        destinationRegisters_[destinationRegisterCount_] =
-            csRegToRegister(op.mem.base);
-        destinationRegisterCount_++;
+    } else if (op.type == AARCH64_OP_MEM) {  // Memory operand
+      // Check base register exists
+      if (op.mem.base != AARCH64_REG_INVALID) {
+        accessesMemory = true;
```
jj16791, line 329: Do we have instructions that have `AARCH64_OP_MEM` ops with no base register?

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -208,186 +225,140 @@ void Instruction::decode() {
...
-            destinationRegisters_[destinationRegisterCount_] = regs[i];
-            destinationRegisterCount_++;
-          }
-        }
-      } else {
-        // SME_INDEX can also be for predicate
-        // Set instruction group
-        setInstructionType(InsnType::isPredicate);
+
+    } else if (op.type == AARCH64_OP_SME) {
```
jj16791, line 347: I'm sure it's correct but just to double check, an SME operand cannot exist in the op.reg field?

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -208,186 +225,140 @@ void Instruction::decode() {
...
         sourceOperandsPending_++;
       }
-    } else if (op.type == ARM64_OP_REG_MSR) {
-      int32_t sysRegTag = architecture_.getSystemRegisterTag(op.imm);
-      if (sysRegTag == -1) {
-        exceptionEncountered_ = true;
-        exception_ = InstructionException::UnmappedSysReg;
-        return;
-      } else {
+      if (op.sysop.sub_type == AARCH64_OP_REG_MSR) {
```
jj16791: Shouldn't this be `else if`?

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -660,36 +831,42 @@ void Instruction::decode() {
     setInstructionType(InsnType::isPredicate);
   }
   // Uncaught float data assignment for FMOV move to general instructions
-  if (((430 <= metadata_.opcode && metadata_.opcode <= 432) ||
-       (2409 <= metadata_.opcode && metadata_.opcode <= 2429)) &&
+  if (((Opcode::AArch64_FMOVD0 <= metadata_.opcode &&
+        metadata_.opcode <= Opcode::AArch64_FMOVS0) ||
+       (Opcode::AArch64_FMOVDXHighr <= metadata_.opcode &&
+        metadata_.opcode <= Opcode::AArch64_FMOVv8f16_ns)) &&
```
jj16791: Aren't the `_ns` instructions moving immediates into vector registers? (and thus aren't "move to general instructions")

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -660,36 +831,42 @@ void Instruction::decode() {
...
 
-  if (!(isInstruction(InsnType::isSMEData))) {
-    // Catch zero register references and pre-complete those operands - not
-    // applicable to SME instructions
-    for (uint16_t i = 0; i < sourceRegisterCount_; i++) {
-      if (sourceRegisters_[i] == RegisterType::ZERO_REGISTER) {
-        sourceValues_[i] = RegisterValue(0, 8);
-        sourceOperandsPending_--;
-      }
+  // if (!(isInstruction(InsnType::isSMEData))) {
```
jj16791: Do we not need this check anymore?

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -660,36 +831,42 @@ void Instruction::decode() {
...
+    if (sourceRegisters_[i] == RegisterType::ZERO_REGISTER) {
+      sourceValues_[i] = RegisterValue(0, 8);
+      sourceOperandsPending_--;
     }
-  } else {
-    // For SME instructions, resize the following structures to have the exact
-    // amount of space required
+  }
+  // } else {
+  if (isInstruction(InsnType::isSMEData)) {
```
jj16791: To remind myself, for SME, do we expand these containers to their maximum and then shrink them back down to the actual usage?

jj16791 (changes requested):
A fair few comments but there's a lot of work done here so expected. Great job on this, a significant amount of effort done!

CMakeLists.txt
FinnWilkinson: These checks are still in place so I will update them inline with Alex's initial comment

src/include/simeng/arch/aarch64/Instruction.hh
FinnWilkinson: They are present in the register enum in Capstone / LLVM, however Capstone does not use them but has each register as an individual operand.

TLDR - this todo needs to be removed

src/include/simeng/arch/aarch64/helpers/auxiliaryFunctions.hh
FinnWilkinson: I thought I removed this comment.... was some confusions when first starting the update

test/regression/aarch64/AArch64RegressionTest.hh
FinnWilkinson: My argument for removing them is the Arm spec doesn't acknowledge them anymore (as of the latests spec at least) and we have already seen using "old" VLs cause issued.

 @JosephMoore25 recently had a VL-agnostic STREAM kernel (I think?) not working correctly as the compiler (rightfully) assumed that the VL was a power of two. Hence, VL=384 for example got stuck in an infinite loop.

If we are using non VL-agnostic code or hand written code, then the "old" VLs should be ok. But If modern versions of compilers are assuming only powers of 2 (i.e. {128, 256, 512, 1024, 2048}) then we may run into more issues similar to the above

src/lib/config/ModelConfig.cc
FinnWilkinson: See https://github.com/UoB-HPC/SimEng/pull/429#discussion_r1818819483

src/lib/arch/aarch64/Architecture.cc
FinnWilkinson: This option reverts the aliasing so that operand details etc are correct and eliminates the need for `revertAlias()` in `instructionMetadata.cc`. I.e., it means the Capstone operands (registers, mem, etc) match the "verbose" opcode (our addition in our fork) exactly rather than the ID (the more generic opcode Capstone uses).

Currently, the mnemonic and operand string in the metadata will still be that of the alias, but in the exceptionHandler printout there is a message stating this and I have raised this with the Capstone developers who are working to fix this.

src/lib/arch/aarch64/InstructionMetadata.cc
FinnWilkinson: I'm not sure I understand fully... Most opcodes don't need an entry in this switch case anymore so should go to the default case. Are you suggesting keeping all the old opcodes as fallthrough into the default case?

src/lib/arch/riscv/InstructionMetadata.cc
FinnWilkinson: Yes, some RISC-V updates were also included which made this function un-used

src/lib/arch/aarch64/InstructionMetadata.cc
FinnWilkinson: Based on the assumption that all of these instruction have associated tests (a fair one given they're predominantly SVE / SME) I commented out all entries and assessed which ones were still required to make the regression test suite function.

As far as I can tell, all our usual benchmarks still work as expected - but the CI/CD pipeline can verify this once operational

src/include/simeng/arch/aarch64/helpers/sve.hh
FinnWilkinson: Not sure I understand what you're pointing out

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: This is a safeguard against one of the un-used operands (as the Capstone array is fixed at 16 entries) having the type of `AARCH64_OP_MEM` but then no register. I think it came up once with one instruction hence the addition

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: Not anymore, no - changes were made to the aarch64 Capstone detail such that any operand accessing ZA (tile or otherwise) will be of type `AARCH64_OP_SME`. Whereas before, only instructions with an operand of (for example) `za[w12, #0]` would have had a type of `AARCH64_OP_SME`.

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: Probably yes, good spot

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: For this I replaced the numbers with Opcodes (and referred to the previous Capstone version to make sure they aligned). If this is the case then we can remove the `_ns` ones from this catch

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: No - I think it is now possible for some SME instructions to include zero registers (will double check this though)

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: Correct yes

docs/sphinx/user/building_simeng.rst
dANW34V3R: Do lower versions of LLVM not support SME? If so that would be worth mentioning here (or at least somewhere in the docs)

docs/sphinx/user/configuring_simeng.rst
dANW34V3R: +1

src/include/simeng/arch/aarch64/helpers/sve.hh
dANW34V3R: Add a comma?

src/lib/arch/aarch64/Architecture.cc
dANW34V3R: Could you add a short comment e.g. "Prevent of aliasing"

src/lib/arch/riscv/InstructionMetadata.cc
dANW34V3R: This will be because the aliasing has been "turned off", so don't need to "correct" the operands

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -36,133 +33,147 @@ constexpr int32_t signExtend(uint32_t value, int currentLength) {
...
+ * register representation.
  *
  * WARNING: this conversion is FRAGILE, and relies on the structure of the
- * `arm64_reg` enum. Updates to the Capstone library version may cause this to
- * break. */
-Register csRegToRegister(arm64_reg reg) {
-  // Check from top of the range downwards
+ * `aarch64_reg` enum. Updates to the Capstone library version may cause this to
+ * break.
+ * TODO: Add multi-register enum decoding.
```
dANW34V3R: Can this TODO be removed?

dANW34V3R (commented):
Not much to add on top of the others' comments. Will approve once theirs have been resolved

CMakeLists.txt
FinnWilkinson: Updated message

CMakeLists.txt
FinnWilkinson: Nope, removed

src/include/simeng/arch/aarch64/Instruction.hh
FinnWilkinson: Updated to 32 as I don't know either

src/include/simeng/arch/aarch64/MicroDecoder.hh
FinnWilkinson: Added missing false. Not sure either on `{}` front....

src/lib/arch/aarch64/Architecture.cc
FinnWilkinson: Comment added to provide some explanation

src/lib/arch/riscv/InstructionMetadata.cc
FinnWilkinson: I'm not 100% sure, but I don't think the alias reversing effected RISC-V as that Capstone arch hasn't been updated extensively in a while. Either way, this function is no longer required

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: Upon checking, SME based loads with a scalar register memory offset default to XZR

FinnWilkinson: This could be semi-optimised however for SME instructions (who could have loads of source operands to check through) by having a second check for ZERO registers in the `else if (op.type == AARCH64_OP_MEM) {  // Memory operand` check; as this is the only place I can think of where a ZERO reg would be for an SME instruction

docs/sphinx/user/building_simeng.rst
FinnWilkinson: Versions from ~14 do support SME. Can add something to this

test/regression/aarch64/AArch64RegressionTest.hh
jj16791: I'd argue that using a VL for a code that doesn't support it is a user problem, not a SimEng problem

src/lib/arch/aarch64/Instruction_decode.cc
jj16791: Double-checked and we should remove `_ns` from this catch

test/regression/aarch64/AArch64RegressionTest.hh
FinnWilkinson: After offline discussion, reverted to use all VLs which are multiples of 128 between the values of 128 and 2048 inclusive

src/lib/config/ModelConfig.cc
FinnWilkinson: After offline discussion, reverted to use all VLs which are multiples of 128 between the values of 128 and 2048 inclusive

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: Updated catch

dANW34V3R (dismissed):

ABenC377 (dismissed):

dANW34V3R (dismissed):

ABenC377 (dismissed):

ABenC377 (dismissed):

JosephMoore25 (dismissed):
Great work, and clean implementation. Couldn't find anything of note!

jj16791 (dismissed):

ABenC377 (dismissed):

JosephMoore25 (dismissed):

CMakeLists.txt
```diff
@@ -149,46 +149,89 @@ if(SIMENG_ENABLE_TESTS)
...
+                -DLLVM_BUILD_TOOLS=OFF
+                -DLLVM_BUILD_EXAMPLES=OFF
+                -DLLVM_ENABLE_BINDINGS=OFF
+                -DLLVM_ENABLE_WARNINGS=OFF
+                "-DLLVM_TARGETS_TO_BUILD=AArch64\\;RISCV"
+
+                ${COMMAND_ECHO_OPTION}
+                RESULT_VARIABLE SUCCESS)
+
+        # XXX replace with COMMAND_ERROR_IS_FATAL in the future (>= 3.19)
```
dANW34V3R: what is "XXX" referring to

dANW34V3R: @tom91136 should this just say "TODO"

tom91136: Oh, XXX means _attention_,  see https://stackoverflow.com/a/1452945/896997

dANW34V3R: I think we should change to TODO for consistency

tom91136: @FinnWilkinson Do you want to make the change? It's much easier from your side.

FinnWilkinson: Yup - I'll change this

dANW34V3R (approved):

JosephMoore25 (approved):

jj16791 (approved):

PR #434, opened 2024-10-06
Fixes #409

### LINKED ISSUE #409: Version.hh
The version.hh file is currently generated in the `src` directory. This should be moved to the build folders. @tom91136 has volunteered to fix this

tom91136:
Fix a bug where the generated header `version.hh` gets placed in the source, causing full rebuilds from time to time.


ABenC377 (approved):

dANW34V3R (approved):
This seems to do the trick

jj16791 (commented):
I'm sure there's good reasoning but just so I know. WHy did we change the bin dir include from private to public? Was it just to allow for the visibility of the version.hh file?

tom91136:
Yes, otherwise the header in not visible in dependent projects and won't compile.

jj16791 (approved):

PR #430, opened 2024-09-18
Apple Clang 16 fix

FinnWilkinson:
With Apple Clang 16, the static local variable `SimInfoClass` inside the `getInstance()` function had two instantiations - one in `main.cc` and one in `libsimeng.dylib`. This is due to an optimisation made in the Link Time Optimisation in CMake.

This PR moves the function definitions into a `.cc` file to mitigate this error.

PR #437, opened 2024-10-17
CI/CD

### LINKED ISSUE #36: Add CI coverage for macOS

### LINKED ISSUE #341: Overhaul the Jenkin's test pipeline
Additional to the current regression testing and performance testing that the Jenkin's pipeline does, it needs to be updated to a) test more benchmakrs, b) ensure said benchmarks are functionally correct, and c) ensure that the pass/fail criteria of performance tests are robust and consistent.

We have also seen unexpected errors and behaviour when running SimEng on different Linux distributions / Operating systems. Hence, the testing that a new PR undergoes should include running tests on multiple operating systems - perhaps via docker instances

ABenC377:
Upstreaming the build, test and benchmark components of Leo's CI/CD pipeline


.github/workflows/MACOS_BUILD_TEST.yml
```diff
@@ -0,0 +1,104 @@
...
+
+      #######################################
+      # Run Regression RISCV Tests.
+      #######################################
+      - name: regression test (riscv)
+        if: always()
+        shell: bash
+        run: |
+          ./build/test/regression/riscv/regression-riscv
+
```
dANW34V3R, line 93: Is there a way we can do "cmake --build build --target test" rather than running each binary individually? This would reduce future maintenance

.github/workflows/LINUX_BUILD_TEST.yml
```diff
@@ -0,0 +1,165 @@
...
+
+jobs:
+  Build_and_Run:
+    runs-on: ${{ inputs.RUNNER }}
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        
+        COMPILER: ['gcc-7', 'gcc-8', 'gcc-9', 'gcc-10'] # armclang ] compiler names
```
dANW34V3R: Can we get higher versions of GCC as well as LLVM clang?

.github/workflows/LINUX_BUILD_TEST.yml
```diff
@@ -0,0 +1,165 @@
...
+
+      #######################################
+      # Run Regression RISCV Tests.
+      #######################################
+      - name: Regression Test (riscv)
+        if: always()
+        shell: bash
+        run: |
+          ./build/test/regression/riscv/regression-riscv
+
```
dANW34V3R, line 155: Again, can this be one command, rather than running each binary separately

.github/actions/simeng_benchmarks/action.yml
```diff
@@ -0,0 +1,49 @@
...
+  PAT:
+    description: personal access token required to checkout private benchmark repo
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+
+    ##################################
+    # clones repo with different versions of the checkout action, as container has different node versions and some compilers run on older OSs that have differening node versions
```
dANW34V3R: Comment formatting needs another row of # I think

.github/actions/setup_gcc_ubuntu/action.yml
```diff
@@ -0,0 +1,107 @@
...
+    # Running this prevents gcc-10 on ubuntu 20 from segfaulting in unit tests for some reason
+    # (installs some packages, not sure which ones prevent failure, but it doesn't take long to run).
+    #######################################
+    - if: ${{ inputs.gcc-version == 'gcc-10' }}
+      name: install llvm
+      shell: bash
+      run: |
+        wget https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
+        ./llvm.sh 14
```
dANW34V3R: This sounds like something we should fix (or at least note in the docs), rather than patch in the CI

dANW34V3R: What actually is the issue here?

.github/actions/setup_gcc_ubuntu/action.yml
```diff
@@ -0,0 +1,107 @@
...
+    - name: Install gcc
+      shell: bash
+      run: |
+        apt-get -y install ${{ inputs.gcc-version }}
+        apt-get -y install g++-$( echo ${{ inputs.gcc-version }} | cut -d '-' -f 2)
+        apt update && apt upgrade -y
+        
+    #######################################
+    # Build SimEng without llvm or ninja
+    #######################################
```
dANW34V3R: The below command is building with LLVM (bc tests are on) but not the local install. So the comment is a little confusing. 

Do we still need the llvm install above for other tests?

.github/actions/setup_gcc_rocky/action.yml
```diff
@@ -0,0 +1,127 @@
...
+      name: save gcc
+      uses: actions/cache/save@v4
+      id: gcc-save-v4
+      with: 
+        path: /usr/local/${{ inputs.gcc-version }}.5.0
+        key: ${{ inputs.gcc-version }}-${{ inputs.OS }}
+
+    #######################################
+    # Build SimEng without llvm or ninja
+    #######################################
```
dANW34V3R: Can probably change the comment to just be "Build SimEng"

.github/actions/setup_gcc_macos/action.yml
```diff
@@ -0,0 +1,55 @@
...
+      run: |
+        cmake -B build -S . -DCMAKE_BUILD_TYPE=${{ inputs.MODE }} -DSIMENG_ENABLE_TESTS=ON -DSIMENG_OPTIMIZE=ON -DCMAKE_C_C_COMPILER=${{ env.C_COMPILER }} \
+          -DCMAKE_CXX_C_COMPILER=${{ env.CPP_COMPILER }}
+
+        if [[ ${{ inputs.MODE }} == "Release" ]]; then
+          cmake --build build -j $(sysctl -n hw.ncpu)
+        else
+          echo "cores: $(sysctl -n hw.ncpu)"
+          cmake --build build -j 5
+        fi
```
dANW34V3R: Is this complication necessary? And why the magic number of 5

.github/actions/setup_gcc_rocky/action.yml
dANW34V3R: Whatever you choose, please keep consistent throughout

.github/actions/setup_armclang_ubuntu/action.yml
```diff
@@ -0,0 +1,101 @@
...
+
+        ./arm-compiler-for-linux_24.04_Ubuntu-20.04/arm-compiler-for-linux_24.04_Ubuntu-20.04.sh --install-to ./armclang_compiler
+        export MODULEPATH=$MODULEPATH:$(pwd)/armclang_compiler/modulefiles
+
+        module avail
+        module load acfl/24.04
+        armclang -v
+        echo "C_COMPILER=$(which armclang)" >> $GITHUB_ENV
+        echo "CPP_COMPILER=$(which armclang++)" >> $GITHUB_ENV
+
```
dANW34V3R, line 91: Add comment

.github/actions/average_single_benchmark/action.yml
```diff
@@ -0,0 +1,68 @@
...
+
+          # Loop to run the benchmark 'runs' times
+          for (( i=1; i<=runs; i++ ))
+          do  
+              # Run the benchmark and redirect output to a file
+              if [ $datafile_path ]; then
+                  sudo simeng "${{ github.workspace }}/configs/a64fx.yaml" $benchmark_name -n 64 -i 1 --deck "${{ github.workspace }}/$datafile_path"  > "$output_file"
+              else
+                  sudo simeng "${{ github.workspace }}/configs/a64fx.yaml" $benchmark_name  > "$output_file"
+              fi
```
dANW34V3R: I think this could be generalised more. Instead of checking whether a "data file path" has been provided and hard coding the rest of the miniBUDE input, the "function" should have a `workload_parameters` input. This can be empty or a string of many parameters. 

This "function" would then simply call `sudo simeng a64fx $workload_parameters` and would be completely generalised to the workload. This would be more easily understandable in my opinion.

Not sure the best way to achieve this currently

dANW34V3R (commented):
Looks good on the whole and soo much better than what we had before. Just some minor comments

.github/workflows/MACOS_BUILD_TEST.yml
ABenC377: Doing it this way means each of the tests shows up as a separate step on the github action so the immediate interface of the job is more descriptive.  Therefore, I think that this is the preferable way to do it.

.github/workflows/LINUX_BUILD_TEST.yml
ABenC377: Potentially.  But I think that is a further development job.  Currently I'm just doing the minimum to get Leo's pipeline upstreamed as I don't have capacity at the moment to do further development.  Happy for you to have a shot at it though...

.github/workflows/LINUX_BUILD_TEST.yml
ABenC377: see above comment

.github/actions/simeng_benchmarks/action.yml
ABenC377: updated

.github/actions/average_single_benchmark/action.yml
dANW34V3R: Having found the correct branch on the simeng-benchmarks repo, you could somehow pipe the `cmd` field from the json file directly here

.github/actions/setup_gcc_ubuntu/action.yml
ABenC377: Yeah, I think the second clause in this comment is there for historic reasons -- previously in development this was building in a different way.  Therefore, replacing with "build simeng", as you suggest below.

.github/actions/setup_gcc_rocky/action.yml
ABenC377: Yep, agreed

.github/workflows/MACOS_BUILD_TEST.yml
dANW34V3R: I see your point. There is a trade off here, but as we probably aren't adding new suites very often I agree we should leave as is

.github/workflows/LINUX_BUILD_TEST.yml
dANW34V3R: Will see if I have time. Might spend an hour on it but no more

.github/workflows/LINUX_BUILD_TEST.yml
dANW34V3R: Resolving as above

.github/actions/setup_gcc_ubuntu/action.yml
```diff
@@ -0,0 +1,108 @@
...
+    - name: Install gcc
+      shell: bash
+      run: |
+        apt-get -y install ${{ inputs.gcc-version }}
+        apt-get -y install g++-$( echo ${{ inputs.gcc-version }} | cut -d '-' -f 2)
+        apt update && apt upgrade -y
+        
+    #######################################
+    # Build SimEng
+    # Then store the C compiler and CPP compiler as environmental variables to be used by LINUX_BUILD_TEST
```
dANW34V3R: Why is it only necessary to store these env vars for gcc ubuntu and not the other distros?

.github/actions/simeng_benchmarks/action.yml
dANW34V3R: This is a VERY long comment on 1 line. Is there a way we can run clang format ON the .yml files?

.github/actions/setup_gcc_ubuntu/action.yml
ABenC377: It's only needed once.  So I think the choice of distro is arbitrary and happened to fall on Ubuntu

.github/actions/setup_gcc_macos/action.yml
ABenC377: Not necessary.  I've removed it.  We're using the variable number of jobs for all now

.github/actions/simeng_benchmarks/action.yml
ABenC377: I don't think we have clang format rules for yaml files.  Nonetheless, happy to separate out into different lines, if that is our stylistic preference

FinnWilkinson:
Fixes issue #36 
Fixes issue #341

.github/actions/simeng_benchmarks/action.yml
dANW34V3R: I think yes, split it over a couple of lines. Don't worry about clang format.

I don't think there were any other cases that were too bad

.github/actions/setup_gcc_macos/action.yml
dANW34V3R: I think it is still as a comment unless you haven't pushed those changes. Will resolve once removed

.github/actions/setup_gcc_ubuntu/action.yml
ABenC377: No longer an issue.  This workaround has been removed.

.github/actions/average_single_benchmark/action.yml
ABenC377: This is actually superfluous for the non-performance workflows.  Have removed from the PR.

.github/workflows/LINUX_BUILD_TEST.yml
FinnWilkinson: Why are are not testing with armclang?

src/include/simeng/Register.hh
FinnWilkinson, line 1: Is this not already in dev from PR #427

src/include/simeng/config/SimInfo.hh
FinnWilkinson, line 1: I think this is already in `dev` too from PR #430 

(Applies to `src/lib/CMakeLists.txt` and `src/lib/config/SimInfo.cc` also)

FinnWilkinson (changes requested):
Very minor things

.github/actions/setup_gcc_macos/action.yml
ABenC377: I think now pushed

src/include/simeng/Register.hh
ABenC377: It is -- not sure why github is showing it as a changed file.  Potentially a mistake on my end in terms of rebasing/merging

src/include/simeng/config/SimInfo.hh
ABenC377: It is -- not sure why github is showing it as a changed file.  Potentially a mistake on my end in terms of rebasing/merging

.github/workflows/LINUX_BUILD_TEST.yml
ABenC377: Screenshot 2024-10-28 at 15 24 16

Armclang isn't working yet -- added a todo to emphasise that this is something we want to be adding

FinnWilkinson (dismissed):
Congrats on getting this to work!!

src/include/simeng/Register.hh
FinnWilkinson: Given the changes are identical, and to prevent a rebase making this PR take longer, I think we can merge as is. Will only effect git blame anyway

src/include/simeng/config/SimInfo.hh
FinnWilkinson: Given the changes are identical, and to prevent a rebase making this PR take longer, I think we can merge as is. Will only effect git blame anyway

.github/actions/simeng_benchmarks/action.yml
dANW34V3R: Has the comment been removed?

.github/actions/simeng_benchmarks/action.yml
```diff
@@ -2,48 +2,45 @@ name: simeng-benchmarks
...
 
 inputs:
-  BENCHMARK_BRANCH:
-    description: name of the branch on the benchmark repo you want to checkout 
+  RUN_DIR:
+    description: directory from which the benchmark binary should be run
+    required: false
+  DATA_PATH:
+    description: location of the data file(s) for this benchmarks
+    required: false
```
dANW34V3R: As per https://github.com/UoB-HPC/SimEng/pull/437#discussion_r1810502735  I think this should be more generic. Rather than passing just the data path, pass all parameters to be passed to the binary under simulation

.github/actions/simeng_benchmarks/action.yml
```diff
@@ -2,48 +2,45 @@ name: simeng-benchmarks
...
+        if [ ${{ inputs.DATA_PATH }} ]
+        then
+          simeng "$GITHUB_WORKSPACE/configs/a64fx.yaml" "${{ inputs.BIN_PATH }}" -n 64 -i 1 --deck "${{ inputs.DATA_PATH }}"  > $GITHUB_WORKSPACE/simeng.tmp
+        elif [ ${{ inputs.RUN_DIR }} ]
+        then
+          cd ${{ inputs.RUN_DIR }}
+          simeng "$GITHUB_WORKSPACE/configs/a64fx.yaml" "${{ inputs.BIN_PATH }}" -n 64 -i 1  > $GITHUB_WORKSPACE/simeng.tmp
+        else
+          simeng "$GITHUB_WORKSPACE/configs/a64fx.yaml" "${{ inputs.BIN_PATH }}" -n 64 -i 1  > $GITHUB_WORKSPACE/simeng.tmp
+        fi
```
dANW34V3R: I think in all cases, all benchmarks are being executed with hardcoded miniBUDE parameters. AFAIK, for our specific workloads atm this shouldn't cause issues, but it isn't future proof. As per above, if the whole string "-n 64 -i 1 /path/to/bm1" were passed into this action then this part would simplify a lot and should prevent issues. Tell me if there is anything that actually prevents you from doing that

.github/workflows/LINUX_BUILD_TEST.yml
```diff
@@ -153,13 +154,583 @@ jobs:
...
+        with:
+          DATA_PATH: $GITHUB_WORKSPACE/simeng-benchmarks/Data_Files/miniBUDE/bm1
+          BIN_PATH: $GITHUB_WORKSPACE/simeng-benchmarks/binaries/miniBUDE/openmp/minibude_gcc10.3.0_armv8.4
+          PASS_STRING: "Largest difference was 0.000%."
+
+      - if: ${{ inputs.SIMENG-MODE == 'Release' }}
+        name: miniBUDE openmp armclang20 armv8.4
+        uses: ./.github/actions/simeng_benchmarks
+        with:
+          DATA_PATH: $GITHUB_WORKSPACE/simeng-benchmarks/Data_Files/miniBUDE/bm1
```
dANW34V3R: Following on from previous comments, here you could have something along the lines of:  `PARAM_STRING: "-n n64 -i 1 --deck " + $GITHUB_WORKSPACE/simeng-benchmarks/Data_Files/miniBUDE/bm1` . Not sure of the syntax, but hopefully this illustrates the point

dANW34V3R (changes requested):
Inclusion of a pass string is great. I think the benchmarks are still being run slightly incorrectly

FinnWilkinson (dismissed):

dANW34V3R (dismissed):
This is soo much better. Nice work. And worth saying thanks to @haelai77 for doing the bulk of the work

.github/workflows/LINUX_BUILD_TEST.yml
```diff
@@ -0,0 +1,736 @@
...
+
+jobs:
+  Build_and_Run:
+    runs-on: ${{ inputs.RUNNER }}
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        
+        COMPILER: ['gcc-7', 'gcc-8', 'gcc-9', 'gcc-10'] # todo 'armclang'] # compiler names
```
JosephMoore25, line 29: Is armclang still not supported? What about apple clang?

.github/workflows/LINUX_BUILD_TEST.yml
```diff
@@ -0,0 +1,736 @@
...
+        OS: ['ubuntu:18.04','ubuntu:20.04', 'rockylinux:8', 'redhat/ubi8:latest', 'redhat/ubi9:latest', 'debian:10', 'debian:11'] # Docker images
+        
+        #######################################
+        # Removes unecessary jobs as jobs are generated in the order seen in the matrix.
+        # "Exclude" is to keep job ordering nice i.e. keeping ubuntu jobs next to each other in the list.
```
JosephMoore25, line 34: Currently 12 are excluded, 14 are included. It _would_ be a bit cleaner to specify inclusions, and that'd mean the `select_setup` action could be cleaned to a single case where it takes in `OS` and `Compiler` and always works. That may even work atm, depending on where the following actually get excluded.

Tradeoff between time to rework vs how much it actually matters though, as current implementation seems clean enough

.github/workflows/LINUX_BUILD_TEST.yml
```diff
@@ -0,0 +1,736 @@
...
+jobs:
+  Build_and_Run:
+    runs-on: ${{ inputs.RUNNER }}
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        
+        COMPILER: ['gcc-7', 'gcc-8', 'gcc-9', 'gcc-10'] # todo 'armclang'] # compiler names
+        OS: ['ubuntu:18.04','ubuntu:20.04', 'rockylinux:8', 'redhat/ubi8:latest', 'redhat/ubi9:latest', 'debian:10', 'debian:11'] # Docker images
```
JosephMoore25, line 30: Depending on current runtime, we can probably get away with just redhat 9 and rocky 8 (or I'd personally prefer to see rocky 9 and redhat 8). RHEL and Rocky are very similar OS's (designed to be 1:1 compatible), so a bit redundant to have two of the "same", but worth testing both 8 and 9 as well as RHEL and Rocky in _some_ combination

.github/workflows/LINUX_BUILD_TEST.yml
```diff
@@ -0,0 +1,736 @@
...
+
+      - if: ${{ inputs.SIMENG-MODE == 'Release' }}
+        name: CloverLeaf serial gcc8.3.0 armv8.4
+        uses: ./.github/actions/simeng_benchmarks
+        with:
+          RUN_DIR: $GITHUB_WORKSPACE/simeng-benchmarks/Data_Files/CloverLeaf
+          BIN_PATH: $GITHUB_WORKSPACE/simeng-benchmarks/binaries/CloverLeaf/serial/cloverleaf_gcc8.3.0_armv8.4
+          PASS_STRING: "This test is considered PASSED"
+
+      - if: ${{ inputs.SIMENG-MODE == 'Release' }}
```
JosephMoore25, line 188: Can these instead be one command per benchmark, generic for compiler? If you name the binary correctly (with compiler in), then you could just use the variable of the compiler in the bin path? That way, adding a new benchmark takes one block for all OS/compilers which is much nicer imo as this is **all** repeated. Same in `MACOS_BUILD_TEST`.

JosephMoore25: Seeing one of the resolved comments, I understand now why it's all separated. Can you please add a comment here and in `MACOS_BUILD_TEST` as to why they're separated.

JosephMoore25 (changes requested):
Looks great so far, really nice to have this sorted so elegantly. A couple comments on the clarity and conciseness of some of the code, but these may be agreed to not be necessary. Will be glad to see this go through soon!

.github/workflows/LINUX_BUILD_TEST.yml
ABenC377: See Finn's previous comment.  Doesn't work yet.  Don't have time to implement before this PR

.github/workflows/LINUX_BUILD_TEST.yml
ABenC377: Yeah, could change but doesn't really make much of a difference.  So not going to change here.

.github/workflows/LINUX_BUILD_TEST.yml
ABenC377: These are run in parallel so wouldn't change run time.

JosephMoore25: Can be changed at a later date if we run out of runners

JosephMoore25 (dismissed):

FinnWilkinson (dismissed):

dANW34V3R (dismissed):

.github/actions/select_setup/action.yml
```diff
@@ -0,0 +1,113 @@
+name: setup armclang
+description: installs dependencies and correct armclang version to build and test simeng
```
jj16791: Should the name and description be generic here? looks like this script calls the specific ones

.github/actions/setup_armclang_debian/action.yml
```diff
@@ -0,0 +1,95 @@
+name: setup gcc
```
jj16791: Incorrect name/desc?

.github/actions/setup_armclang_redhat/action.yml
```diff
@@ -0,0 +1,91 @@
+name: setup gcc
```
jj16791: Incorrect name/desc?

.github/actions/setup_armclang_rocky/action.yml
```diff
@@ -0,0 +1,86 @@
+name: setup gcc
```
jj16791: Incorrect name/desc?

jj16791 (commented):
Really nice Job @ABenC377, a strong basis which can hopefully be moved over to a self hosted runner in the future.

FinnWilkinson (dismissed):

dANW34V3R (dismissed):

.github/actions/setup_armclang_debian/action.yml
```diff
@@ -0,0 +1,95 @@
+name: setup armclang
+description: installs dependencies and correct gcc version to build and test simeng
```
jj16791: Isn't this description still wrong?

.github/actions/setup_armclang_redhat/action.yml
```diff
@@ -0,0 +1,91 @@
+name: setup armclang
+description: installs dependencies and correct gcc version to build and test simeng
```
jj16791: Isn't this description still wrong?

.github/actions/setup_armclang_rocky/action.yml
```diff
@@ -0,0 +1,86 @@
+name: setup armclang
+description: installs dependencies and correct gcc version to build and test simeng
```
jj16791: Isn't this description still wrong?

FinnWilkinson (approved):

JosephMoore25 (approved):

dANW34V3R (approved):

PR #439, opened 2024-11-04
[AArch64] NEON, SVE2 and SME2 instruction support with tests

FinnWilkinson:
This PR adds a wide range of different NEON, SVE2, SME2 instructions with regressions tests. These facilitate a subset of some internal SME-based GEMM and GEMV codes.

There is some BF16 prototypical instruction support which by default is disabled (using a new build option and an if statement in each appropriate switch statement case) due to some usage of `__bf16` which is not compiler agnostic, some hacky usage of memcpy to re-interpret `uint16_t`, and a lack of regression tests for the BF16 instructions in question. 

These BF16 instructions can be enabled through a new CMake option `-DSIMENG_ENABLE_BF16=ON`. I have deliberately **_not_** included this in the documentation given the possible instibility of the BF16 implementation and to keep it for (mainly) internal usage only.

This branch is based on `sme2-support` (PR #429 ) and so should be merged after this brnch has been merged into `dev`.

Some SM2 instructions which use multi-vector operands can be non-trivial to read or understand. Please ask for clarification and suggest any additional comments that may help future understanding.


CMakeLists.txt
```diff
@@ -155,10 +156,10 @@ if(SIMENG_ENABLE_TESTS)
 
       # Print message containing if the full test suite will run
     if (${LLVM_PACKAGE_VERSION} VERSION_LESS "14.0")
-      message(STATUS "LLVM version does not support AArch64 extensions SME or SVE2. These test suites will be skipped.")
+      message(STATUS "LLVM version does not support AArch64 extensions SVE2, SVE2.1, SME, or SME2. Related tests will fail.")
```
ABenC377, line 159: Not entirely sure on the behaviour of message().  Would LLVM versions less than 14 result in both messages being printed?  If so, could we use an if/else logic to avoid duplicated information?

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -283,6 +283,42 @@ enum class InsnType : uint32_t {
...
+  const bool invert = (predAsCounter & 0b1000000000000000) != 0;
+  const uint64_t predElemCount =
+      (predAsCounter & static_cast<uint64_t>(0b0111111111111111)) >>
+      predCountShiftVals[sizeof(T)];
+
+  for (int r = 0; r < V; r++) {
+    for (int i = 0; i < elemsPerVec; i++) {
+      // Move bit to next position based on element type
+      uint64_t shifted_active = 1ull << ((i % (64 / sizeof(T))) * sizeof(T));
+      // If invert = 1, predElemCount dictates number of initial inactive
```
ABenC377: invert is of type bool.  Even though it is the same thing, I think it would be more easily understandable if this comment and the below comment were updated to true/false rather than 1/0

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -626,6 +626,27 @@ std::enable_if_t<std::is_floating_point_v<T>, RegisterValue> sveFDivPredicated(
   return {out, 256};
 }
 
+/** Helpfer function for SVE instructions with the format `faddv rd, pg, zn.
```
ABenC377: type -- Help(f)er

ABenC377 (commented):
Only a few comments

CMakeLists.txt
FinnWilkinson: Yup, you're correct - have updated to ifelse

src/include/simeng/arch/aarch64/Instruction.hh
FinnWilkinson: I will make the comment a mash up of 1/True and 0/False, as the variable type is bool but original comment was relating to the invert-bit in the predicate itself. Agree with your point though

src/include/simeng/arch/aarch64/helpers/sve.hh
ABenC377: God, the irony of me misspelling typo here.

ABenC377: God, the irony of me misspelling typo here.

src/include/simeng/arch/aarch64/helpers/neon.hh
```diff
@@ -585,9 +590,14 @@ RegisterValue vecUMinP(srcValContainer& sourceValues) {
   const T* n = sourceValues[0].getAsVector<T>();
   const T* m = sourceValues[1].getAsVector<T>();
 
+  // Concatenate the vectors
+  T temp[2 * I];
+  memcpy(temp, m, sizeof(T) * I);
```
jj16791: Shouldn't `m` and `n` be switched here?

src/include/simeng/arch/aarch64/operandContainer.hh
```diff
@@ -10,7 +10,7 @@ namespace arch {
 namespace aarch64 {
 
 /** The maximum number of source registers a non-SME instruction can have. */
-const uint8_t MAX_SOURCE_REGISTERS = 6;
+const uint8_t MAX_SOURCE_REGISTERS = 7;
```
jj16791, line 13: Need to be careful this doesn't overwrite the changes in #429

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -548,7 +549,7 @@ void Instruction::decode() {
       } else if (metadata_.operands[0].is_vreg) {
         setInstructionType(InsnType::isVectorData);
       } else if ((metadata_.operands[0].reg >= AARCH64_REG_ZAB0 &&
-                  metadata_.operands[0].reg <= AARCH64_REG_ZT0) ||
+                  metadata_.operands[0].reg < AARCH64_REG_ZT0) ||
```
jj16791: Can `ZT0` be used in a SVE context?

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -1884,6 +2252,62 @@ void Instruction::execute() {
         results_[0] = vecFmlsIndexed_3vecs<float, 4>(sourceValues_, metadata_);
         break;
       }
+      case Opcode::AArch64_BFMOPA_MPPZZ: {  // bfmopa zada.s, pn/m, pm/m, zn.h,
+                                            // zm.h
+        // SME
+        // BF16 -- EXPERIMENTAL
+        if (std::string(SIMENG_ENABLE_BF16) == "OFF") return executionNYI();
```
jj16791: I think this would be better implemented through a preprocessor directive for the entire case

test/regression/aarch64/AArch64RegressionTest.hh
```diff
@@ -190,6 +190,24 @@ inline std::vector<std::tuple<CoreType, std::string>> genCoreTypeSVLPairs(
     checkMatrixRegisterCol<type>(tag, index, __VA_ARGS__); \
   }
 
+/** Check each element of the Lookup Table register ZT0 against expected values.
+ *
+ * The `tag` argument is the register index (must be 0), and the `type` argument
```
jj16791: If `tag` must always be 0, then why not hardcode it as such?

test/regression/aarch64/AArch64RegressionTest.hh
```diff
@@ -239,13 +257,16 @@ class AArch64RegressionTest : public RegressionTest {
...
   std::string getSubtargetFeaturesString() {
-#if SIMENG_LLVM_VERSION < 14
-    return "+sve,+lse";
-#elif SIMENG_LLVM_VERSION < 18
-    return "+sve,+lse,+sve2,+sme,+sme-f64";
-#else
-    return "+sve,+lse,+sve2,+sme,+sme-f64f64,+sme-i16i64,+sme2";
+    std::string features = "+dotprod,+sve,+lse";
+#if SIMENG_LLVM_VERSION > 13
+    // "+dotprod,+sve,+lse,+sve2,+sme,+sme-f64";
```
jj16791, line 261: Do we need this comment?

test/regression/aarch64/instructions/float.cc
```diff
@@ -1453,6 +1453,37 @@ TEST_P(InstFloat, ucvtf) {
   CHECK_NEON(9, float, {static_cast<float>(UINT64_C(1) << 48), 0.f, 0.f, 0.f});
   CHECK_NEON(10, float, {static_cast<float>(UINT64_MAX), 0.f, 0.f, 0.f});
   CHECK_NEON(11, float, {0.f, 0.f, 0.f, 0.f});
+
+  // 32-bit unsigned fixed-point to float
+  // Numbers have been chosen to have less than 0.0005 fixed-point
+  // representation error to ensure tests pass
+  initialHeapData_.resize(12);
+  heap32 = reinterpret_cast<uint32_t*>(initialHeapData_.data());
+  heap32[0] = 0x000001EE;
```
jj16791: Could you comment on what these values represent in float form, e.g. `heap32[0] = 0x000001EE; // 0.0....`

test/regression/aarch64/instructions/load.cc
```diff
@@ -282,6 +317,53 @@ TEST_P(InstLoad, ld1_multi_struct) {
...
+  RUN_AARCH64(R"(
+    # Get heap address
+    mov x0, 0
+    mov x8, 214
+    svc #0
+
+    mov x1, #32
+
+    # Load values from heap
+    # ld1 {v0.8h, v1.8h}, [x0]
```
jj16791, line 330: I assume this is a placeholder for when this is implemented.

test/regression/aarch64/Exception.cc
```diff
@@ -151,7 +151,6 @@ TEST_P(Exception, unmapped_sys_reg) {
   EXPECT_EQ(stdout_.substr(0, strlen(err)), err);
 }
 
-#if SIMENG_LLVM_VERSION >= 14
```
jj16791: Is this in response to the SVE vs SVE2 identification issue or something else? I feel like we can keep this sort of checking in

src/lib/arch/aarch64/Instruction_execute.cc
```diff
@@ -486,6 +520,66 @@ void Instruction::execute() {
         branchAddress_ = instructionAddress_ + metadata_.operands[0].imm;
         break;
       }
+      case Opcode::AArch64_BF16DOTlanev8bf16: {  // bfdot vd.4s, vn.8h,
```
jj16791, line 666: Do the bf16 instructions have test cases?

jj16791, line 666: Do the bf16 instructions have test cases?

src/include/simeng/arch/aarch64/Instruction.hh
```diff
@@ -283,6 +283,43 @@ enum class InsnType : uint32_t {
   isBranch = 1 << 14
 };
 
+/** Predefined shift values for converting pred-as-counter to pred-as-mask. */
+const uint64_t predCountShiftVals[9] = {0, 1, 2, 0, 3, 0, 0, 0, 4};
```
jj16791: Unless I've missed something, this is used in one location. Why is the data defined as a variable outside of all function scopes?

CMakeLists.txt
```diff
@@ -155,10 +156,9 @@ if(SIMENG_ENABLE_TESTS)
 
       # Print message containing if the full test suite will run
     if (${LLVM_PACKAGE_VERSION} VERSION_LESS "14.0")
-      message(STATUS "LLVM version does not support AArch64 extensions SME or SVE2. These test suites will be skipped.")
-    endif()
-    if (${LLVM_PACKAGE_VERSION} VERSION_LESS "18.0")
-      message(STATUS "LLVM version does not support AArch64 extensions SME2. These test suites will be skipped.")
+      message(STATUS "LLVM version does not support AArch64 extensions SVE2, SVE2.1, SME, or SME2. Related tests will fail.")
```
jj16791, line 159: Why can't we place preprocessor directives around the SME tests? I though it was just a SVE vs SVE2 problem?

jj16791 (changes requested):

src/include/simeng/arch/aarch64/helpers/neon.hh
FinnWilkinson: This will be fixed in PR #415 which will be merged before any SME2 PR

src/include/simeng/arch/aarch64/operandContainer.hh
FinnWilkinson: This PR is based on #429 and so cannot be merged before it. Will keep an eye out for this though

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: `ZT0` is enabled / disabled in the same way as `Z0` but has a fixed width of 512-bits, with the logic for detecting whether a `ZT0` related instruction can / can't be executed done in `instruction_execute` as with all other SME instructions.

Regarding where in a core/implementation `ZT0` based instructions are executed, there is no fixed rule in the spec as far as I can tell.... Given its fixed width, to me it seems more SVE-like than SME hence the grouping seen here. And given we don't have co-processor SME support, theres no offload / seperate chip logic to come into play yet

test/regression/aarch64/AArch64RegressionTest.hh
FinnWilkinson: No, these comments are only here for clarity to show all extensions that are enabled for each LLVM version clause

test/regression/aarch64/Exception.cc
FinnWilkinson: Yes - in response to the SVE / SVE2 and SME / SME2 checks (i.e. it is not trivial). This one can stay though, yes

src/lib/arch/aarch64/Instruction_execute.cc
FinnWilkinson: No, I've kept them as experimental implementations and undocumented, including how to enable them. You would need to look through the sourcecode and CMake files to know it is there

FinnWilkinson: No, I've kept them as experimental implementations and undocumented, including how to enable them. You would need to look through the sourcecode and CMake files to know it is there

src/lib/arch/aarch64/Instruction_execute.cc
FinnWilkinson: Updated

src/include/simeng/arch/aarch64/Instruction.hh
FinnWilkinson: Removed as can calculate it automatically

CMakeLists.txt
FinnWilkinson: There is a similar problem with SME and SME2

ABenC377 (approved):

ABenC377 (dismissed):

ABenC377 (dismissed):

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1319,6 +1340,40 @@ std::array<uint64_t, 4> svePtrue(
...
 
+/** Helper function for SVE instructions with the format `ptrue pnd.
+ * T represents the type of sourceValues (e.g. for pnd.d, T = uint64_t).
+ * Returns an array of 4 uint64_t elements. */
+template <typename T>
+std::array<uint64_t, 4> svePtrue_counter(const uint16_t VL_bits) {
+  // Predicate as counter is 16-bits and has the following encoding:
+  //    - Up to first 4 bits encode the element size (0b1, 0b10, 0b100, 0b1000
+  //    for b h s d respectively)
+  //            - bits 0->LSZ
```
jj16791, line 1351: LSZ?

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1319,6 +1340,40 @@ std::array<uint64_t, 4> svePtrue(
...
+  //    - Bits LSZ -> 14 represent a uint of the number of consecutive elements
+  //    from element 0 that are active / inactive
+  //            - If invert bit = 0 it is number of active elements
+  //            - If invert bit = 1 it is number of inactive elements
+  //    - Bit 15 represents the invert bit
+  std::array<uint64_t, 4> out = {0, 0, 0, 0};
+
+  // Set invert bit to 1 and count to 0
```
jj16791: add some context for this choice, i.e. relate it back to ptrue. I assume it's because you want to denote true as 0 inactive elements?

src/include/simeng/arch/aarch64/helpers/sve.hh
```diff
@@ -1563,6 +1618,69 @@ RegisterValue sveTrn2_3vecs(srcValContainer& sourceValues,
...
 
+/** Helper function for SVE instructions with the format `udot zd, zn, zm`.
+ * D represents the element type of the destination register (i.e. for zd.s,
+ * D = uint32_t).
+ * N represents the element type of the source registers (i.e. for zn.b, N =
+ * uint8_t).
+ * W represents how many source elements are multiplied to form an output
+ * element (i.e. for 4-way, W = 4).
+ * Returns correctly formatted RegisterValue. */
+template <typename D, typename N, int W>
```
jj16791, line 1629: isn't N always uint8_t and W 4? May have missed something in the docs

src/lib/arch/aarch64/Instruction_decode.cc
jj16791: I am not sure I follow the answer given here. I was just wondering if the ZT0 is used in SVE instructions, as it seems to be solely used in SME instructions when looking at the spec. If this is the case, would we not want to identify it as an SME instruction? May have missed something though.

test/regression/aarch64/instructions/sme.cc
```diff
@@ -7,8 +7,52 @@ namespace {
 
 using InstSme = AArch64RegressionTest;
 
-#if SIMENG_LLVM_VERSION >= 14
```
jj16791, line 10: Are we not able to keep this check in? I assumed this just concerned no SME and SME support as opposed to SME and SME2 support. May have misremembered the LLVM versioning though

src/include/simeng/arch/aarch64/helpers/sve.hh
FinnWilkinson: This is terminology from the spec on how predicate-as-counter works. It is `LSZ` as the number of bits used is dynamic. I'll try find a webpage with this info to better explain it in the comment

src/include/simeng/arch/aarch64/helpers/sve.hh
FinnWilkinson: Again, this is how the spec defined it -- will try and find a source to better explain

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: Having gone through the spec more and how ZT0 is used, you're right - it should be SME not SVE ---- Overthinked this a bit previously I think...

test/regression/aarch64/instructions/sme.cc
FinnWilkinson: Yeah you're correct. Will re-add

src/include/simeng/arch/aarch64/helpers/sve.hh
FinnWilkinson: Will double check and alter if so

FinnWilkinson: SVE UDOT has 4 possible encodings. See:
 - https://developer.arm.com/documentation/ddi0602/2024-12/SVE-Instructions/UDOT--4-way--vectors---Unsigned-integer-dot-product-
 - https://developer.arm.com/documentation/ddi0602/2024-12/SVE-Instructions/UDOT--2-way--vectors---Unsigned-integer-dot-product-

ABenC377 (dismissed):

src/lib/arch/aarch64/Instruction_decode.cc
```diff
@@ -532,8 +532,8 @@ void Instruction::decode() {
 
     if (isInstruction(InsnType::isStoreData)) {
       // Identify store instruction group
-      if (AARCH64_REG_Z0 <= metadata_.operands[0].reg &&
-          metadata_.operands[0].reg <= AARCH64_REG_Z31) {
+      if ((AARCH64_REG_Z0 <= metadata_.operands[0].reg &&
```
jj16791: Remove excess parentheses

src/include/simeng/arch/aarch64/helpers/sve.hh
jj16791: I'm not sure where this question came from. May have read some old docs

test/regression/aarch64/instructions/sme.cc
```diff
@@ -1354,4 +1356,8 @@ INSTANTIATE_TEST_SUITE_P(AArch64, InstSme,
                          ::testing::ValuesIn(genCoreTypeSVLPairs(EMULATION)),
                          paramToString);
 
+#else
+GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(InstSme);
```
jj16791, line 4217: I assume this is something new we had to do in light of the GTEST version update?

jj16791 (changes requested):

test/regression/aarch64/instructions/sme.cc
FinnWilkinson: Yeah - if the SME test file is empty this clause is needed so that things don't break

src/lib/arch/aarch64/Instruction_decode.cc
FinnWilkinson: Removed

ABenC377 (approved):

jj16791 (approved):

JosephMoore25 (approved):
All looks good to me. Clear implementations and comments. Can't see anything wrong without digging into each instruction individually, but given tests have been written for each instruction and (I'm assuming) all pass, I'm happy for this to be merged.

Nice work!

PR #442, opened 2024-12-05
Making GCC7 work on ubuntu 20

ABenC377:
Removing ubuntu18 support from testing pipeline due to Github actions no longer supporting the required version of GLIBC

### DIFF
--- a/.github/actions/setup_gcc_ubuntu/action.yml
+++ b/.github/actions/setup_gcc_ubuntu/action.yml
@@ -70,6 +70,9 @@ runs:
     - name: Install gcc
       shell: bash
       run: |
+        if [[ ${{ inputs.gcc-version == 'gcc-7' }} ]]; then
+          echo "deb http://archive.ubuntu.com/ubuntu focal main universe" >> "/etc/apt/sources.list"
+        fi
         apt-get -y install ${{ inputs.gcc-version }}
         apt-get -y install g++-$( echo ${{ inputs.gcc-version }} | cut -d '-' -f 2)
         apt update && apt upgrade -y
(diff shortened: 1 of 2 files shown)


dANW34V3R (dismissed):

FinnWilkinson (dismissed):

JosephMoore25 (dismissed):

.github/actions/setup_gcc_ubuntu/action.yml
```diff
@@ -76,6 +70,9 @@ runs:
     - name: Install gcc
       shell: bash
       run: |
+        if [[ ${{ inputs.gcc-version == 'gcc-7' }} ]]; then
+          echo "deb http://archive.ubuntu.com/ubuntu focal main universe" >> "/etc/apt/sources.list"
```
dANW34V3R: Using `apt-add-repository "deb http://archive.ubuntu.com/ubuntu focal main universe"` should do exactly the same thing and would be consistent with above, but not that important

dANW34V3R (dismissed):

FinnWilkinson (dismissed):

.github/actions/setup_gcc_ubuntu/action.yml
ABenC377: Good shout.  Also, combining it with the above (just chucking universe onto the end of the original command) works.  So that's what we're going with

JosephMoore25 (approved):
Looks good, glad you found a fix! Perhaps rename the PR to more accurately represent the change for the sake of git history?

FinnWilkinson (approved):

jj16791 (approved):

dANW34V3R (approved):

PR #443, opened 2024-12-09
Tage

ABenC377:
Adding a TAGE branch predictor. 

Performance relative to previous best (Perceptron) summarised below:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/no22498/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/no22498/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">

</head>

<body link="#0563C1" vlink="#954F72">

Benchmark | BP_update time | BP_update mispredict | TAGE time | TAGE mispredict | Performance change (percentage) | Mispredict change (raw)
-- | -- | -- | -- | -- | -- | --
CloverLeaf serial gcc8.3.0   armv8.4 | 11715 | 10.70% | 10511 | 2.06% | -10.28% ✅ | -8.64% ✅
CloverLeaf serial gcc9.3.0   armv8.4 | 11830 | 10.40% | 10321 | 1.85% | -12.76% ✅ | -8.55% ✅
CloverLeaf serial gcc10.3.0   armv8.4 | 12457 | 12.70% | 10550 | 1.74% | -15.31% ✅ | -10.96% ✅
CloverLeaf serial armclang20   armv8.4 | 11041 | 14.50% | 9069 | 1.85% | -17.86% ✅ | -12.65% ✅
CloverLeaf openmp gcc8.3.0   armv8.4 | 15635 | 9.78% | 14530 | 2.00% | -7.07% ✅ | -7.78% ✅
CloverLeaf openmp gcc9.3.0   armv8.4 | 15559 | 7.82% | 14680 | 1.64% | -5.65% ✅ | -6.18% ✅
CloverLeaf openmp gcc10.3.0   armv8.4 | 15291 | 9.33% | 13872 | 1.56% | -9.28% ✅ | -7.77% ✅
CloverLeaf openmp armclang20   armv8.4 | 13939 | 13.30% | 12273 | 1.74% | -11.95% ✅ | -11.56% ✅
miniBUDE openmp gcc8.3.0   armv8.4 | 21292 | 7.37% | 20613 | 5.71% | -3.19% ✅ | -1.66% ✅
miniBUDE openmp gcc9.3.0   armv8.4 | 20652 | 7.43% | 20364 | 5.71% | -1.39% ✅ | -1.72% ✅
miniBUDE openmp gcc10.3.0   armv8.4 | 20810 | 7.41% | 20691 | 5.71% | -0.57% ✅ | -1.70% ✅
miniBUDE openmp armclang20   armv8.4 | 19365 | 10.30% | 19808 | 5.05% | +2.29% ❌ | -5.25% ✅
STREAM serial gcc8.3.0 armv8.4 | 6619 | 0.51% | 6672 | 0.42% | +0.80% ❌ | -0.09% ✅
STREAM serial gcc9.3.0 armv8.4 | 6612 | 0.50% | 6778 | 0.42% | +2.51% ❌ | -0.08% ✅
STREAM serial gcc10.3.0   armv8.4 | 6583 | 0.44% | 6649 | 0.42% | +1.00% ❌ | -0.03% ✅
STREAM serial armclang20   armv8.4 | 7018 | 1.01% | 7071 | 0.75% | +0.76% ❌ | -0.26% ✅
STREAM openmp gcc8.3.0 armv8.4 | 10200 | 1.90% | 10066 | 0.67% | -1.31% ✅ | -1.23% ✅
STREAM openmp gcc9.3.0 armv8.4 | 10037 | 2.08% | 9859 | 0.50% | -1.77% ✅ | -1.58% ✅
STREAM openmp gcc10.3.0   armv8.4 | 9814 | 1.81% | 9799 | 0.63% | -0.15% ✅ | -1.18% ✅
STREAM openmp armclang20   armv8.4 | 10292 | 3.04% | 10459 | 1.21% | +1.62% ❌ | -1.83% ✅
TeaLeaf 2D serial gcc8.3.0   armv8.4 | 11874 | 15.20% | 10871 | 1.04% | -8.45% ✅ | -14.16% ✅
TeaLeaf 2D serial gcc9.3.0   armv8.4 | 11846 | 15.20% | 10847 | 1.05% | -8.43% ✅ | -14.15% ✅
TeaLeaf 2D serial gcc10.3.0   armv8.4 | 12020 | 15.20% | 11078 | 1.05% | -7.84% ✅ | -14.15% ✅
TeaLeaf 2D serial armclang20   armv8.4 | 21353 | 9.16% | 20597 | 3.27% | -3.54% ✅ | -5.89% ✅
TeaLeaf 2D openmp gcc8.3.0   armv8.4 | 17221 | 7.05% | 16118 | 1.18% | -6.40% ✅ | -5.87% ✅
TeaLeaf 2D openmp gcc9.3.0   armv8.4 | 17224 | 7.75% | 16485 | 1.43% | -4.29% ✅ | -6.32% ✅
TeaLeaf 2D openmp gcc10.3.0   armv8.4 | 16595 | 6.70% | 16221 | 0.85% | -2.25% ✅ | -5.85% ✅
TeaLeaf 2D openmp armclang20   armv8.4 | 52356 | 9.29% | 50683 | 2.38% | -3.20% ✅ | -6.91% ✅
TeaLeaf 3D serial gcc8.3.0   armv8.4 | 13645 | 8.73% | 13203 | 1.15% | -3.24% ✅ | -7.58% ✅
TeaLeaf 3D serial gcc9.3.0   armv8.4 | 14157 | 10.70% | 13463 | 1.57% | -4.90% ✅ | -9.13% ✅
TeaLeaf 3D serial gcc10.3.0   armv8.4 | 14331 | 11.00% | 13167 | 1.50% | -8.12% ✅ | -9.50% ✅
TeaLeaf 3D serial armclang20   armv8.4 | 19199 | 22.40% | 16675 | 1.69% | -13.15% ✅ | -20.71% ✅
TeaLeaf 3D openmp gcc8.3.0   armv8.4 | 22251 | 7.62% | 21775 | 2.30% | -2.14% ✅ | -5.32% ✅
TeaLeaf 3D openmp gcc9.3.0   armv8.4 | 22774 | 9.03% | 21750 | 1.47% | -4.50% ✅ | -7.56% ✅
TeaLeaf 3D openmp gcc10.3.0   armv8.4 | 22229 | 8.33% | 20867 | 0.99% | -6.13% ✅ | -7.34% ✅
TeaLeaf 3D openmp armclang20   armv8.4 | 40910 | 16.90% | 37148 | 1.30% | -9.20% ✅ | -15.60% ✅
CloverLeaf serial gcc8.3.0   armv8.4+sve | 11137 | 10.90% | 9820 | 2.31% | -11.83% ✅ | -8.59% ✅
CloverLeaf serial gcc9.3.0   armv8.4+sve | 11051 | 9.98% | 9909 | 1.86% | -10.33% ✅ | -8.12% ✅
CloverLeaf serial gcc10.3.0   armv8.4+sve | 11140 | 12.80% | 9462 | 1.80% | -15.06% ✅ | -11.00% ✅
CloverLeaf serial armclang20   armv8.4+sve | 11051 | 13.30% | 9280 | 1.50% | -16.03% ✅ | -11.80% ✅
CloverLeaf openmp gcc8.3.0   armv8.4+sve | 14845 | 9.65% | 13076 | 1.87% | -11.92% ✅ | -7.78% ✅
CloverLeaf openmp gcc9.3.0   armv8.4+sve | 15310 | 7.96% | 13208 | 1.80% | -13.73% ✅ | -6.16% ✅
CloverLeaf openmp gcc10.3.0   armv8.4+sve | 14754 | 9.62% | 13397 | 1.56% | -9.20% ✅ | -8.06% ✅
CloverLeaf openmp armclang20   armv8.4+sve | 14309 | 12.20% | 12493 | 1.29% | -12.69% ✅ | -10.91% ✅
miniBUDE openmp gcc8.3.0   armv8.4+sve | 8707 | 14.20% | 7917 | 3.61% | -9.07% ✅ | -10.59% ✅
miniBUDE openmp gcc9.3.0   armv8.4+sve | 8440 | 7.43% | 7656 | 3.54% | -9.29% ✅ | -3.89% ✅
miniBUDE openmp gcc10.3.0   armv8.4+sve | 8458 | 7.41% | 7695 | 3.41% | -9.02% ✅ | -4.00% ✅
miniBUDE openmp armclang20   armv8.4+sve | 8655 | 20.30% | 7961 | 1.21% | -8.02% ✅ | -19.09% ✅
STREAM serial gcc8.3.0   armv8.4+sve | 3521 | 1.51% | 3572 | 1.28% | +1.45% ❌ | -0.23% ✅
STREAM serial gcc9.3.0   armv8.4+sve | 3534 | 1.50% | 3980 | 1.28% | +12.62% ❌ | -0.22% ✅
STREAM serial gcc10.3.0   armv8.4+sve | 3481 | 1.44% | 3787 | 1.28% | +8.79% ❌ | -0.16% ✅
STREAM serial armclang20   armv8.4+sve | 2207 | 2.01% | 2268 | 1.46% | +2.76% ❌ | -0.55% ✅
STREAM openmp gcc8.3.0   armv8.4+sve | 6821 | 3.90% | 6740 | 1.79% | -1.19% ✅ | -2.11% ✅
STREAM openmp gcc9.3.0   armv8.4+sve | 7198 | 3.08% | 6686 | 1.01% | -7.11% ✅ | -2.07% ✅
STREAM openmp gcc10.3.0   armv8.4+sve | 7067 | 2.81% | 6646 | 1.18% | -5.96% ✅ | -1.63% ✅
STREAM openmp armclang20   armv8.4+sve | 6428 | 3.04% | 5867 | 1.44% | -8.73% ✅ | -1.60% ✅
TeaLeaf 2D serial gcc8.3.0   armv8.4+sve | 11725 | 15.20% | 10942 | 1.04% | -6.68% ✅ | -14.16% ✅
TeaLeaf 2D serial gcc9.3.0   armv8.4+sve | 11584 | 15.20% | 10899 | 1.04% | -5.91% ✅ | -14.16% ✅
TeaLeaf 2D serial gcc10.3.0   armv8.4+sve | 11879 | 15.20% | 11034 | 1.05% | -7.11% ✅ | -14.15% ✅
TeaLeaf 2D serial armclang20   armv8.4+sve | 11137 | 9.16% | 7370 | 0.87% | -33.82% ✅ | -8.29% ✅
TeaLeaf 2D openmp gcc8.3.0   armv8.4+sve | 17041 | 7.05% | 16349 | 1.18% | -4.06% ✅ | -5.87% ✅
TeaLeaf 2D openmp gcc9.3.0   armv8.4+sve | 17451 | 7.75% | 16532 | 1.43% | -5.27% ✅ | -6.32% ✅
TeaLeaf 2D openmp gcc10.3.0   armv8.4+sve | 16463 | 6.70% | 16187 | 0.85% | -1.68% ✅ | -5.85% ✅
TeaLeaf 2D openmp armclang20   armv8.4+sve | 52701 | 9.29% | 49203 | 1.63% | -6.64% ✅ | -7.66% ✅
TeaLeaf 3D serial gcc8.3.0   armv8.4+sve | 12169 | 18.73% | 10629 | 1.60% | -12.66% ✅ | -17.13% ✅
TeaLeaf 3D serial gcc9.3.0   armv8.4+sve | 12183 | 18.70% | 10639 | 1.07% | -12.67% ✅ | -17.63% ✅
TeaLeaf 3D serial gcc10.3.0   armv8.4+sve | 12405 | 18.30% | 10616 | 1.58% | -14.42% ✅ | -16.72% ✅
TeaLeaf 3D serial armclang20   armv8.4+sve | 19363 | 22.40% | 15654 | 1.42% | -19.16% ✅ | -20.98% ✅
TeaLeaf 3D openmp gcc8.3.0   armv8.4+sve | 21676 | 7.62% | 18948 | 2.22% | -12.59% ✅ | -5.40% ✅
TeaLeaf 3D openmp gcc9.3.0   armv8.4+sve | 20728 | 9.03% | 18989 | 1.51% | -8.39% ✅ | -7.52% ✅
TeaLeaf 3D openmp gcc10.3.0   armv8.4+sve | 20438 | 8.33% | 18652 | 0.85% | -8.74% ✅ | -7.48% ✅
TeaLeaf 3D openmp armclang20   armv8.4+sve | 41040 | 8.49% | 37791 | 0.88% | -7.92% ✅ | -7.61% ✅

</body>

</html>


docs/sphinx/user/configuring_simeng.rst
```diff
@@ -164,7 +164,16 @@ RAS-entries
...
+    Only needed for ``Generic`` and ``Tage`` predictors.  The static predictor used when no dynamic prediction is available. The options are either ``"Always-Taken"`` or ``"Always-Not-Taken"``.
+
+Tage-Table-Bits
+    Only needed for a ``Tage`` predictor.  The number of bits used to index entries in the tagged tables.  The number of entries in each of the tagged tables is obtained from the calculation: 1 << ``bits``.  For examples, a ``bits`` value of 12 would result in tagged tables with 4096 entries.
+
+Num-Tage-Tables
+    Only needed for a ``Tage`` predictor.  The number of tagged tables used by the predictor, in addition to a default prediction table (i.e., the BTB).  Therefore, a value of 3 for ``Num-Tage-Tables`` would result in four total prediction tables: one BTB and three tagged tables.  If no tagged tables are desired, it is recommended to use the ``GenericPredictor`` instead.
+
+Tage-Length
+    Only needed for a ``Tage`` predictor.  The number of bits used to tage the entries of the tagged tables.
```
FinnWilkinson: Following on from last weeks meeting where all BP related options are ignored if not needed and given a default value --- I wonder if having SimEng print all config values that are using their default value (i.e. it hasn't been explicitly defined in the config.yaml) could be useful? Gives an extra layer of debugging help. 

Open to thoughts from everyone on this @jj16791 @ABenC377 @dANW34V3R @JosephMoore25

src/include/simeng/branchpredictors/BranchHistory.hh
```diff
@@ -0,0 +1,115 @@
...
+
+class BranchHistory {
+ public:
+  BranchHistory(uint64_t size) : size_(size) {
+    history_ = {0};
+    for (uint8_t i = 0; i < (size_ / 64); i++) {
+      history_.push_back(0);
+    }
+  }
+  ~BranchHistory() {};
```
FinnWilkinson, line 21: Add newline before the de-constructor

src/include/simeng/branchpredictors/BranchHistory.hh
```diff
@@ -0,0 +1,115 @@
...
+ * access and manipulate large branch histories, as are needed in
+ * sophisticated branch predictors.
+ *
+ * The bits of the branch history are stored in a vector of uint64_t values,
+ * and their access/manipulation is facilitated by the public functions. */
+
+class BranchHistory {
+ public:
+  BranchHistory(uint64_t size) : size_(size) {
+    history_ = {0};
```
FinnWilkinson: I think this line is not needed as we are using push_back anyway?

src/include/simeng/branchpredictors/BranchHistory.hh
```diff
@@ -0,0 +1,115 @@
...
+ private:
+  /** The number of bits of branch history stored in this branch history */
+  uint64_t size_;
+
+  /** A vector containing the bits of the branch history.  The bits are
+   * arranged such that the most recent branches are stored in uint64_t at
+   * index 0 of the vector, then the next most recent at index 1 and so forth.
+   * Within each uint64_t, the most recent branches are recorded in the
+   * least-significant bits. */
+  std::vector<uint64_t> history_;
```
FinnWilkinson: Given `history_`'s size doesn't change after initialisation and we are storing the `size_` anyway, we could use a pointer (smart or C-style) rather than a vector to make it more light weight? Vector typically over allocates, but this isn't useful for this class' case

src/include/simeng/branchpredictors/TagePredictor.hh
```diff
@@ -0,0 +1,168 @@
...
+ * A prediction is made on the basis of the prediction table using the longest
+ * global history that has an entry corresponding to the present branch
+ * (matching tag).
+ * */
+
+class TagePredictor : public BranchPredictor {
+ public:
+  /** Initialise predictor models. */
+  TagePredictor(ryml::ConstNodeRef config = config::SimInfo::getConfig());
+  ~TagePredictor();
```
FinnWilkinson: newline before destructor

src/include/simeng/branchpredictors/TagePredictor.hh
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * A prediction is made on the basis of the prediction table using the longest
+ * global history that has an entry corresponding to the present branch
+ * (matching tag).
+ * */
+
+class TagePredictor : public BranchPredictor {
+ public:
+  /** Initialise predictor models. */
+  TagePredictor(ryml::ConstNodeRef config = config::SimInfo::getConfig());
```
FinnWilkinson: Are there situations where a non-SimInfo config is used? otherwise we could just call getConfig inside the function itself

src/include/simeng/branchpredictors/TagePredictor.hh
```diff
@@ -0,0 +1,168 @@
...
+  /** A return address stack. */
+  std::deque<uint64_t> ras_;
+
+  /** RAS history with instruction address as the keys. A non-zero value
+   * represents the target prediction for a return instruction and a 0 entry for
+   * a branch-and-link instruction. */
+  std::map<uint64_t, uint64_t> rasHistory_;
+
+  /** The size of the RAS. */
+  uint16_t rasSize_;
```
FinnWilkinson: Is this different to `ras_.size()`?

FinnWilkinson (changes requested):
Generally good, a few minor points

docs/sphinx/user/configuring_simeng.rst
ABenC377: Yeah, I think that this is a really good idea!

src/include/simeng/branchpredictors/BranchHistory.hh
ABenC377: Weirdly it was needed (heap memory errors when it was removed).  BUT, moving to a smart pointer removes its need regardless.

src/include/simeng/branchpredictors/TagePredictor.hh
ABenC377: Not that I could think of -- I just carried over the approach used in the other BP classes.  However, I can't see a downside to doing it this way?

src/include/simeng/branchpredictors/TagePredictor.hh
ABenC377: Yes, rasSize_ is the size of the RAS (i.e., the number of elements that can be stored in the RAS).  Whereas ras_.size() will give the number of elements currently in the RAS. 

I've updated the comment to make this more clear.

src/include/simeng/branchpredictors/BranchHistory.hh
ABenC377: Yep, that makes sense.  I've changed to a unique_ptr

docs/sphinx/user/configuring_simeng.rst
ABenC377: Discussed in meeting -- not going to print to terminal.  Adding issue to have a separate file outputting config options, including flagging the defaults we've chosen.

docs/sphinx/user/configuring_simeng.rst
```diff
@@ -149,13 +149,13 @@ The Branch-Prediction section contains those options to parameterise the branch
 The current options include:
 
 Type
-    The type of branch predictor that is used, the options are ``Generic``, and ``Perceptron``.  Both types of predictor use a branch target buffer with each entry containing a direction prediction mechanism and a target address.  The direction predictor used in ``Generic`` is a saturating counter, and in ``Perceptron`` it is a perceptron.
+    The type of branch predictor that is used, the options are ``Generic``, ``Perceptron``, and ``Tage``.  Each of these types of predictor use prediction tables with each entry containing a direction prediction mechanism and a target address.  The direction predictor used in ``Generic`` and ``TAGE`` is a saturating counter, and in ``Perceptron`` it is a perceptron.  ``TAGE`` also uses a series of further, tagged prediction tables to provide predictions informed by greater branch histories.
```
jj16791: is there a good reason behind using `Tage` and `TAGE`?

configs/a64fx.yaml
```diff
@@ -29,10 +29,15 @@ Queue-Sizes:
   Load: 40
   Store: 24
 Branch-Predictor:
```
jj16791, line 33: Some TX2 diagrams note it's use of a multi-history branch predictor. I assume this is TAGE-like so maybe apply this config update to the TX2 YAML as well?

src/lib/branchpredictors/TagePredictor.cc
```diff
@@ -0,0 +1,279 @@
...
+
+  // Set up non-tagged default prediction table
+  btb_ = std::vector<std::pair<uint8_t, uint64_t>>(1ul << btbBits_,
+                                                   {satCntVal, 0});
+
+  // Set up tagged prediction tables
+  for (uint32_t i = 0; i < numTageTables_; i++) {
+    std::vector<TageEntry> newTable;
+    for (uint32_t j = 0; j < (1ul << tageTableBits_); j++) {
+      TageEntry newEntry = {2, 0, 1, 0};
```
jj16791: Would we not want to initialise the `TageEntry` with a SatCnt equal to the once used in the `btb_`?

src/lib/branchpredictors/TagePredictor.cc
```diff
@@ -0,0 +1,279 @@
...
+  uint64_t target = btb_[index].second;
+  return {direction, target};
+}
+
+uint64_t TagePredictor::getTaggedIndex(uint64_t address, uint8_t table) {
+  // Get the XOR of the address (sans two least-significant bits) and the
+  // global history (folded onto itself to make it of the correct size).
+  uint64_t h1 = (address >> 2);
+  uint64_t h2 = globalHistory_.getFolded(1ull << (table + 1), tageTableBits_);
+  // Then truncat the XOR to make it fit thed esired size of an index
```
jj16791: *the desired

jj16791: *truncate

jj16791 (changes requested):

FinnWilkinson (dismissed):

FinnWilkinson:
Could you also add tage to the a64fx_SME.yaml config

docs/sphinx/user/configuring_simeng.rst
ABenC377: There is not.  I've udpated to `Tage` throughout, as this is the capitalisation used in the config yaml.

configs/a64fx.yaml
ABenC377: Yes, that sounds like it would be.  I've updated the TX2 config as well.

src/lib/branchpredictors/TagePredictor.cc
ABenC377: Yep, good catch

FinnWilkinson (dismissed):

docs/sphinx/user/configuring_simeng.rst
dANW34V3R: Seems like the creator uses all forms of capitalisation

src/include/simeng/branchpredictors/BranchHistory.hh
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /** Adds a branch outcome ('isTaken') to the global history */
+  void addHistory(bool isTaken) {
+    for (int8_t i = size_ / 64; i >= 0; i--) {
+      history_[i] <<= 1;
+      if (i == 0) {
+        history_[i] |= ((isTaken) ? 1 : 0);
+      } else {
+        history_[i] |= (((history_[i - 1] & (1ull << 63)) > 0) ? 1 : 0);
```
dANW34V3R: Does this need the conditional statement? After doing the AND you could shift right by 63 to get your 0 or 1. Would be slightly fewer cycles and more understandable/readable in my eyes (you may disagree)

src/include/simeng/branchpredictors/BranchHistory.hh
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  /** Updates the state of a branch that has already been added to the global
+   * history at 'position', where 'position' is 0-indexed and starts from the
+   * most recent history.  I.e., to update the most recently added branch
+   * outcome, 'position' would be 0.
+   * */
+  void updateHistory(bool isTaken, uint64_t position) {
+    if (position < size_) {
```
dANW34V3R, line 82: Should we assert position being < size_ as above, or are there cases where this could "validly" be greater? For instance, if you are trying to update an entry that has been lost from the history because there have been too many branches in the meantime?

src/include/simeng/branchpredictors/BranchHistory.hh
```diff
@@ -0,0 +1,114 @@
...
+#include <cstdint>
+#include <memory>
+
+namespace simeng {
+/** A class for storing a branch history.  Needed for cases where a branch
+ * history of more than 64 bits is required.  This class makes it easier to
+ * access and manipulate large branch histories, as are needed in
+ * sophisticated branch predictors.
+ *
+ * The bits of the branch history are stored in a vector of uint64_t values,
```
dANW34V3R: "vector" should be "array"

src/include/simeng/branchpredictors/TagePredictor.hh
```diff
@@ -0,0 +1,170 @@
...
+   * will have 2^bits entries. */
+  uint8_t btbBits_;
+
+  /** A 2^bits length vector of pairs containing a satCntBits_-bit saturating
+   * counter and a branch target.  This is the untagged, default prediction
+   * table. */
+  std::vector<std::pair<uint8_t, uint64_t>> btb_;
+
+  /** The bitlength of the Tagged tables' indices.
+   * Each tagged table with have 2^bits entries. */
```
dANW34V3R: With -> will

src/lib/branchpredictors/TagePredictor.cc
```diff
@@ -0,0 +1,283 @@
...
+  // global history (folded onto itself to make it of the correct size).
+  uint64_t h1 = (address >> 2);
+  uint64_t h2 = globalHistory_.getFolded(1ull << (table + 1), tageTableBits_);
+  // Then truncate the XOR to make it fit the desired size of an index
+  return (h1 ^ h2) & ((1 << tageTableBits_) - 1);
+}
+
+uint64_t TagePredictor::getTag(uint64_t address, uint8_t table) {
+  // Hash function here is pretty arbitrary
+  uint64_t h1 = address;
```
dANW34V3R: Any reason not to remove the 2 LSBs here?

docs/sphinx/user/configuring_simeng.rst
```diff
@@ -174,7 +174,16 @@ RAS-entries
...
+    Only needed for ``Generic`` and ``Tage`` predictors.  The static predictor used when no dynamic prediction is available. The options are either ``"Always-Taken"`` or ``"Always-Not-Taken"``.
+
+Tage-Table-Bits
+    Only needed for a ``Tage`` predictor.  The number of bits used to index entries in the tagged tables.  The number of entries in each of the tagged tables is obtained from the calculation: 1 << ``bits``.  For examples, a ``bits`` value of 12 would result in tagged tables with 4096 entries.
+
+Num-Tage-Tables
+    Only needed for a ``Tage`` predictor.  The number of tagged tables used by the predictor, in addition to a default prediction table (i.e., the BTB).  Therefore, a value of 3 for ``Num-Tage-Tables`` would result in four total prediction tables: one BTB and three tagged tables.  If no tagged tables are desired, it is recommended to use the ``GenericPredictor`` instead.
+
+Tage-Length
+    Only needed for a ``Tage`` predictor.  The number of bits used to tage the entries of the tagged tables.
```
dANW34V3R: Is the "tage" in the latter sentence meant to be that or rather "tag"

dANW34V3R (commented):
Very clean PR. Nicely precisely commented and everything is very easily readable. I like the branch history class which is also well explained

src/include/simeng/branchpredictors/BranchHistory.hh
ABenC377: I think the conditional is needed here.  Whats being loaded into the uint64 depends on where it is in the vector.  All but the least-significant uint64s get the MSB of the next uint64 added as the LSB. But the least-significant uint64 gets isTaken added as the LSB.  However, if I'm misunderstanding your Q LMK.

src/include/simeng/branchpredictors/BranchHistory.hh
ABenC377: Exactly as you say, I don't think that this should be an assert as the core may validly try to update a history that is no longer being tracked.  The reason that we should allow this is to allow the pipeline not to need to know the size of the branch history.  We're already ensuring that this doesn't cause problems with our if statement on 82.

src/lib/branchpredictors/TagePredictor.cc
ABenC377: Yes.  Ideally the hashes for the tag and the index should never each produce the same value for two different branches.  Therefore, because the index does remove the 2 LSBs, keeping them here makes the information being passed into the hashes different and so improves the accuracy of the BP (reduces the risk of this type of accidental clashing).

docs/sphinx/user/configuring_simeng.rst
jj16791: The original paper uses TAGE. Motion to replace all instances to "TAGE"

ABenC377: Yep, now replaced Tage with TAGE

src/include/simeng/branchpredictors/BranchHistory.hh
FinnWilkinson: I guess you could change line 71 to `history_[i] |= (history_[i - 1] & 0x80000000) >> 63;` as this logic just wants to know if the MSB is set. This new logic would shift the MSB to be the LSB, hence, matching the previous logic

src/include/simeng/branchpredictors/TAGEPredictor.hh
```diff
@@ -0,0 +1,170 @@
...
+  std::deque<ftqEntry> ftq_;
+
+  /** The number of bits used to form the saturating counter in a BTB entry. */
+  uint8_t satCntBits_;
+
+  /** The number of previous branch directions recorded globally. */
+  uint16_t globalHistoryLength_;
+
+  /** A return address stack. */
+  std::deque<uint64_t> ras_;
```
FinnWilkinson, line 143: Given the name (i.e. return address **stack**), why the need for a deque?

src/lib/branchpredictors/TAGEPredictor.cc
```diff
@@ -0,0 +1,283 @@
...
+      rasHistory_[address] = ras_.back();
+      ras_.pop_back();
+    }
+    predTable = -1;
+  } else if (type == BranchType::SubroutineCall) {
+    prediction.isTaken = true;
+    // Subroutine call branches must push their associated return address to RAS
+    if (ras_.size() >= rasSize_) {
+      ras_.pop_front();
+    }
```
FinnWilkinson, line 81: This seems a bit dangerous no? If we have too many nested sub-routine calls then we remove the oldest?

FinnWilkinson: I guess its only a RAS for branch prediction so not actually dangerous.... but would mean loads of sub-routine calls would mean the outer most returns are un-predictable?

src/lib/branchpredictors/TAGEPredictor.cc
```diff
@@ -0,0 +1,283 @@
...
+  // If address interacted with RAS, rewind entry
+  auto it = rasHistory_.find(address);
+  if (it != rasHistory_.end()) {
+    uint64_t target = it->second;
+    if (target != 0) {
+      // If history entry belongs to a return instruction, push target back onto
+      // stack
+      if (ras_.size() >= rasSize_) {
+        ras_.pop_front();
+      }
```
FinnWilkinson, line 133: Similar comment to above

FinnWilkinson (commented):
All still looks great - just a few minor things / clarification on the TAGE's RAS

src/include/simeng/branchpredictors/TAGEPredictor.hh
ABenC377: So, it's because it has to be of a set size, which means we need to pop from the front if we get too many entries.  therefore, we can't use the stack structure in C++.  Whereas, deque works.  In reality I think the RAS would be a loop buffer.

src/lib/branchpredictors/TAGEPredictor.cc
ABenC377: Yeah, not dangerous in the sense of making the CPU work incorrectly.  But it would make the predictor work less well if we go over the size.  This is reflective of real-world CPUs though, and we can change this with the RAS-entries value in the config file.  So I think that this is okay.

FinnWilkinson: Thanks for the explanation

src/include/simeng/branchpredictors/BranchHistory.hh
```diff
@@ -0,0 +1,114 @@
...
+      }
+    }
+  }
+
+  /** Removes the most recently added branch from the history */
+  void rollBack() {
+    for (uint8_t i = 0; i <= (size_ / 64); i++) {
+      history_[i] >>= 1;
+      if (i < (size_ / 64)) {
+        history_[i] |= (((history_[i + 1] & 1) > 0) ? (1ull << 63) : 0);
```
FinnWilkinson: Could this logic also be simplified?
`history_[i] |= (history_[i + 1] & 0x00000001) << 63;`

ABenC377: Yep, love it

FinnWilkinson (dismissed):

CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ FetchContent_Declare(
   googletest
   GIT_REPOSITORY https://github.com/google/googletest.git
   # this versions requires CMake >= 2.8.12 which silences the deprecation warning
-  GIT_TAG        release-1.11.0
+  GIT_TAG        v1.13.0
```
jj16791, line 33: Any reason for 1.13 over other versions?

FinnWilkinson (dismissed):

JosephMoore25 (dismissed):
Looks all good to me. Clean implementation. Would maybe change a couple of data structures (namely branch history to a circular buffer) just to optimise memory usage, but not worth halting this PR further for. Nice work!

src/include/simeng/branchpredictors/BranchHistory.hh
dANW34V3R: This

src/lib/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ set(SIMENG_SOURCES
     branchpredictors/AlwaysNotTakenPredictor.cc
     branchpredictors/GenericPredictor.cc
     branchpredictors/PerceptronPredictor.cc
+        branchpredictors/TAGEPredictor.cc
```
dANW34V3R: Accidentally indented

test/unit/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ set(TEST_SOURCES
     RegisterValueTest.cc
     PerceptronPredictorTest.cc
     SpecialFileDirGenTest.cc
+        TAGEPredictorTest.cc
```
dANW34V3R: Remove indentation

src/lib/branchpredictors/TagePredictor.cc
dANW34V3R: Is this the standard way to reduce clashes?

dANW34V3R (commented):
Very clean. Couple of minor comments but happy to approve

src/lib/branchpredictors/TagePredictor.cc
ABenC377: There isn't a standard way.  I cant find two implementations that use the same hashing policies.  This is arbitrary.  However, keeping the 2 LSBs here does improve BP accuracy.

FinnWilkinson (dismissed):

JosephMoore25 (dismissed):

CMakeLists.txt
ABenC377: 1.13 was the lowest version that fixed the issue.  I went for oldest version on the basis that it would presumably have the fewest other changes.  Just checking to make sure that most recent version (1.17) also works.  Happy to go with either (assuming 1.17 also works).

jj16791: 1.13 sounds good

jj16791 (approved):
All looks good, great work Alex! Looks like this cango in after it passes all the GH checks

JosephMoore25 (approved):

FinnWilkinson (approved):

dANW34V3R (approved):

PR #454, opened 2025-03-03
Updated enums for latency group 11 in both A64FX configs.

FinnWilkinson:
Updated opcode enums for latency group 11 in both a64fx config files as they changed when capstone was updated.


JosephMoore25 (approved):

ABenC377 (approved):

dANW34V3R (approved):

jj16791 (approved):
Looked like an arduous effort, thanks for doing this Finn

PR #455, opened 2025-05-14
NEON microp fix

FinnWilkinson:
For all NEON multi structure loads or stores with micro-op logic (i.e. `ld1     {v4.4s,v5.4s,v6.4s,v7.4s}, [x10], x11`) the last opcode indicator was set to true for multiple micro-ops.

This PR corrects this and makes sure only the last micro-op is set as such.

Performance or functionality doesn't seem to be affected, but it does inflate the instruction retired count.


ABenC377 (approved):

dANW34V3R (approved):

JosephMoore25 (approved):

jj16791 (approved):
Well spotted, lucky this didn't break something.

